PR #1, opened 2023-05-25
feat: gas service and deposit service deployment

deanamiel:
PR for AxelarGasService and AxelarDepositService deployment

Note: Artifacts are added manually to be used by deployment scripts. This ensures that correct contract bytecode is deployed.


contracts/Import.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.9;
+
+import '@axelar-network/axelar-cgp-solidity/contracts/gas-service/AxelarGasService.sol';
```
milapsheth, line 5: Did you check if the same bytecode is derived after compilation? We want to derive the same address on all chains

milapsheth:
fix the lint issues?

contracts/Import.sol
milapsheth: If it gives a different address, then maybe you can try that npm package to compile imported contracts in case that behaves better

deanamiel: It seems to produce the same address on different chains yes, but this is only a temporary solution until we republish the cgp npm module to include the artifacts folder. Let me know if you prefer that I use the hardhat-dependency-compiler npm package in the meantime

deanamiel: ^ as per our discussion bytecode is not same when compared to our previous deployment method so I updated the PR to manually include required artifacts

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Axelar contract deployment scripts",
   "main": "index.js",
   "scripts": {
+    "build": "rm -rf artifacts && hardhat compile",
```
milapsheth, line 7: not needed anymore

milapsheth, line 7:
```suggestion
```

milapsheth (approved):

PR #8, opened 2023-06-27
feat: deploy create3 deployer script

deanamiel:
PR to deploy the create3 deployer contract on supported chains using the constAddressDeployer contract.

### DIFF
--- a/evm/deploy-contract.js
+++ b/evm/deploy-contract.js
@@ -46,17 +46,17 @@ async function getConstructorArgs(contractName, config) {
             return [gateway, governanceChain, governanceAddress, minimumTimeDelay];
         }
 
-        case 'AxelarMultisigMintLimiter': {
+        case 'Multisig': {
             const signers = contractConfig.signers;
 
             if (!isAddressArray(signers)) {
-                throw new Error(`Missing AxelarMultisigMintLimiter.signers in the chain info.`);
+                throw new Error(`Missing Multisig.signers in the chain info.`);
             }
 
             const threshold = contractConfig.threshold;
 
             if (!isNumber(threshold)) {
-                throw new Error(`Missing AxelarMultisigMintLimiter.threshold in the chain info.`);
+                throw new Error(`Missing Multisig.threshold in the chain info.`);
             }
 
             return [signers, threshold];
@@ -74,8 +74,8 @@ async function getConstructorArgs(contractName, config) {
  * Deploy a smart contract using the create3 deployment method.
  */
 async function deploy(options, chain) {
-    const { artifactPath, contractName, privateKey, verify, verifyEnv } = options;
-    const verifyOptions = verify && verifyEnv ? { env: verifyEnv, chain: chain.name } : null;
+    const { env, artifactPath, contractName, privateKey, verify } = options;
+    const verifyOptions = verify ? { env, chain: chain.name } : null;
 
     const wallet = new Wallet(privateKey);
 
@@ -172,7 +172,6 @@ program.addOption(new Option('-n, --chainNames <chainNames>', 'chain names').mak
 program.addOption(new Option('-p, --privateKey <privateKey>', 'private key').makeOptionMandatory(true).env('PRIVATE_KEY'));
 program.addOption(new Option('-s, --salt <salt>', 'salt to use for create2 deployment'));
 program.addOption(new Option('-v, --verify', 'verify the deployed contract on the explorer').env('VERIFY'));
-program.addOption(new Option('-ve, --verifyEnv <verifyEnv>', 'environment for contract verification'));
 
 program.action((options) => {
     main(options);
--- a/evm/deploy-contract.js
+++ b/evm/deploy-contract.js
@@ -71,11 +71,11 @@ async function getConstructorArgs(contractName, config) {
 }
 
 /*
- * Deploy a non-upgradable smart contract using the create3 deployment method.
+ * Deploy a smart contract using the create3 deployment method.
  */
 async function deploy(options, chain) {
-    const { artifactPath, contractName, privateKey, verifyEnv } = options;
-    const verifyOptions = verifyEnv ? { env: verifyEnv, chain: chain.name } : null;
+    const { artifactPath, contractName, privateKey, verify, verifyEnv } = options;
+    const verifyOptions = verify && verifyEnv ? { env: verifyEnv, chain: chain.name } : null;
 
     const wallet = new Wallet(privateKey);
 
@@ -107,11 +107,9 @@ async function deploy(options, chain) {
     const salt = options.salt || contractName;
     printInfo('Contract deployment salt', salt);
 
-    let create3Deployer;
+    const create3Deployer = contracts.Create3Deployer?.address;
 
-    if (contracts.Create3Deployer && isAddress(contracts.Create3Deployer.address)) {
-        create3Deployer = contracts.Create3Deployer.address;
-    } else {
+    if (!create3Deployer) {
         throw new Error(`Create3 deployer does not exist on ${chain.name}.`);
     }
 
@@ -159,7 +157,7 @@ async function main(options) {
 
 const program = new Command();
 
-program.name('deploy-non-upgradable').description('Deploy non-upgradable contracts');
+program.name('deploy-contract').description('Deploy contracts using create3');
 
 program.addOption(
     new Option('-e, --env <env>', 'environment')
@@ -174,6 +172,7 @@ program.addOption(new Option('-n, --chainNames <chainNames>', 'chain names').mak
 program.addOption(new Option('-p, --privateKey <privateKey>', 'private key').makeOptionMandatory(true).env('PRIVATE_KEY'));
 program.addOption(new Option('-s, --salt <salt>', 'salt to use for create2 deployment'));
 program.addOption(new Option('-v, --verify', 'verify the deployed contract on the explorer').env('VERIFY'));
+program.addOption(new Option('-ve, --verifyEnv <verifyEnv>', 'environment for contract verification'));
 
 program.action((options) => {
     main(options);
--- a/evm/deploy-create3-deployer.js
+++ b/evm/deploy-create3-deployer.js
@@ -8,13 +8,13 @@ const { predictContractConstant } = require('@axelar-network/axelar-gmp-sdk-soli
 const { Command, Option } = require('commander');
 const chalk = require('chalk');
 
-const { deployCreate2 } = require('./upgradable');
-const { printInfo, writeJSON } = require('./utils');
+const { printInfo, writeJSON, deployCreate2 } = require('./utils');
 const implementationJson = require('../artifacts/contracts/deploy/Create3Deployer.sol/Create3Deployer.json');
 const contractName = 'Create3Deployer';
 
 async function deploy(options, chain) {
-    const { env, privateKey, verify } = options;
+    const { privateKey, verifyEnv } = options;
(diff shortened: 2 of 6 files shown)


evm/upgradable.js
```diff
@@ -42,6 +42,26 @@ async function deployCreateUpgradable(
     return new Contract(proxy.address, implementationJson.abi, wallet);
 }
 
+async function deployCreate2(
```
milapsheth, line 45: This should be present in a different file since this one is for upgradable contracts, perhaps in `deploy/utils.js`

evm/upgradable.js
```diff
@@ -42,6 +42,26 @@ async function deployCreateUpgradable(
...
+    wallet,
+    contractJson,
+    key = Date.now(),
+    args = [],
+    gasOptions = null,
+    env = 'testnet',
+    chain = 'ethereum',
+    verify = false,
+) {
+    const implementation = await deployContractConstant(constAddressDeployerAddress, wallet, contractJson, key, args, gasOptions?.gasLimit);
```
milapsheth, line 56: there is no notion of implementation here
```suggestion
    const contract = await deployContractConstant(constAddressDeployerAddress, wallet, contractJson, key, args, gasOptions?.gasLimit);
```

evm/deploy-create3-deployer.js
```diff
@@ -0,0 +1,111 @@
...
+
+    for (const chain of chains) {
+        await deploy(options, config.chains[chain.toLowerCase()]);
+        writeJSON(config, `${__dirname}/../info/${options.env}.json`);
+    }
+}
+
+const program = new Command();
+
+program.name('deploy-create3-deployer').description('Deploy create3 deployer');
```
milapsheth, line 93: maybe move the main deploy scripts under a `deploy` subfolder so we can distinguish it better, and we don't need a deploy prefix in the file names

evm/deploy-create3-deployer.js
```diff
@@ -0,0 +1,111 @@
...
+
+    contractConfig.salt = salt;
+    contractConfig.address = contract.address;
+    contractConfig.deployer = wallet.address;
+
+    printInfo(`${chain.name} | ConstAddressDeployer:`, constAddressDeployer);
+    printInfo(`${chain.name} | Create3Deployer`, contractConfig.address);
+}
+
+async function main(options) {
```
milapsheth, line 74: this can be moved to a utils.js in the deploy folder and reused in scripts

evm/deploy-create3-deployer.js
```diff
@@ -0,0 +1,111 @@
...
+            chain.tokenSymbol,
+        )} and nonce ${await provider.getTransactionCount(wallet.address)} on ${chain.name}.`,
+    );
+
+    const contracts = chain.contracts;
+
+    if (!contracts[contractName]) {
+        contracts[contractName] = {};
+    } else {
+        throw new Error(`Create3 deployer is already deployed on ${chain.name} ${env} at address: ${contracts[contractName].address}`);
```
milapsheth, line 36: allow overwriting, might wanna deploy again for testing etc. vs using the official stored value

evm/deploy-create3-deployer.js
```diff
@@ -0,0 +1,111 @@
...
+
+    printInfo('Deployer address', wallet.address);
+
+    const rpc = chain.rpc;
+    const provider = getDefaultProvider(rpc);
+
+    console.log(
+        `Deployer has ${(await provider.getBalance(wallet.address)) / 1e18} ${chalk.green(
+            chain.tokenSymbol,
+        )} and nonce ${await provider.getTransactionCount(wallet.address)} on ${chain.name}.`,
```
milapsheth, line 28: not urgent but can also be refactored. in particular, each chain config can have contract specific or chain specific gas options. We could sanity check if the balance covers the gas limit * gas price or so. can make a note to tackle this later

evm/deploy-create3-deployer.js
```diff
@@ -0,0 +1,111 @@
...
+        await deploy(options, config.chains[chain.toLowerCase()]);
+        writeJSON(config, `${__dirname}/../info/${options.env}.json`);
+    }
+}
+
+const program = new Command();
+
+program.name('deploy-create3-deployer').description('Deploy create3 deployer');
+
+program.addOption(
```
milapsheth, line 95: we can refactor this to a different file to reuse too

evm/deploy-create3-deployer.js
deanamiel: moved deployment scripts under a deploy subfolder within the evm folder

evm/deploy-create3-deployer.js
deanamiel: will make a note

evm/upgradable.js
deanamiel: Instead of having separate files for upgradable and non-upgradable deployment scripts, why not move them all into deploy/utils.js?

evm/deploy-upgradable.js
```diff
@@ -95,7 +95,8 @@ function getUpgradeArgs(contractName, config) {
  * Deploy or upgrade an upgradable contract that's based on the init proxy pattern.
  */
 async function deploy(options, chain) {
-    const { env, artifactPath, contractName, privateKey, upgrade, verify } = options;
+    const { artifactPath, contractName, privateKey, upgrade, verifyEnv } = options;
```
milapsheth, line 98: there is no `verifyEnv` option. Those are separate options and if verify is enabled, then verifyOptions can be created below

evm/deploy-non-upgradable.js
milapsheth, line 1: just call it deploy-contract.js or something? non upgradability is a default behaviour

evm/deploy-non-upgradable.js
```diff
@@ -0,0 +1,182 @@
...
+            const minimumTimeDelay = contractConfig.minimumTimeDelay;
+
+            if (!isNumber(minimumTimeDelay)) {
+                throw new Error(`Missing InterchainGovernance.minimumTimeDelay in the chain info.`);
+            }
+
+            return [gateway, governanceChain, governanceAddress, minimumTimeDelay];
+        }
+
+        case 'AxelarMultisigMintLimiter': {
```
milapsheth, line 49: switch to multisig

evm/deploy-non-upgradable.js
```diff
@@ -0,0 +1,182 @@
...
+    const salt = options.salt || contractName;
+    printInfo('Contract deployment salt', salt);
+
+    let create3Deployer;
+
+    if (contracts.Create3Deployer && isAddress(contracts.Create3Deployer.address)) {
+        create3Deployer = contracts.Create3Deployer.address;
+    } else {
+        throw new Error(`Create3 deployer does not exist on ${chain.name}.`);
+    }
```
milapsheth, line 116:
```suggestion
    const create3Deployer = contracts.Create3Deployer?.address;

    if (!create3Deployer) {
        throw new Error(`Create3 deployer does not exist on ${chain.name}.`);
    }
```

evm/utils.js
```diff
@@ -11,15 +11,57 @@ const { writeFile } = require('fs');
...
+
+    return contract;
+};
+
+const deployCreate2 = async (
+    constAddressDeployerAddress,
+    wallet,
+    contractJson,
+    args = [],
+    key = Date.now(),
```
milapsheth, line 37: switch key to `salt` everywhere

evm/deploy-non-upgradable.js
deanamiel: I agree that's more concise but the contract name is pulled directly from the command line and is used to get the implementation file path. I can add an if-statement to look for AxelarMultisigMintLimiter.json if contract name is Multisig or leave as is, thoughts?

evm/deploy-upgradable.js
deanamiel: I added verify as an option and create verifyOptions only if verify is enabled

evm/deploy-non-upgradable.js
milapsheth: there is no special mint limiter contract anymore though. We switched to [multisig](https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/governance/Multisig.sol) as a general purpose contract

evm/deploy-contract.js
```diff
@@ -174,6 +172,7 @@ program.addOption(new Option('-n, --chainNames <chainNames>', 'chain names').mak
 program.addOption(new Option('-p, --privateKey <privateKey>', 'private key').makeOptionMandatory(true).env('PRIVATE_KEY'));
 program.addOption(new Option('-s, --salt <salt>', 'salt to use for create2 deployment'));
 program.addOption(new Option('-v, --verify', 'verify the deployed contract on the explorer').env('VERIFY'));
+program.addOption(new Option('-ve, --verifyEnv <verifyEnv>', 'environment for contract verification'));
```
milapsheth, line 175: why is this needed? we already have `--env`

evm/deploy-contract.js
```diff
@@ -71,11 +71,11 @@ async function getConstructorArgs(contractName, config) {
 }
 
 /*
- * Deploy a non-upgradable smart contract using the create3 deployment method.
+ * Deploy a smart contract using the create3 deployment method.
  */
 async function deploy(options, chain) {
-    const { artifactPath, contractName, privateKey, verifyEnv } = options;
-    const verifyOptions = verifyEnv ? { env: verifyEnv, chain: chain.name } : null;
+    const { artifactPath, contractName, privateKey, verify, verifyEnv } = options;
```
milapsheth, line 77:
```suggestion
    const { env, artifactPath, contractName, privateKey, verify } = options;
```

evm/upgradable.js
```diff
@@ -47,7 +47,7 @@ async function deployCreate2Upgradable(
     implementationConstructorArgs = [],
     proxyConstructorArgs = [],
     setupParams = '0x',
-    key = Date.now(),
+    salt = Date.now(),
```
milapsheth, line 50: should be a mandatory arg to force user to decide

evm/deploy-contract.js
```diff
@@ -159,7 +157,7 @@ async function main(options) {
 
 const program = new Command();
 
-program.name('deploy-non-upgradable').description('Deploy non-upgradable contracts');
+program.name('deploy-contract').description('Deploy contracts using create3');
```
milapsheth, line 160: can you add a flag (say `--deployer create3`) that decides which strategy to use (create3/create2/create)? The default can be create3, and we don't need a separate script for each approach I think. Extending this can also be a separate PR if you prefer

evm/deploy-non-upgradable.js
deanamiel: ok that makes sense, I'll change it then

evm/deploy-contract.js
deanamiel: good point

evm/upgradable.js
deanamiel: removed default value

evm/deploy-contract.js
deanamiel: That's an interesting option, although I don't think we really deploy any non-upgradable contracts using create2 or create other than the deployers themselves which have their own scripts right? Were you thinking of potentially merging those scripts into this one as well?

deanamiel: We can discuss in more detail and include it in a separate PR, will resolve for the time being

milapsheth: yep, don't need a separate script and gives the choice to use the appropriate method. There are some nicer properties of create2 over create3 that might make sense for certain deployments

PR #15, opened 2023-07-12
fix(scripts): miscellaneous corrections

jcs47:
- Allow regular `http` get request
- import `ConstAddressDeployer` and `ConstCreate3Deployer` from `axelar-network/axelar-gmp-sdk-solidity` repo
- use correct function name for deploying contract at `evm/deploy-const-address-deployer.js`
- check that governance chain address is a string, but not an EVM address
- ran linter and prettier

### DIFF
--- a/evm/deploy-contract.js
+++ b/evm/deploy-contract.js
@@ -33,7 +33,7 @@ async function getConstructorArgs(contractName, config) {
 
             const governanceAddress = contractConfig.governanceAddress;
 
-            if (!isAddress(governanceAddress)) {
+            if (!isString(governanceAddress)) {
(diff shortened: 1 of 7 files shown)


evm/deploy-contract.js
```diff
@@ -33,7 +33,7 @@ async function getConstructorArgs(contractName, config) {
 
             const governanceAddress = contractConfig.governanceAddress;
 
-            if (!isAddress(governanceAddress)) {
+            if (!isString(governanceAddress)) {
```
deanamiel, line 36: makes sense since governance address is on Axelar network

deanamiel (approved):

PR #22, opened 2023-07-25
feat: add generic contract deployment

milapsheth:


evm/deploy-contract.js
```diff
@@ -170,92 +157,43 @@ async function deploy(options, chain) {
...
-    }
+    const contract = await deployContract(
+        deployMethod,
+        wallet,
+        contractJson,
+        constructorArgs,
+        { salt, deployerContract },
+        gasOptions,
+        verifyOptions,
+    );
```
deanamiel, line 197: Do we still want to store the salt in the info file for create2/create3 methods?

deanamiel (approved):

evm/deploy-contract.js
milapsheth: ah, good catch, forgot to refactor that

PR #23, opened 2023-07-26
fix: restore broken flags

jcs47:


evm/deploy-const-address-deployer.js
```diff
@@ -12,9 +12,7 @@ const { printInfo, writeJSON, predictAddressCreate, deployCreate } = require('./
 const contractJson = require('@axelar-network/axelar-gmp-sdk-solidity/dist/ConstAddressDeployer.json');
 const contractName = 'ConstAddressDeployer';
 
-async function deployConstAddressDeployer(wallet, chain, privateKey, verifyOptions, yes = true, force = false, ignore = false) {
-    const deployerWallet = new Wallet(privateKey, wallet.provider);
-
+async function deployConstAddressDeployer(wallet, chain, options = null, verifyOptions = null) {
```
milapsheth, line 15: its partly relies on this, but we can fix that separately

milapsheth (approved):

PR #27, opened 2023-08-01
fix: owner as constructor

deanamiel:
PR to add owner as a required constructor arg for Operators deployment.

### DIFF
--- a/evm/deploy-contract.js
+++ b/evm/deploy-contract.js
@@ -24,7 +24,7 @@ const {
     saveConfig,
 } = require('./utils');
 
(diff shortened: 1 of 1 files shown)


evm/deploy-contract.js
```diff
@@ -113,7 +113,13 @@ async function getConstructorArgs(contractName, config) {
         }
 
         case 'Operators': {
-            return [];
+            const owner = contractConfig.owner;
```
milapsheth, line 116: if owner is not set, use the deployer address instead, and set the field in the contract config

milapsheth (approved):

PR #30, opened 2023-08-24
feat: execute contract script [AXE-1610]

blockchainguyy:
JIRA: https://axelarnetwork.atlassian.net/browse/AXE-1610


blockchainguyy:
Manual testing with mock contracts:

- https://goerli.etherscan.io/tx/0xb3a1a40feebc3f712d6df4420f215e6d445a704adec288f51de93bf36798f5f9
- https://goerli.etherscan.io/tx/0x0e0cbcc2cb708853b3478f9505b8adbec35afe8b0f61ee560afb364869581cfd#eventlog
- https://goerli.etherscan.io/tx/0x96ccedbc5ae47fcaa9757e7bcb6d44b26a68e3630003b72ea82c5617f3aa8387#eventlog

evm/executeContract.js
```diff
@@ -0,0 +1,244 @@
...
+            if (!isNumber(amount)) {
+                throw new Error('Missing or incorrect approval amount from user input.');
+            }
+
+            const callData = targetContract.interface.encodeFunctionData('approve', [recipient, amount]);
+            return callData;
+        }
+
+        default: {
+            throw new Error('The method name do not matches with any of the specified choices');
```
deanamiel, line 96:
```suggestion
            throw new Error('The method name does not match any of the specified choices');
```

evm/executeContract.js
```diff
@@ -0,0 +1,244 @@
...
+program.addOption(
+    new Option('-e, --env <env>', 'environment')
+        .choices(['local', 'devnet', 'stagenet', 'testnet', 'mainnet'])
+        .default('testnet')
+        .makeOptionMandatory(true)
+        .env('ENV'),
+);
+program.addOption(new Option('-n, --chainNames <chainNames>', 'chain names').makeOptionMandatory(true));
+program.addOption(
+    new Option('-pc, --pathCallContract <pathCallContract>', 'artifact path for the called contract').makeOptionMandatory(true),
```
deanamiel, line 207: Do we want to make this optional and add the Operators contract path as the default value? If we expect this script to be used for the Operators contract execute function in most cases

evm/executeContract.js
```diff
@@ -0,0 +1,244 @@
...
+);
+program.addOption(
+    new Option('-pt, --pathTargetContract <pathTargetContract>', 'artifact path for the target contract').makeOptionMandatory(true),
+);
+program.addOption(
+    new Option('-a, --callContractAddress <contractAddress>', 'The contract address in which we will call executeContract function')
+        .makeOptionMandatory(true)
+        .env('CONTRACT_ADDR'),
+);
```
deanamiel, line 216: I assume the call contract will be one of our deployed contracts, so do we want to pass in the contract name and then get the address from the info file? We could also make this optional and add a default value if we change this to contract name, not address.

evm/executeContract.js
```diff
@@ -0,0 +1,244 @@
...
+program.addOption(
+    new Option('-t, --targetAddress <targetAddress>', 'The address of the contract to be called')
+        .makeOptionMandatory(true)
+        .env('TARGET_ADDR'),
```
deanamiel, line 220: Let's discuss with @milapsheth but we might always want to pass target address, native value, and method name via the command line

evm/executeContract.js
```diff
@@ -0,0 +1,244 @@
...
+);
+program.addOption(new Option('-n, --chainNames <chainNames>', 'chain names').makeOptionMandatory(true));
+program.addOption(
+    new Option('-pc, --pathCallContract <pathCallContract>', 'artifact path for the called contract').makeOptionMandatory(true),
+);
+program.addOption(
+    new Option('-pt, --pathTargetContract <pathTargetContract>', 'artifact path for the target contract').makeOptionMandatory(true),
+);
+program.addOption(
+    new Option('-a, --callContractAddress <contractAddress>', 'The contract address in which we will call executeContract function')
```
deanamiel, line 213:
```suggestion
    new Option('-a, --callContractAddress <contractAddress>', 'Contract address for the executeContract call.')
```

evm/executeContract.js
blockchainguyy: @deanamiel as far as I understand artifact path of the called contract would come from a different repo, and its location can change based on the user's setup of repositories. Setting a default path can lead to dependencies between two repositories, which might not be favourable.

If my understanding is incorrect, please let me know. This was my first script for the repo.

evm/executeContract.js
blockchainguyy: This generic option was created to support a wider variety of use cases as per the ticket description, if the requirement is specific to default types- yes we should change it.

evm/executeContract.js
blockchainguyy: For now, I will be following this pattern from the `deploy-contract.js` script. But here as well there is no default artifact location. I can make the tag optional and set different artifact paths based on `contractName` option (once I add it) of the script, but for it I would like to understand local setup preferred by all devs working on these modules.

evm/executeContract.js
blockchainguyy: Adding `contractName` and removing `contractAddress`. Added this option earlier to make things more generic, but nice point. :)

evm/executeContract.js
blockchainguyy: Removing native value and method name from env. For the target address its your call, we can discuss with @milapsheth as well.

blockchainguyy: @milapsheth can you pease comment here.

milapsheth: yeah, cli works

PR #31, opened 2023-08-24
feat: governance script

deanamiel:
Script to perform governance actions as described in AXE-1607:
https://axelarnetwork.atlassian.net/browse/AXE-1607

### DIFF
--- a/evm/governance.js
+++ b/evm/governance.js
@@ -161,8 +161,14 @@ async function processCommand(options, chain, config) {
                 throw new Error(`TimeLock proposal is not yet eligible for execution.`);
             }
 
-            const tx = await governanceContract.executeProposal(target, calldata, nativeValue, gasOptions);
-            const receipt = tx.wait();
+            let receipt;
+
+            try {
+                const tx = await governanceContract.executeProposal(target, calldata, nativeValue, gasOptions);
+                receipt = tx.wait();
+            } catch (error) {
+                console.log(error);
+            }
 
             const eventEmitted = wasEventEmitted(receipt, governanceContract, 'ProposalExecuted');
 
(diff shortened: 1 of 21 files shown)


evm/governance.js
```diff
@@ -0,0 +1,285 @@
...
+
+    const unixEta = etaToUnixTimestamp(eta);
+
+    const types = ['uint256', 'address', 'bytes', 'uint256', 'uint256'];
+    const values = [0, target, calldata, nativeValue, unixEta];
+
+    let gmpPayload;
+
+    switch (governanceAction) {
+        case 'scheduleTimeLock': {
```
blockchainguyy, line 88: Shouldn't we also call the `execute()` function for the first 4 switch cases, as to actually process the specific `ServiceGovernanceCommand` command, or may be i am missing something.

evm/governance.js
```diff
@@ -0,0 +1,285 @@
...
+    wasEventEmitted,
+    printWarn,
+    printError,
+} = require('./utils');
+const IGovernance = require('@axelar-network/axelar-gmp-sdk-solidity/interfaces/IAxelarServiceGovernance.json');
+
+async function processCommand(options, chain) {
+    const { contractName, address, governanceAction, calldata, nativeValue, eta, privateKey, yes } = options;
+
+    if (contractName !== 'AxelarServiceGovernance' && contractName !== 'InterchainGovernance') {
```
blockchainguyy, line 32: If these are the only 2 names then we can add them as choices in the `contractName` option so won't have to check for any other value.

evm/utils.js
```diff
@@ -554,6 +560,54 @@ const deployContract = async (
...
 };
 
+/**
+ * Validate if the input string matches the time format YYYY-MM-DDTHH:mm:ss
+ *
+ * @param {string} timeString - The input time string.
+ * @return {boolean} - Returns true if the format matches, false otherwise.
+ */
+function isValidTimeFormat(timeString) {
+    const regex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}$/;
```
blockchainguyy, line 570: This regex can also pass invalid date time like for ex- `2023-08-10T80:70:70` Here hours is 80, minutes 70, seconds 70 which will pass but is wrong.
We would need to modify this.

evm/governance.js
deanamiel: yeah I like that idea

evm/utils.js
deanamiel: Good catch, I updated the regex to validate months, days, hours, minutes, and seconds. Please take a look

evm/governance.js
deanamiel: The external execute function is in AxelarExecutable and it checks for gateway approval before calling the internal execute function so calling it as part of the switch case would fail as the call will not yet have been approved by the gateway.

milapsheth (approved):

PR #34, opened 2023-08-28
feat: script to manage ownership

deanamiel:
Script to manage ownership for all contracts that inherit the SDK Ownable.sol as per AXE-1594:
https://axelarnetwork.atlassian.net/browse/AXE-1594

### DIFF
--- a/evm/ownership.js
+++ b/evm/ownership.js
@@ -8,19 +8,28 @@ const {
     getDefaultProvider,
     utils: { isAddress },
     constants: { AddressZero },
-    ContractFactory,
+    Contract,
 } = ethers;
 const { Command, Option } = require('commander');
-const { printInfo, printWalletInfo, loadConfig, saveConfig } = require('./utils');
+const { printInfo, printWalletInfo, loadConfig } = require('./utils');
+const IOwnableJson = require('@axelar-network/axelar-gmp-sdk-solidity/artifacts/contracts/interfaces/IOwnable.sol/IOwnable.json');
 
 async function processCommand(options, chain) {
-    const { artifactPath, contractName, ownershipAction, privateKey, newOwner } = options;
+    const { contractName, address, ownershipAction, privateKey, newOwner } = options;
 
     const contracts = chain.contracts;
     const contractConfig = contracts[contractName];
 
-    if (contractConfig && !contractConfig.address) {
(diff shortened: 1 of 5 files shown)


evm/ownership.js
```diff
@@ -0,0 +1,200 @@
...
+async function main(options) {
+    const config = loadConfig(options.env);
+
+    const chain = options.chain;
+
+    if (config.chains[chain.toLowerCase()] === undefined) {
+        throw new Error(`Chain ${chain} is not defined in the info file`);
+    }
+
+    await processCommand(options, config.chains[chain.toLowerCase()], config);
```
blockchainguyy, line 165: Function definition has 2 input args: `options`, `chain`. here we are supplying 3 args. We should remove `config` if not used.

blockchainguyy:
https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/interfaces/IOwnable.sol

This interface implements both:
- transfer ownership in a single tx
- 2 step transfer: Propose and accept

Why have we implemented both options?

evm/ownership.js
```diff
@@ -0,0 +1,200 @@
...
+
+    const rpc = chain.rpc;
+    const provider = getDefaultProvider(rpc);
+
+    const wallet = new Wallet(privateKey, provider);
+    await printWalletInfo(wallet);
+
+    printInfo('Contract name', contractName);
+
+    const contractPath = artifactPath + contractName + '.sol/' + contractName + '.json';
```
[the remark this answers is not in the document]
deanamiel: changed to use IOwnable and added an optional address override

evm/ownership.js
deanamiel: Good catch

deanamiel:
> https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/interfaces/IOwnable.sol
> 
> This interface implements both:
> 
> * transfer ownership in a single tx
> * 2 step transfer: Propose and accept
> 
> Why have we implemented both options?

My understanding is that it provides more flexibility. There may be some circumstances where we want to perform a single step ownership transfer, but other instances where it would be more secure to perform a 2-step transfer where the proposed owner has to accept the ownership role. Not only for us, but also for developers using our SDK.

milapsheth (approved):

PR #35, opened 2023-08-28
feat: allow gateway deployment with amplifier worker set

eguajardo:
Update to gateway deployment script to allow choosing from where to load operators for auth module, either amplifier contract or core

https://axelarnetwork.atlassian.net/browse/AXE-1537

### DIFF
--- a/evm/deploy-gateway-v5.0.x.js
+++ b/evm/deploy-gateway-v5.0.x.js
@@ -16,7 +16,7 @@ const chalk = require('chalk');
 
 async function getAuthParams(config, chainName, options) {
     printLog('retrieving addresses');
-    const { addresses, weights, threshold } = await getEVMAddresses(config, chainName, options.keyId, options.amplifier);
+    const { addresses, weights, threshold } = await getEVMAddresses(config, chainName, options);
     printObj(JSON.stringify({ addresses, weights, threshold }));
     const paramsAuth = [defaultAbiCoder.encode(['address[]', 'uint256[]', 'uint256'], [addresses, weights, threshold])];
     return paramsAuth;
--- a/evm/deploy-gateway-v5.0.x.js
+++ b/evm/deploy-gateway-v5.0.x.js
@@ -29,7 +29,7 @@ function getProxyParams(governance, mintLimiter) {
 async function deploy(config, options) {
     const { chainName, privateKey, reuseProxy, verify, yes } = options;
 
-    const contractName = options.amplifier ? 'AxelarAmplifierGateway' : 'AxelarGateway';
+    const contractName = 'AxelarGateway';
 
     const chain = config.chains[chainName] || { contracts: {}, name: chainName, id: chainName, tokenSymbol: 'ETH' };
     const rpc = options.rpc || chain.rpc;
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
     "ethers": "^5.7.2"
   },
   "devDependencies": {
+    "@cosmjs/cosmwasm-stargate": "^0.31.1",
```
milapsheth, line 31: this should be a normal dependency, since we'll export the deployment scripts from this package

evm/utils.js
```diff
@@ -466,6 +470,21 @@ const getEVMAddresses = async (config, chain, keyID = '') => {
     return { addresses, weights, threshold };
 };
 
+const getAmplifierKeyAddresses = async (config, chainName, keyID = '') => {
+    const client = await CosmWasmClient.connect(config.axelar.rpc);
+    const key = await client.queryContractSmart(config.axelar.contracts.Multisig.address, {
+        get_key: { key_id: { owner: config.axelar.chains[chainName].contracts.MultisigProver.address, subkey: keyID } },
```
milapsheth, line 476:
```suggestion
        get_key: { key_id: { owner: config.axelar.contracts.MultisigProver[chains].address, subkey: keyID } },
```
not all axelar contracts will be per chain

evm/deploy-gateway-v5.0.x.js
```diff
@@ -29,7 +29,7 @@ function getProxyParams(governance, mintLimiter) {
 async function deploy(config, options) {
     const { chainName, privateKey, reuseProxy, verify, yes } = options;
 
-    const contractName = 'AxelarGateway';
+    const contractName = options.amplifier ? 'AxelarAmplifierGateway' : 'AxelarGateway';
```
milapsheth, line 32: the gateway is the same, no need for a separate identifier i think

evm/deploy-gateway-v5.0.x.js
```diff
@@ -14,9 +14,9 @@ const readlineSync = require('readline-sync');
 const { Command, Option } = require('commander');
 const chalk = require('chalk');
 
-async function getAuthParams(config, chain) {
+async function getAuthParams(config, chainName, options) {
     printLog('retrieving addresses');
-    const { addresses, weights, threshold } = await getEVMAddresses(config, chain);
+    const { addresses, weights, threshold } = await getEVMAddresses(config, chainName, options.keyId, options.amplifier);
```
milapsheth, line 19:
```suggestion
    const { addresses, weights, threshold } = await getEVMAddresses(config, chainName, options);
```
pass in options to the helper too

package.json
eguajardo: moved it to normal dependencies, I put it in dev since there were other dependencies used in the script that were in dev as well, should those be moved as well?

evm/deploy-gateway-v5.0.x.js
eguajardo: aren't we going to have two separate gateways for chains that are migrating from consensus to amplifier?

milapsheth (approved):

PR #42, opened 2023-09-01
feat: update amp worker set query to not use keyID

eguajardo:
Adjust amplifier contract query to new key rotation logic. Key ID is not longer needed to get current worker set

https://axelarnetwork.atlassian.net/browse/AXE-1683

### DIFF
--- a/evm/utils.js
+++ b/evm/utils.js
@@ -543,7 +543,7 @@ const getAmplifierKeyAddresses = async (config, chain) => {
     });
 
     const weightedAddresses = workerSet.signers.map((signer) => ({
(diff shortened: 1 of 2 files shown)


evm/utils.js
```diff
@@ -407,7 +407,7 @@ const getEVMAddresses = async (config, chain, options = {}) => {
     const keyID = options.keyID || '';
```
milapsheth, line 407: `keyID` from CLI is being used here so shouldn't remove the option above

eguajardo: added the option again

milapsheth (approved):

PR #46, opened 2023-09-13
feat: deploy gas service v6.1 with operators contract on testnet

milapsheth:
- [x] Improve gas service deployment
- [x] Add Operators deployment
- [x] Add gas service upgrade deployment

### DIFF
--- a/evm/deploy-upgradable.js
+++ b/evm/deploy-upgradable.js
@@ -20,15 +20,19 @@ function getProxy(wallet, proxyAddress) {
     return new Contract(proxyAddress, IUpgradable.abi, wallet);
 }
 
-async function getImplementationArgs(contractName, config) {
+async function getImplementationArgs(contractName, config, options) {
     const contractConfig = config[contractName];
 
     switch (contractName) {
         case 'AxelarGasService': {
+            if (options.args) {
+                contractConfig.collector = options.args;
+            }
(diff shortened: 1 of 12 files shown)


evm/ownership.js
```diff
@@ -148,6 +159,8 @@ async function processCommand(options, chain) {
 
             printInfo(`New contract owner: ${newOwner}`);
 
+            contractConfig.owner = newOwner;
```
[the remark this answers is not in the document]
milapsheth: we're already checking if the pending owner retrieved from the contract is the same as wallet address. i'd avoid storing since it's transient data

evm/deploy-upgradable.js
```diff
@@ -150,6 +159,7 @@ async function deploy(options, chain) {
         if (!yes) {
             const anwser = readlineSync.question(`Perform an upgrade for ${chain.name}? ${chalk.green('(y/n)')} `);
             if (anwser !== 'y') return;
+            console.log('');
```
deanamiel, line 162: Linter?

deanamiel (approved):

evm/deploy-upgradable.js
milapsheth: just adding a newline so next log is more readable

PR #47, opened 2023-09-13
feat: multisig script

deanamiel:
Multisig script as per [AXE-1612](https://axelarnetwork.atlassian.net/browse/AXE-1612)

[AXE-1612]: https://axelarnetwork.atlassian.net/browse/AXE-1612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evm/multisig.js
```diff
@@ -0,0 +1,325 @@
...
+    if (!isSigner) {
+        throw new Error(`Caller ${wallet.address} is not an authorized multisig signer.`);
+    }
+
+    const multisigInterface = new Interface(multisigContract.interface.fragments);
+
+    let topic;
+
+    if (multisigAction === 'withdraw') {
+        topic = multisigInterface.encodeFunctionData('withdraw', [target, nativeValue]);
```
milapsheth, line 42:
```suggestion
        topic = multisigContract.interface.encodeFunctionData('withdraw', [target, nativeValue]);
```
could also do this to avoid creating a separate interface object

evm/multisig.js
```diff
@@ -0,0 +1,325 @@
...
+
+// options for setTokenMintLimits
+program.addOption(new Option('-s, --symbols <symbols>', 'token symbols').makeOptionMandatory(false));
+program.addOption(new Option('-l, --limits <limits>', 'token limits').makeOptionMandatory(false));
+
+// option for transferMintLimiter
+program.addOption(new Option('-m, --mintLimiterAddress <mintLimiterAddress>', 'new mint limiter address').makeOptionMandatory(false));
+
+// options for withdraw
+program.addOption(new Option('-r, --recipientAddress <recipientAddress>', 'withdraw recipient address').makeOptionMandatory(false));
```
milapsheth, line 308: address suffix isn't needed since it's implied i think

evm/multisig.js
```diff
@@ -0,0 +1,325 @@
...
+
+            if (!isAddress(multisigTarget) && multisigTarget !== AddressZero) {
+                throw new Error(`Missing AxelarGateway address in the chain info.`);
+            }
+
+            const gatewayContract = new Contract(multisigTarget, IGateway.abi, wallet);
+            const targetInterface = new Interface(gatewayContract.interface.fragments);
+            const multisigCalldata = targetInterface.encodeFunctionData('setTokenMintLimits', [symbolsArray, limitsArray]);
+
+            await preExecutionChecks(multisigContract, multisigAction, wallet, multisigTarget, multisigCalldata, 0, yes);
```
milapsheth, line 147: `yes` doesn't need to be passed in. we can just ask a question here right before execution to proceed

evm/multisig.js
```diff
@@ -0,0 +1,325 @@
...
+                    )} is less than native value amount: ${formatEther(BigNumber.from(nativeValue))}`,
+                );
+            }
+
+            try {
+                const tx = await governanceContract.executeMultisigProposal(target, calldata, nativeValue);
+                await tx.wait();
+            } catch (error) {
+                printError(error);
+            }
```
milapsheth, line 261: don't think we need to catch the failure for execution. and if we want to allow that, it's simpler to do in `main`

evm/multisig.js
```diff
@@ -0,0 +1,325 @@
...
+        }
+    }
+}
+
+async function main(options) {
+    const config = loadConfig(options.env);
+
+    const chain = options.destinationChain;
+
+    if (config.chains[chain.toLowerCase()] === undefined) {
```
milapsheth, line 273: support `all` chain names

milapsheth (approved):

PR #51, opened 2023-09-15
chore: move getBytecodeHash function to axelar-chains-config

npty:


axelar-chains-config/src/utils/getBytecodeHash.js
```diff
@@ -0,0 +1,43 @@
...
+        bytecode = await provider.getCode(contractObject.address);
+    } else if (contractObject.bytecode) {
+        // Contract factory
+        bytecode = contractObject.bytecode;
+    } else {
+        throw new Error('Invalid contract object. Expected ethers.js Contract or ContractFactory.');
+    }
+
+    if (chain.toLowerCase() === 'polygon-zkevm') {
+        throw new Error('polygon-zkevm is not supported');
```
milapsheth, line 31:
```suggestion
        throw new Error('polygon-zkevm is not supported. Use getBytecodeHash from axelar-contract-deployments that handles polygon-zkevm's hash function');
```

milapsheth (approved):

PR #52, opened 2023-09-15
feat: offline signing using ledger [AXE-1805]

blockchainguyy:
Jira: https://axelarnetwork.atlassian.net/browse/AXE-1805

Commands Used:
- To Sign Offline for V5 Script: `node evm/deploy-gateway-v5.0.x.js -e testnet -n ethereum -p ledger -u --offline` 
- To Sign for send-tokens Script: `node evm/send-tokens.js -e testnet -n ethereum -p ledger -a 0.01 -y -r "testAddress1, testAddress2" --offline`
- To Broadcast transactions: `node evm/broadcast-transactions.js -e testnet -n ethereum --filePath "./tx/signed-tx-testnet-ethereum-send-tokens-address-"ledgerSignerAddress"-nonce-35.json" -y`
- To Update Static gasOptions when online using provider in chain config:   `node evm/update-static-gas-options.js -e testnet -n ethereum -y`
- To Update Nonces for specific addresses when online using provider in chain config:   `node evm/updateNonces.js -e testnet -n ethereum -a '["testAddress1"]' -y`
- To checl Wallet balance for specific addresses when online using provider in chain config:   `node evm/check-wallet-balance.js -e testnet -n ethereum -a '["testAddress1"]'`


evm/broadcast-transactions.js
```diff
@@ -0,0 +1,112 @@
...
+
+    await processTransactions(directoryPath, fileName, provider, signerAddress);
+}
+
+const program = new Command();
+
+program.name('broadcast-transactions').description('Broadcast all the pending signed transactions of the signer');
+
+program.addOption(
+    new Option('-d, --directoryPath <directoryPath>', 'The folder where all the signed tx files are stored').makeOptionMandatory(true),
```
milapsheth, line 99: Instead of a directory, maybe just take a list of files and loop over it. Easier when there's a single tx, and can use glob pattern in shell for multiple files.

evm/deploy-gateway-v5.0.x.js
```diff
@@ -2,6 +2,18 @@
```
milapsheth, line 2: here's how to test upgrade in online mode

1. Set test key in .env along with ADMIN_ADDRESSES=["test account"], and ADMIN_THRESHOLD=1
2. Deploy an v4.3 gateway, `node evm/deploy-gateway-v4.3.x.js -e testnet -n avalanche`
3. Deploy a new gateway while reusing proxy, `node evm/deploy-gateway-v5.0.x.js -e testnet -n avalanche -g test_account -m test_account --reuseProxy`
4. Upgrade gateway to the new implementation, `node evm/deploy-gateway-v5.0.x.js -e testnet -n avalanche -u`

After the changes, this should work as well as the offline mode

evm/broadcast-transactions.js
```diff
@@ -0,0 +1,112 @@
...
+);
+program.addOption(new Option('-f, --fileName <fileName>', 'The fileName where the signed tx are stored').makeOptionMandatory(true));
+program.addOption(
+    new Option('-r, --rpcUrl <rpcUrl>', 'The rpc url for creating a provider to broadcast the transactions').makeOptionMandatory(true),
+);
+program.addOption(new Option('-s, --signerAddress <signerAddress>', 'private key').makeOptionMandatory(true));
```
milapsheth, line 105: signer address shouldn't be needed for broadcasting since it's already in the tx and can be recovered from that if needed. Might even have different signers for the txs. I'd say just print the tx hashes for now (or if you want dump to a txs.json file with broadcast status) to keep it simpler.

evm/broadcast-transactions.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+const program = new Command();
+
+program.name('broadcast-transactions').description('Broadcast all the pending signed transactions of the signer');
+
+program.addOption(
+    new Option('-d, --directoryPath <directoryPath>', 'The folder where all the signed tx files are stored').makeOptionMandatory(true),
+);
+program.addOption(new Option('-f, --fileName <fileName>', 'The fileName where the signed tx are stored').makeOptionMandatory(true));
```
milapsheth, line 101: we can avoid introducing shorthands like `-f` except for the most important cli flags i think

evm/deploy-gateway-v5.0.x.js
```diff
@@ -341,10 +366,63 @@ async function upgrade(config, options) {
...
+    let signersData, transactions;
+
+    if (isOffline) {
+        directoryPath = directoryPath || './txs';
+        fileName = fileName || env.toLowerCase() + '-' + chain.name.toLowerCase() + '-' + 'signedUpgradeTransactions';
+        printInfo(`Storing signed Txs offline in file ${fileName}`);
+        nonce = await getLatestNonceAndUpdateData(directoryPath, fileName, wallet);
+        signersData = await getAllSignersData(directoryPath, fileName);
+        transactions = await getTransactions(directoryPath, fileName, signerAddress);
+        const { baseTx, signedTx } = await ledgerSign(
```
milapsheth, line 394: this is complicated to work with.

Ledger sign util should just take a tx and a chain object, populate any fields like chainId, gasPrice, nonce etc if missing to keep ux simpler

evm/deploy-gateway-v5.0.x.js
```diff
@@ -341,10 +366,63 @@ async function upgrade(config, options) {
...
+        gasLimit = block.gasLimit.toNumber() / 1000;
+
+        printInfo('Gas Price:', gasPrice.toString());
+        printInfo('Gas Limit:', gasLimit.toString());
+    } catch (error) {
+        printError('Gas price and limit could not be fetched from provider');
+        printObj(error);
+    }
+
+    let nonce = await getNonceFromProvider(provider, signerAddress);
```
milapsheth, line 384: Can you add a small script that fetches the current nonce for a given list of addresses and updates it in a nonces.json file? `ledgerSign` can just try reading for the nonce from this file for a signer. we do this in the [old scripts](https://github.com/axelarnetwork/key-management-scripts/blob/main/contracts/offline-sign.js#L842)

evm/deploy-gateway-v5.0.x.js
```diff
@@ -341,10 +366,63 @@ async function upgrade(config, options) {
...
+        tx.nonce = nonce;
+        tx.msg = `This transaction will perform upgrade of AxelarGateway contract having address ${gateway.address} with implementation ${contractConfig.implementation} on chain ${chain.name} with chainId ${chain.chainId}`;
+        tx.baseTx = baseTx;
+        tx.signedTx = signedTx;
+        tx.status = 'PENDING';
+        transactions.push(tx);
+
+        if (transactions) {
+            signersData[signerAddress] = transactions;
+            await updateSignersData(directoryPath, fileName, signersData);
```
milapsheth, line 418: don't think we need directory management in the script itself. we can write the tx to a provided file path. some bash scripts that sit one level higher can handle packaging etc.

evm/send-tokens.js
```diff
@@ -31,17 +62,80 @@ async function sendTokens(chain, options) {
...
+                        }`;
+                        throw new Error(error);
+                    }
+
+                    printInfo('Transaction hash', response.transactionHash);
+                } catch (error) {
+                    printError('Broadcasting Transaction failed');
+                    printObj(error);
+                }
+            }
```
milapsheth, line 121: hide details inside `sendTx`? It can print any errors, and return the tx hash, receipt, status if caller needs to know

evm/utils.js
```diff
@@ -540,6 +546,16 @@ function isValidTimeFormat(timeString) {
     return regex.test(timeString);
 }
 
+// Validate if the input privateKey is correct
+function isValidPrivateKey(privateKey) {
+    // Check if it's a valid hexadecimal string
+    if (!isHexString(privateKey) || privateKey.length !== 66) {
```
milapsheth, line 552: we often provide it without 0x prefix in our infra scripts which fails here

evm/deploy-gateway-v5.0.x.js
milapsheth: Would be useful to do this for [gas price](https://github.com/axelarnetwork/key-management-scripts/blob/main/contracts/offline-sign.js#L94) and gasLimit as well.   Could be read from the chain config as well temporarily, but I'd like to keep it separate as a next step since we need to keep a high enough gas price/limit to avoid hitting usual variances in the networks

evm/send-tokens.js
```diff
@@ -2,25 +2,50 @@
...
-    const balance = await printWalletInfo(wallet);
-    const amount = ethers.utils.parseEther(options.amount);
+    const { wallet, providerNonce } = await getWallet(privateKey, provider, ledgerPath);
+    const signerAddress = await wallet.getAddress();
+    let nonce = getLocalNonce(chain, signerAddress);
 
-    if (balance.lte(amount)) {
-        throw new Error(`Wallet has insufficient funds.`);
+    if (providerNonce !== undefined && providerNonce !== null && providerNonce > nonce) {
+        config.chains[chain.name.toLowerCase()] = updateLocalNonce(chain, providerNonce, signerAddress);
```
milapsheth, line 37: let's avoid updating nonces as part of the tx generation script. offline wallets won't be used online (since the benefit of keeping it offline is lost when connected online). As part of the process, update-nonces should be run before tx creation + sign

evm/send-tokens.js
```diff
@@ -31,18 +56,63 @@ async function sendTokens(chain, options) {
...
+    }
 
-        await tx.wait();
+    if (transactions) {
+        signersData[signerAddress] = transactions;
+        updateSignersData(filePath, signersData);
     }
+    // Updating Nonce data for this Address
+
+    config.chains[chain.name.toLowerCase()] = updateLocalNonce(chain, nonce, signerAddress);
```
milapsheth, line 114: same comment

evm/send-tokens.js
```diff
@@ -31,18 +56,63 @@ async function sendTokens(chain, options) {
...
+            if (!isValidNumber(nonceOffset)) {
+                throw new Error('Provided nonce offset is not a valid number');
+            }
+
+            nonce += parseInt(nonceOffset);
+        }
+
+        printInfo(`Storing signed Txs offline in file ${filePath}`);
+        signersData = await getAllSignersData(filePath);
+        transactions = await getTransactions(filePath, signerAddress);
```
milapsheth, line 77: can we skip managing txs via this JSON for now? it'll complicate our flows from before. For e.g. we don't always won't to broadcast all txs from the same signer right away (maybe broadcast for one chain first and test etc.). Instead, I'd just save the signed tx under a [standardized file name](https://github.com/axelarnetwork/key-management-scripts/blob/main/contracts/offline-sign.js#L644) (you can add signer address in the name as well). Maybe if filePath is not passed in, then this approach is used.

evm/updateNonces.js
milapsheth, line 1: update-nonces.js for consistency

evm/updateNonces.js
```diff
@@ -0,0 +1,74 @@
...
+const fs = require('fs');
+const { ethers } = require('hardhat');
+const { getDefaultProvider } = ethers;
+const readlineSync = require('readline-sync');
+
+const { printError, printInfo, printObj } = require('./utils');
+const { getNonceFromProvider, getAllSignersData } = require('./offline-sign-utils');
+
+function updateNonce(provider, addresses, filePath) {
+    try {
```
milapsheth, line 14: `nonces.json` or whatever filePath provided can have a schema like
`{ env: { chain: { signer: nonce } } }`

signer addresses can be passed in as done below

evm/broadcast-transactions.js
```diff
@@ -0,0 +1,82 @@
...
+    await processTransactions(filePath, provider);
+}
+
+const program = new Command();
+
+program.name('broadcast-transactions').description('Broadcast all the pending signed transactions of the signer');
+
+program.addOption(new Option('--filePath <filePath>', 'The file where the signed tx are stored').makeOptionMandatory(true));
+program.addOption(
+    new Option('-r, --rpcUrl <rpcUrl>', 'The rpc url for creating a provider to broadcast the transactions').makeOptionMandatory(true),
```
milapsheth, line 74: read chain name here. provider can be retrieved from chain config, otherwise user needs to feed this in themselves for every chain

evm/broadcast-transactions.js
```diff
@@ -0,0 +1,82 @@
...
+    providers: { JsonRpcProvider },
+} = ethers;
+const readlineSync = require('readline-sync');
+
+const { printError, printInfo, printObj } = require('./utils');
+const { sendTx, getAllSignersData, updateSignersData } = require('./offline-sign-utils');
+
+async function processTransactions(filePath, provider) {
+    try {
+        const signersData = await getAllSignersData(filePath);
```
milapsheth, line 16: i'd like to keep this super simple for now. read file -> read chain config -> send tx -> log success/error

evm/update-static-gas-options.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    return chain;
+}
+
+async function main(options) {
+    const { env, chainNames } = options;
+
+    const filePath = `${__dirname}/../axelar-chains-config/info/${env === 'local' ? 'testnet' : env}.json`;
+    const config = require(filePath);
```
milapsheth, line 47: see `loadConfig` in utils.js

evm/update-static-gas-options.js
```diff
@@ -0,0 +1,84 @@
...
+        const chain = config.chains[chainName.toLowerCase()];
+        config.chains[chainName.toLowerCase()] = await updateStaticGasOptions(chain, options, filePath);
+    }
+
+    saveConfig(config, env);
+}
+
+const program = new Command();
+
+program.name('Update-GasOptions').description('Update gasOptions to be used in offline signing');
```
milapsheth, line 67:
```suggestion
program.name('update-static-gas-options').description('Update staticGasOptions to be used when offline signing');
```

evm/update-static-gas-options.js
```diff
@@ -0,0 +1,84 @@
...
+
+program.addOption(
+    new Option('-e, --env <env>', 'environment')
+        .choices(['local', 'devnet', 'stagenet', 'testnet', 'mainnet'])
+        .default('testnet')
+        .makeOptionMandatory(true)
+        .env('ENV'),
+);
+program.addOption(new Option('-n, --chainNames <chainNames>', 'chain names').makeOptionMandatory(true));
+program.addOption(new Option('-r, --rpcUrl <rpcUrl>', 'The rpc url for creating a provider to fetch gasOptions').makeOptionMandatory(true));
```
milapsheth, line 77:
```suggestion
program.addOption(new Option('-r, --rpc <rpc>', 'The rpc for a chain'));
```
for consistency with existing scripts. rpc shouldn't be mandatory anywhere since we chain configs

evm/update-static-gas-options.js
```diff
@@ -0,0 +1,84 @@
...
+                network.chainId,
+            )} ${chalk.green('(y/n)')} `,
+        );
+        if (anwser !== 'y') return;
+    }
+
+    try {
+        const gasPrice = parseUnits((await provider.getGasPrice()).toString(), 'gwei') * 5;
+        const staticGasOptions = {};
+        staticGasOptions.gasLimit = 130000;
```
milapsheth, line 31:
```suggestion
        chain.staticGasOptions.gasLimit = chain.staticGasOptions.gasLimit || 3e6;
```
we can keep a high default

but if it's already set, then we shouldn't override. some chains have different gas metering.

evm/update-static-gas-options.js
```diff
@@ -0,0 +1,84 @@
...
+        staticGasOptions.gasPrice = gasPrice;
+        chain.staticGasOptions = staticGasOptions;
+        printInfo(`GasOptions updated succesfully and stored in config file ${filePath}`);
```
milapsheth, line 34:
```suggestion
        chain.staticGasOptions.gasLimit = chain.staticGasOptions.gasLimit || 3e6;
        chain.staticGasOptions.gasPrice = gasPrice;
        printInfo(`GasOptions updated succesfully and stored in config file ${filePath}`);
```

we can keep a high default

but if it's already set, then we shouldn't override. some chains have different gas metering.

evm/offline-sign-utils.js
```diff
@@ -0,0 +1,240 @@
...
+    BigNumber,
+    utils: { isAddress },
+} = ethers;
+const path = require('path');
+const { LedgerSigner } = require('@ethersproject/hardware-wallets');
+
+const { printError, printInfo, printObj, isValidPrivateKey } = require('./utils');
+
+// function to create a ledgerSigner type wallet object
+function getLedgerWallet(provider, path) {
```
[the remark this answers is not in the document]
milapsheth: it'll just cause more confusion and harder debugging if something goes wrong. don't think path needs to be passed into various tx scripts for now. we can always consider it in the future. but let's start with keeping it simple

milapsheth (approved):

PR #71, opened 2023-10-16
chore: explorer urls in config for mantle chain [AXE-2199]

blockchainguyy:
Have updated the config for new explorer url's for mantle chain


milapsheth:
closing since new api doesn't work yet

PR #90, opened 2023-10-23
feat: testnet centrifuge tests

kalidax:


axelar-chains-config/info/testnet.json
```diff
@@ -1,5 +1,15 @@
 {
   "chains": {
+    "centrifuge": {
+      "name": "centrifuge",
+      "id": "centrifuge",
```
milapsheth, line 5:
```suggestion
      "id": "centrifuge-2",
```
since it was added before already

PR #106, opened 2023-11-07
feat: dedicated ITS script

deanamiel:
Dedicated ITS script as per [AXE-2437](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-2437)

[AXE-2437]: https://axelarnetwork.atlassian.net/browse/AXE-2437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evm/cli-utils.js
```diff
@@ -24,7 +24,7 @@ const addBaseOptions = (program, options = {}) => {
     }
 
     if (options.address) {
-        program.addOption(new Option('-a, --address <address>', 'override address'));
+        program.addOption(new Option('--address <address>', 'override address'));
```
milapsheth, line 27: can we keep this and remove -a for artifact path?

evm/its.js
```diff
@@ -0,0 +1,723 @@
...
+                printWarn('Event not emitted in receipt.');
+            }
+
+            break;
+        }
+
+        case 'callContractWithInterchainToken': {
+            const isPaused = await interchainTokenService.paused();
+
+            if (isPaused) {
```
milapsheth, line 439: not a very important check. we'd know if we paused it

evm/its.js
```diff
@@ -0,0 +1,723 @@
...
+                throw new Error(`${action} invalid while service is paused.`);
+            }
+
+            const { tokenId, destinationChain, destinationAddress, amount, data } = options;
+
+            if (!isValidTokenId(tokenId)) {
```
milapsheth, line 445: Lot of code duplication. Some checks you can perform in the beginning if the option was provided, or create reusable functions

evm/its.js
```diff
@@ -0,0 +1,723 @@
...
+
+            const tokenIdBytes32 = hexZeroPad(tokenId.startsWith('0x') ? tokenId : '0x' + tokenId, 32);
+
+            const tokenManagerAddress = await interchainTokenService.tokenManagerAddress(tokenIdBytes32);
+            printInfo(`TokenManager address for tokenId: ${tokenId}:`, tokenManagerAddress);
+
+            break;
+        }
+
+        case 'validTokenManagerAddress': {
```
milapsheth, line 96: not needed, we can just print in the above method if token manager actually exists or not

evm/its.js
```diff
@@ -0,0 +1,723 @@
...
+
+            const receipt = await tx.wait(chain.confirmations);
+
+            const eventEmitted =
+                wasEventEmitted(receipt, interchainTokenService, 'TokenManagerDeployed') ||
+                wasEventEmitted(receipt, interchainTokenService, 'TokenManagerDeploymentStarted');
+
+            if (!eventEmitted) {
+                printWarn('Event not emitted in receipt.');
+            }
```
milapsheth, line 245: if it was a tx, then wait + event check can be done generically at the end for all cases, event to be checked can be defined here

evm/its.js
```diff
@@ -0,0 +1,723 @@
...
+        }
+
+        case 'deployTokenManager': {
+            const isPaused = await interchainTokenService.paused();
+
+            if (isPaused) {
+                throw new Error(`${action} invalid while service is paused.`);
+            }
+
+            const { salt, destinationChain, type, params, gasValue } = options;
```
milapsheth, line 210: Shall we just do arbitrary JSON encoded args instead?

evm/utils.js
```diff
@@ -293,6 +293,35 @@ function isValidCalldata(input) {
     return hexPattern.test(input.slice(2));
 }
 
+/**
+ * Checks if a given string is a valid tokenId.
+ *
+ * @param {string} input - The input string to check.
+ * @returns {boolean} - True if the input is a valid tokenId, false otherwise.
+ */
+function isValidTokenId(input) {
```
milapsheth, line 302: this is not a general utils method. it should be in ITS script

milapsheth (changes requested):

evm/its.js
deanamiel: moved this check to a separate function

evm/its.js
deanamiel: Mind elaborating here?

evm/its.js
```diff
@@ -0,0 +1,479 @@
...
+
+async function handleTx(tx, chain, contract, action, firstEvent, secondEvent) {
+    printInfo(`${action} tx`, tx.hash);
+
+    const receipt = await tx.wait(chain.confirmations);
+
+    const eventEmitted =
+        (firstEvent ? wasEventEmitted(receipt, contract, 'TokenManagerDeployed') : true) ||
+        (secondEvent ? wasEventEmitted(receipt, contract, 'TokenManagerDeploymentStarted') : false);
+
```
milapsheth, line 49: why are the event names hardcoded?

milapsheth (approved):

evm/its.js
deanamiel: good catch! will fix

.github/workflows/test.yaml
```diff
@@ -65,12 +65,13 @@ jobs:
...
           }' > ./axelar-chains-config/info/local.json
 
-      # Create .env file with default hardhat private key that's prefunded
+      # Create .env file with default hardhat private key that's prefunded, default salt is keccak256('salt')
       - name: Prepare .env
         run: |
           echo 'PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80' >> .env
           echo 'ENV=local' >> .env
           echo 'CHAINS=test' >> .env
+          echo 'SALT=0xa05e334153147e75f3f416139b5109d1179cb56fef6a4ecb4c4cbc92a7c37b70' >> .env
```
milapsheth, line 74: what's this for? Also, salt is usually a string and we hash in the script. If it's only needed by a specific script, then it should be passed in there and not applied globally to all cmds below

deanamiel: will change to provide both `rawSalt` and `salt` options

PR #140, opened 2023-12-19
fix: allow updating multiple trusted chains

milapsheth:


evm/its.js
```diff
@@ -486,14 +553,25 @@ async function processCommand(config, chain, options) {
...
 
+            // check if all trusted addresses match ITS address
+            for (const trustedAddress of trustedAddresses) {
+                if (trustedAddress !== interchainTokenServiceAddress) {
+                    printError(
+                        `Error: Trusted address ${trustedAddress} does not match InterchainTokenService address ${interchainTokenServiceAddress}`,
+                    );
+                }
+
+                break;
```
deanamiel, line 564: Why `break` here? Wouldn't this only check the first address in the array?

evm/its.js
```diff
@@ -539,6 +617,7 @@ if (require.main === module) {
 
     addExtendedOptions(program, { address: true, salt: true });
 
+    program.addOption(new Option('-c, --contractName <contractName>', 'contract name').default('InterchainTokenService'));
```
deanamiel, line 620: Can also add via addExtendedOptions, it will be mandatory but there won't be a default value:
`addExtendedOptions(program, { address: true, salt: true, contractName: true });`

milapsheth: the goal was to have the default value, and keep the flag for consistency (it's used by `mainProcessor` to skip chains where the contract doesn't exist)

evm/its.js
milapsheth: good catch, this was supposed to be inside the if. Finding one error at a time is sufficient (to avoid printing tons of addresses)

deanamiel: got it makes sense

evm/its.js
deanamiel: that makes sense

PR #141, opened 2023-12-19
feat: destination chain check

deanamiel:
Destination chain checks for ITS scripts as per [AXE-2853](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-2853)

### DIFF
--- a/evm/its.js
+++ b/evm/its.js
@@ -87,6 +87,24 @@ function compare(contractValue, configValue, variableName) {
     }
 }
 
+function isValidDestinationChain(config, chain, destinationChain, originalChain = null) {
+    const chains = config.chains;
+
+    const validDestination = Object.values(chains).some((chainObject) => chainObject.id === destinationChain);
+
+    if (!validDestination) {
+        throw new Error(`Invalid destination chain: ${destinationChain}`);
+    }
+
+    if (chain.id === destinationChain) {
+        throw new Error(`Destination chain ${destinationChain} equals source chain ${chain.id}`);
+    }
+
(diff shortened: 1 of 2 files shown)


evm/its.js
```diff
@@ -87,6 +87,24 @@ function compare(contractValue, configValue, variableName) {
     }
 }
 
+function isValidDestinationChain(config, chain, destinationChain, originalChain = null) {
```
deanamiel, line 90: Should I convert chain names to lower case or are we expecting them to be given with the correct case?

evm/its.js
```diff
@@ -87,6 +87,24 @@ function compare(contractValue, configValue, variableName) {
...
+
+    if (!validDestination) {
+        throw new Error(`Invalid destination chain: ${destinationChain}`);
+    }
+
+    if (chain.id === destinationChain) {
+        throw new Error(`Destination chain ${destinationChain} equals source chain ${chain.id}`);
+    }
+
+    if (originalChain && destinationChain === originalChain) {
```
deanamiel, line 103: Add these checks as well because I assume there would be no instance where the destination chain should equal the source chain or original chain (for factory functions)

evm/its.js
milapsheth, line 90: where is original chain passed? make this a generic util instead? can allow us to check if any chain is valid

deanamiel: Added a more generic function to general utils to check if a chain exists in the config

milapsheth (approved):

PR #149, opened 2024-01-10
fix: update avalanche explorer url for mainnet and testnet [AXL-2907]

blockchainguyy:
[Jira issue](https://axelarnetwork.atlassian.net/browse/AXE-2907)

### DIFF
--- a/axelar-chains-config/info/mainnet.json
+++ b/axelar-chains-config/info/mainnet.json
@@ -224,7 +224,7 @@
         }
       },
       "explorer": {
-        "url": "https://snowtrace.io",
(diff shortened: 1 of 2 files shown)


axelar-chains-config/info/mainnet.json
```diff
@@ -224,7 +224,7 @@
         }
       },
       "explorer": {
-        "url": "https://snowtrace.io",
+        "url": "https://avalanche.routescan.io",
```
milapsheth, line 227: this just redirects to snowtrace.dev. snowtrace.io still works so I think it's ok to keep that?

PR #151, opened 2024-01-12
feat: verify contracts on filfox

deanamiel:
Script to verify contracts on filfox explorer as per [AXE-2937](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-2937)

[AXE-2937]: https://axelarnetwork.atlassian.net/browse/AXE-2937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evm/verify-filfox.js
```diff
@@ -0,0 +1,85 @@
+const axios = require('axios');
+const fs = require('fs');
+const { Command, Option } = require('commander');
+const { addBaseOptions } = require('./cli-utils');
+const { validateParameters, loadConfig, printInfo, printError } = require('./utils');
+
+async function verifyFilfox(options) {
```
milapsheth, line 7: this should be integrated into our normal verifyContract method. If chain is filecoin then call this helper method with appropriate data. you can look into how the existing verify method extracts the compiler version/optimizer runs to do the same (maybe even import that method if possible from hardhat-verify)

evm/verify-filfox.js
```diff
@@ -0,0 +1,85 @@
...
+        isValidNumber: { optimizeRuns },
+    });
+
+    const sourceFiles = {
+        [`${contractName}.sol`]: {
+            content: fs.readFileSync(contractPath, 'utf8'),
+        },
+    };
+
+    const optimizerDetails = '';
```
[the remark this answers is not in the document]
deanamiel: Yeah I agree, was having some trouble importing the configurations from the hardhat config in other repositories

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -1,5 +1,139 @@
...
+
+    const sourceFiles = {
+        [`${contractFileName}.sol`]: {
+            content: fs.readFileSync(contractPath, 'utf8'),
+        },
+    };
+
+    let buildInfo;
+
+    try {
```
milapsheth, line 81: move try/catch to be specifically over the code that can fail

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -1,5 +1,139 @@
...
+async function verifyFilfox(env, contract, options) {
+    const { dir, contractName } = options;
+
+    if (!dir || !contractName) {
+        throw new Error('Invalid verification options');
+    }
+
+    const config = require(`../../info/${env}.json`);
+    const contractConfig = config.chains.filecoin.contracts[contractName];
+    const contractFileName = getContractFileName(contractConfig, contract);
```
milapsheth, line 69: isn't contractName already being passed in? you can enforce it to be passed in, and update usages in our code instead of the limited reverse map

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -1,5 +1,139 @@
...
+        optimizerDetails,
+        sourceFiles,
+        license: 'MIT License (MIT)',
+        evmVersion,
+        viaIR: false,
+        libraries: '',
+        metadata: '',
+    };
+
+    const api = config.chains.filecoin.explorer?.api;
```
milapsheth, line 119: get the API from hardhat network or options.api?

axelar-chains-config/src/utils/verifyContract.js
blockchainguyy: The config map has different names for addresses than actual contract file names, so this should fail @deanamiel.

blockchainguyy: @milapsheth , @deanamiel can we modify all `verifyContract()` calls within`verify-contract.js` to have a hardcoded contract name with it, which will simplify script execution for all the contracts. I will also be able to use this in my PR: https://github.com/axelarnetwork/axelar-contract-deployments/pull/162

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -1,5 +1,139 @@
...
+    while (directoriesToSearch.length > 0) {
+        const currentDirectory = directoriesToSearch.pop();
+        const filesAndDirectories = fs.readdirSync(currentDirectory, { withFileTypes: true });
+
+        for (const fileOrDirectory of filesAndDirectories) {
+            const fullPath = path.join(currentDirectory, fileOrDirectory.name);
+
+            if (fileOrDirectory.isDirectory()) {
+                directoriesToSearch.push(fullPath);
+            } else if (fileOrDirectory.isFile() && fileOrDirectory.name === targetFileName) {
```
blockchainguyy, line 51: Optional: if we convert `fileOrDirectory.name` and `targetFileName` to lowercase and compare them we should be able to simplify `getContractFileName` by removing slicing and first letter conversion to uppercase, which should increase code readability.

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -1,5 +1,139 @@
...
+        [`${contractFileName}.sol`]: {
+            content: fs.readFileSync(contractPath, 'utf8'),
+        },
+    };
+
+    let buildInfo;
+
+    try {
+        const buildInfoPath = path.join(dir, 'build-info');
+        const smallestFile = await getSmallestFile(buildInfoPath);
```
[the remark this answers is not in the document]
deanamiel: It seems that there are often multiple build files produced within the `build-info` folder, some are extremely large and unfeasible to parse just for compiler/optimize information but all seem to contain the same config information which is why I added a function to just parse the smallest file present. I agree it is a bit convoluted but I'm not sure how else to go about it

axelar-chains-config/src/utils/verifyContract.js
deanamiel: The `contractName` being passed would just be InterchainTokenService whereas I need each individual contract name within ITS to find the correct contract path to the flattened source code. @blockchainguyy yeah that is true, it will not work for a few contract names, I thought of adding an optional argument for the hardcoded contract name to `verifyContract()` as well but am not sure it's worth all these changes for only filfox verification for their specific verification endpoint requirements, thoughts @milapsheth ?

PR #153, opened 2024-01-17
feat: combine signatures into gateway batch

deanamiel:
Draft PR for manually combining validator signatures into gateway batch as per [AXE-2733](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-2733)

[AXE-2733]: https://axelarnetwork.atlassian.net/browse/AXE-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+                }
+
+                const validatorAddress = getAddressFromPublicKey(pubKey);
+
+                // const signer = verifyMessage(msgHash, signature);
+
+                // if (signer.toLowerCase() !== validatorAddress.toLowerCase()) {
+                //     printError('Signature is invalid for the given validator address', validatorAddress);
+                //     return;
+                // }
```
deanamiel, line 166: I want to add this sanity check but the way I was trying here kept failing even while `auth.validateProof` returns true, I assume that the function I am using to retrieve the original signer address is wrong, if you have any ideas @milapsheth @re1ro @blockchainguyy please let me know :)

deanamiel: Works with `ethers.utils.recoverAddress`

re1ro (approved):

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+    validateParameters,
+    getContractJSON,
+    getRandomBytes32,
+    getEVMAddresses,
+} = require('./utils');
+const { handleTx } = require('./its');
+const { getWallet } = require('./sign-utils');
+const { addBaseOptions } = require('./cli-utils');
+const IAxelarGateway = getContractJSON('IAxelarGateway');
+const IAxelarAuth = getContractJSON('IAxelarAuth');
```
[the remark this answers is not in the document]
deanamiel: yeah I guess that's a bit more efficient

evm/utils.js
```diff
@@ -32,6 +32,10 @@ const getSaltFromKey = (key) => {
     return keccak256(defaultAbiCoder.encode(['string'], [key.toString()]));
 };
 
+function getRandomBytes32() {
+    return keccak256(defaultAbiCoder.encode(['uint256'], [Math.floor(new Date().getTime() * Math.random())]));
```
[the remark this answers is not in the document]
deanamiel: looks cleaner, will change

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+                const pubKey = `0x${signatureJSON.pub_key}`;
+                const signature = `0x${signatureJSON.signature}`;
+
+                validateParameters({
+                    isNonEmptyString: { keyId },
+                    isKeccak256Hash: { msgHash },
+                    isValidCalldata: { pubKey, signature },
+                });
+
+                if (expectedKeyId !== keyId) {
```
milapsheth, line 149: current key id might have changed/rotated. instead check if key id in the file is valid, and matches for all signatures

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+            const batchSignatures = [];
+
+            let totalWeight = 0;
+
+            const expectedMessageHash = hashMessage(arrayify(keccak256(message)));
+
+            for (const signatureJSON of sortedSignatures) {
+                const keyId = signatureJSON.key_id;
+                const msgHash = `0x${signatureJSON.msg_hash}`;
+                const pubKey = `0x${signatureJSON.pub_key}`;
```
milapsheth, line 140: verify that pub key matches the validator address in the file as well

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+            }
+
+            if (!isValidProof) {
+                printError('Invalid batch proof');
+                return;
+            }
+
+            const input = defaultAbiCoder.encode(['bytes', 'bytes'], [message, proof]);
+
+            printInfo('Batch input data for gateway execute function', input);
```
milapsheth, line 213: if --execute is passed, execute the batch to keep it simpler

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+
+            validateParameters({
+                isNonEmptyString: { sourceChain, sourceAddress },
+                isValidAddress: { contractAddress },
+                isKeccak256Hash: { commandId, payloadHash, sourceTxHash },
+                isValidNumber: { sourceEventIndex },
+            });
+
+            const chainId = chain.chainId;
+            const commandID = commandId || getRandomBytes32();
```
milapsheth, line 85: start from the largest value and decrement until you get one that hasn't been executed yet perhaps

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+
+            const data = defaultAbiCoder.encode(
+                ['uint256', 'bytes32[]', 'string[]', 'bytes[]'],
+                [chainId, [commandID], [command], [params]],
+            );
+
+            const dataHash = hashMessage(arrayify(keccak256(data)));
+
+            printInfo('Original bytes message (pre-hash)', data);
+            printInfo('Message hash for validators to sign', dataHash);
```
milapsheth, line 100: print the full vald-sign cmd that needs to be run

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+    const rpc = chain.rpc;
+    const provider = getDefaultProvider(rpc);
+
+    const wallet = await getWallet(privateKey, provider, options);
+    await printWalletInfo(wallet);
+
+    printInfo('Batch Action', action);
+
+    switch (action) {
+        case 'computeMessageHash': {
```
milapsheth, line 74:
```suggestion
        case 'createBatchData': {
```

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+            const { commandId, sourceChain, sourceAddress, contractAddress, payloadHash, sourceTxHash, sourceEventIndex } = options;
+
+            validateParameters({
```
milapsheth, line 77: set sane defaults. in practice destination will be interchain/axelar service gov address. src chain/address needs to be axelar gov module address. payloadHash will be fed from the output of governance.js

srcTxHash/event can be set to 0

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+    addBaseOptions(program, { address: true });
+
+    program.addOption(
+        new Option('--action <action>', 'signature action').choices(['computeMessageHash', 'constructBatch', 'executeBatch']),
+    );
+    program.addOption(new Option('-m, --message <message>', 'bytes message (validators sign the hash of this message)').env('MESSAGE'));
+    program.addOption(new Option('-i, --input <input>', 'batch input consisting of bytes message (data) and bytes proof').env('INPUT'));
+
+    program.addOption(new Option('--commandId <commandId>', 'gateway command id').env('COMMAND_ID'));
+    program.addOption(new Option('--sourceChain <sourceChain>', 'source chain for contract call').env('SOURCE_CHAIN'));
```
milapsheth, line 261: update readme with example usage

evm/combine-signatures.js
```diff
@@ -0,0 +1,273 @@
...
+
+    program.name('combine-signatures').description('script to combine manually created signatures and construct gateway batch');
+
+    addBaseOptions(program, { address: true });
+
+    program.addOption(
+        new Option('--action <action>', 'signature action').choices(['computeMessageHash', 'constructBatch', 'executeBatch']),
+    );
+    program.addOption(new Option('-m, --message <message>', 'bytes message (validators sign the hash of this message)').env('MESSAGE'));
+    program.addOption(new Option('-i, --input <input>', 'batch input consisting of bytes message (data) and bytes proof').env('INPUT'));
```
milapsheth, line 258:
```suggestion
    program.addOption(new Option('-i, --batchData <input>', 'batch input consisting of bytes message (data) and bytes proof').env('INPUT'));
```
need a readable name

evm/combine-signatures.js
deanamiel:
```
const validatorWeight = validatorWeights[validatorAddress.toLowerCase()];

if (!validatorWeight) {
    printError('Validator does not belong to current epoch', validatorAddress);
    return;
}
```
This check should ensure that all the validator public keys correspond to valid validators in the current epoch

evm/README.md
```diff
@@ -181,3 +181,72 @@ To get details of options provided in the command run:
...
+    kubectl get pods -n stagenet-validator1
+    ```
+
+6. Exec into the validator pod. Example for stagenet validator 1:
+
+    ```bash
+    kubectl exec -ti axelar-core-node-validator1-5b68db5c49-mf5q6 -n stagenet-validator1 -c vald -- sh
+    ```
+
+7. Within the validator container, run the vald command from step 2 with the correct validator address. Validator addresses can be found by appending this endpoint `/axelar/multisig/v1beta1/key?key_id=[KEY_ID]` to the Axelar lcd url in the chains config file. Running this command should result in a JSON output containing the signature, similar to the example below:
```
blockchainguyy, line 232: @milapsheth is there a simpler way to identify the validator address of a container than looping through all the validators mentioned in the endpoint. this approach might not be viable on testnet or mainnet.

evm/README.md
```diff
@@ -181,3 +181,72 @@ To get details of options provided in the command run:
...
+        "validator": "axelarvaloper1yfrz78wthyzykzf08h7z6pr0et0zlzf0dnehwx"
+    }
+    ```
+
+8. Create a signatures folder in the root directory of this repository. Repeat step 7 until enough signatures have been generated to surpass the threshold weight for the current epoch (threshold weight can be found at the same endpoint as validator addresses). Store each signature JSON in a separate file within the signatures folder, filenames do not have to follow any specific convention.
+
+9. Run the construct batch command passing in the raw batch data logged in step 1. This will generate the input bytes that can be passed directly into the execute command on Axelar gateway. You can also optionally pass the `--execute` command which will automatically execute the batch on the Axelar gateway contract.
+
+    ```bash
+    node evm/combine-signatures.js -e [ENV] -n [CHAIN_NAME] --action constructBatch --batchData [RAW_BATCH_DATA]
```
blockchainguyy, line 249: This script is appending 0x in front of `pub_key` values of signature files, but the logs generated when signing within the validator already have 0x with it, so it would be good to handle this.

blockchainguyy: @deanamiel ^

blockchainguyy: Same with other values in the signature logged by validator

evm/combine-signatures.js
```diff
@@ -0,0 +1,294 @@
...
+
+                batchSignatures.push(signature);
+
+                if (totalWeight >= threshold) {
+                    break;
+                }
+            }
+
+            if (totalWeight < threshold) {
+                printError('Total signer weight less than threshold', totalWeight);
```
blockchainguyy, line 215: Can update to: 

```                
printError(`Total signer weight ${totalWeight} less than threshold`, threshold);
```

blockchainguyy (changes requested):

evm/combine-signatures.js
```diff
@@ -0,0 +1,294 @@
...
+                    isKeccak256Hash: { msgHash },
+                    isValidCalldata: { pubKey, signature },
+                });
+
+                if (prevKeyId !== keyId) {
+                    printError('Signatures do not contain consistent key IDs', keyId);
+                    return;
+                }
+
+                prevKeyId = keyId;
```
blockchainguyy, line 175: This line is not needed as if prevKeyId was not equal to keyId then it would have entered if block and returned.

evm/README.md
```diff
@@ -181,3 +181,72 @@ To get details of options provided in the command run:
...
 ```
+
+## Combine Validator Signatures
+
+The following script is for emergency situations where we need to ask validators to manually sign gateway approvals. These validator signatures must be collected and stored within separate files in the signatures folder in the root directory. The script will then combine these signatures into a gateway batch and optionally execute this batch on the Axelar gateway contract on the specified chain.
+
+Below are instructions on how to utilize this script:
+
+1. Construct the gateway batch data by passing in the payload hash gotten from running the `governance.js` script.
+
```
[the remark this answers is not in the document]
deanamiel: Let's run that by Milap if it should support multiple payloads, I will update the readme with the example above in the meantime

evm/README.md
deanamiel: When I generated the signatures they looked like the example below, so it was still necessary to append `0x`, do you see any signatures in different formats?
```
{
  "key_id": "evm-ethereum-2-3799882",
  "msg_hash": "9dd126c2fc8b8a...",
  "pub_key": "02884f2544...",
  "signature": "976da9bb8...",
  "validator": "axelarvaloper12yyp..."
}
```

evm/README.md
```diff
@@ -181,3 +181,72 @@ To get details of options provided in the command run:
 ```bash
 node evm/verify-contract.js --help
 ```
+
+## Combine Validator Signatures
+
+The following script is for emergency situations where we need to ask validators to manually sign gateway approvals. These validator signatures must be collected and stored within separate files in the signatures folder in the root directory. The script will then combine these signatures into a gateway batch and optionally execute this batch on the Axelar gateway contract on the specified chain.
+
```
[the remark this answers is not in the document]
deanamiel: I will mention that for stagenet the lcd endpoint must be manually entered in the chains config

evm/combine-signatures.js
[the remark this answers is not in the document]
deanamiel: added a console log

evm/combine-signatures.js
deanamiel: This line is for the next iteration of the loop, it reassigns `prevKeyId` to the current `keyId` so it can be compared to the next `keyId` and ensure that all keyIds match

evm/README.md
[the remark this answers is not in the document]
blockchainguyy: @deanamiel were you able to successfully update a test gateway as well?

evm/README.md
blockchainguyy: yep for me signatures look like this:

```json
{
  "key_id": "evm-fantom-3784501",
  "msg_hash": "0xd66d9...",
  "pub_key": "0x02407e...",
  "signature": "0xc072da18731dd498...",
  "validator": "axelarvaloper1n30vm3rym25kqrs3rpql..."
}
```

The command I used:
`axelard vald-sign evm-fantom-3784501 axelarvaloper1n30vm3rym25kqrs3rpql0m78wst5fgm
6zvr77h 0xd66d9988d2e618f2dc025bb7ae42b6e27a4f249e3275a6033026953c3fe3869c`

@milapsheth can diff. chains return a diff revert for axelard?

blockchainguyy: Anyway @deanamiel we should be prepared for both formats and add `0x` only if it is not present.

evm/combine-signatures.js
deanamiel: Ah yeah that's true

evm/README.md
deanamiel: Weird, yeah I can update the logic to support both formats then

evm/README.md
```diff
@@ -187,3 +187,97 @@ To get details of options provided in the command run:
...
+    "rpc": "",
+    "lcd": "",
+    "grpc": "",
+    "tokenSymbol": "AXL"
+  }
+```
+
+Below are instructions on how to utilize this script:
+
+1. Construct the gateway batch data by passing in the payload hash gotten from running the `governance.js` script.
```
blockchainguyy, line 210: Checked `governance.js` we are not encoding data in the format expected by `execute()`, check: 
- https://github.com/axelarnetwork/axelar-contract-deployments/blob/cc6c0ca5284ff17adf1b9116422f43833450745b/evm/governance.js#L192
- https://github.com/axelarnetwork/axelar-cgp-solidity/blob/968c8964061f594c80dd111887edb93c5069e51e/contracts/AxelarGateway.sol#L487C69-L487C106

blockchainguyy: So how can we get payload hash from `governance.js`? @milapsheth @deanamiel

PR #156, opened 2024-01-23
fix: make `getAmplifierKeyAddresses` compatible with query response

jcs47:


evm/utils.js
```diff
@@ -651,7 +651,12 @@ const getAmplifierKeyAddresses = async (config, chain) => {
     const client = await CosmWasmClient.connect(config.axelar.rpc);
     const workerSet = await client.queryContractSmart(config.axelar.contracts.MultisigProver[chain].address, 'get_worker_set');
 
-    const weightedAddresses = workerSet.signers.map((signer) => ({
```
milapsheth, line 654: Map over Object.values instead directly?

milapsheth (approved):

PR #162, opened 2024-01-30
feat: add script to verify contracts on Sourcify and integrate with current verify-contract script[AXE-2812]

blockchainguyy:
The idea is to verify contracts on Sourcify automatically when verifying on other explorers, this script will try and verify on Sourcify using source code and metadata, if data dir is present, otherwise will verify on Sourcify using explorers like Etherscan.

### DIFF
--- a/axelar-chains-config/info/testnet.json
+++ b/axelar-chains-config/info/testnet.json
@@ -2025,6 +2025,39 @@
         "name": "",
         "url": ""
       }
+    },
+    "optimism-sepolia": {
+      "name": "Optimism-Sepolia",
+      "axelarId": "optimism-sepolia",
+      "chainId": 11155420,
+      "rpc": "https://sepolia.optimism.io",
+      "tokenSymbol": "ETH",
+      "gasOptions": {
+        "gasLimit": 8000000
+      },
+      "confirmations": 3,
+      "explorer": {
+        "name": "Opscan",
+        "url": "https://sepolia-optimistic.etherscan.io",
+        "api": "https://api-sepolia-optimistic.etherscan.io/api"
+      }
+    },
+    "base-sepolia": {
+      "name": "Base Sepolia",
+      "id": "base-sepolia",
+      "axelarId": "base-sepolia",
+      "chainId": 84532,
+      "rpc": "https://sepolia.base.org",
+      "tokenSymbol": "ETH",
+      "gasOptions": {
+        "gasLimit": 3000000
+      },
+      "confirmations": 2,
+      "explorer": {
+        "name": "Basescan",
+        "url": "https://sepolia-explorer.base.org",
+        "api": "https://api-sepolia.basescan.org/api"
+      }
     }
   },
   "axelar": {


axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+
+async function readAndParseFile(filePath, targetKey) {
+    try {
+        const jsonData = await fs.promises.readFile(filePath, { encoding: 'utf8' });
+        const parsedData = JSON.parse(jsonData);
+        const result = [];
+        await findValue(parsedData, targetKey, result);
+        return result;
+    } catch (error) {
+        console.error(`Error reading or parsing file ${filePath}: ${error.message}`);
```
deanamiel, line 28:
```suggestion
        printError(`Error reading or parsing file ${filePath}`, error.message);
```

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+}
+
+async function processDirectory(directoryPath, targetKey) {
+    try {
+        const fileNames = await fs.promises.readdir(directoryPath);
+        const filePaths = fileNames.map((fileName) => path.join(directoryPath, fileName));
+
+        const results = await Promise.all(filePaths.map((filePath) => readAndParseFile(filePath, targetKey)));
+
+        const uniqueResults = [...new Set(results.flat().filter(Boolean))];
```
deanamiel, line 40:
```suggestion
        const uniqueResults = [...(new Set(results.flat().filter(Boolean)))];
```

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+        return uniqueResults;
+    } catch (error) {
+        console.error(`Error reading directory ${directoryPath}: ${error.message}`);
```
deanamiel, line 43:
```suggestion
        printError(`Error reading directory ${directoryPath}`, error.message);
```

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+        const fullPath = path.join(rootFolder, fileOrFolder);
+
+        if (fs.statSync(fullPath).isDirectory()) {
+            const contractName = searchContractName(fullPath, targetValue);
+
+            if (contractName) {
+                return contractName;
+            }
+        } else if (fileOrFolder.endsWith('.json')) {
+            const jsonData = JSON.parse(fs.readFileSync(fullPath, 'utf-8'));
```
deanamiel, line 61: Wrap `JSON.parse` in a try catch

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+
+async function verify(dir, provider, address, chainId) {
+    try {
+        const bytecode = await provider.getCode(address);
+
+        if (!dir) {
+            throw new Error('Need to specify directory to verify on sourcify via solidity code');
+        }
+
+        const artifactsDir = path.join(__dirname, `../../../${dir}/artifacts/contracts`);
```
deanamiel, line 78: Can this path be made less hardcoded?

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+        if (!contractName) throw new Error('Byte code match not found');
+
+        const directoryPath = artifactsDir.substring(0, artifactsDir.length - 'contracts'.length) + 'build-info';
+        const res = await processDirectory(directoryPath, contractName);
+
+        if (res.length !== 2) throw new Error('Unable to find require metadata and solidity code for sourcify verification');
+        const [sol, metadata] = res;
+
+        await uploadToSourcify(metadata, sol, address, chainId);
+        console.log('Verified on Sourcify via uplaoding solidity code and metadata');
```
deanamiel, line 90: In general should use `printInfo`, `printWarn`, and `printError` to keep consistent with other files in this repository

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+        files: {
+            'metadata.json': metadata,
+            'Source.sol': sol,
+        },
+    };
+
+    const payload = JSON.stringify(payloadObj);
+    const apiUrl = 'https://sourcify.dev/server/verify';
+    const res = await axios.post(apiUrl, payload, { headers });
+    assert(res.data.result[0].status === 'perfect');
```
deanamiel, line 120: Won't `assert` throw an error if the condition is not met? Should it be wrapped in a try catch or replaced with an if-statement so the function fails more gracefully with a more informative error message?

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+            const jsonData = JSON.parse(fs.readFileSync(fullPath, 'utf-8'));
+
+            if (jsonData.deployedBytecode === targetValue) {
+                return jsonData.contractName;
+            }
+        }
+    }
+}
+
+async function verify(dir, provider, address, chainId) {
```
milapsheth, line 70:
```suggestion
async function verifySourcify(dir, provider, address, chainId) {
```
or something similar to distinguish it from other verify methods

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            await findValue(obj[key], contractName, result);
+        }
+    }
+}
+
+async function readAndParseFile(filePath, targetKey) {
+    try {
+        const jsonData = await fs.promises.readFile(filePath, { encoding: 'utf8' });
```
milapsheth, line 22: we have readJSON helper somewhere in utils I think

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+                result.push(obj[key].metadata ? obj[key].metadata : obj[key].content);
+            }
+
+            await findValue(obj[key], contractName, result);
+        }
+    }
+}
+
+async function readAndParseFile(filePath, targetKey) {
+    try {
```
milapsheth, line 21: avoid adding try/catch in helper methods. In most cases, the caller should decide how to handle an error, or ensure the method shouldn't error, so we don't continue on error

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+
+            if (jsonData.deployedBytecode === targetValue) {
+                return jsonData.contractName;
+            }
+        }
+    }
+}
+
+async function verify(dir, provider, address, chainId) {
+    try {
```
milapsheth, line 71: let the caller catch the error depending on the context

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+        const bytecode = await provider.getCode(address);
```
milapsheth, line 72: `getBytecodeHash`, reuse existing utils

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+        if (res.length !== 2) throw new Error('Unable to find require metadata and solidity code for sourcify verification');
+        const [sol, metadata] = res;
+
+        await uploadToSourcify(metadata, sol, address, chainId);
+        console.log('Verified on Sourcify via uplaoding solidity code and metadata');
+    } catch (error) {
+        console.error('Unable to verify on Sourcify by uploading solidity code and metadata: ', error.message);
+
+        try {
+            await verifyFromEtherscan(chainId, address);
```
milapsheth, line 95: only need to catch the error for uploading to sourcify and do this as a fallback. But if this fails, we should propagate the failure to the caller

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -29,6 +30,8 @@ const verifyContract = (env, chain, contract, args, options = {}) => {
 
     execSync(cmd, { stdio: 'inherit' });
 
+    await verify(options.dir, options.provider, contract, options.chainId.toString());
```
milapsheth, line 33: catch errors here

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -12,7 +13,7 @@ const { writeFileSync } = require('fs');
  * @param {any[]} args
  * @returns {void}
  */
-const verifyContract = (env, chain, contract, args, options = {}) => {
+const verifyContract = async (env, chain, contract, args, options = {}) => {
```
milapsheth, line 16: check if all usages of verifyContract have `await`

evm/verify-contract.js
```diff
@@ -18,7 +18,7 @@ async function processCommand(config, chain, options) {
     const { env, contractName, dir } = options;
     const provider = getDefaultProvider(chain.rpc);
     const wallet = Wallet.createRandom().connect(provider);
-    const verifyOptions = {};
+    const verifyOptions = { chainId: chain.chainId, provider };
```
milapsheth, line 21: just pass the `chain` object to verifyContract instead of duplicating it with options

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,140 @@
...
+
+        const uniqueResults = [...new Set(results.flat().filter(Boolean))];
+        return uniqueResults;
+    } catch (error) {
+        console.error(`Error reading directory ${directoryPath}: ${error.message}`);
+        return [];
+    }
+}
+
+function searchContractName(rootFolder, targetValue) {
```
milapsheth, line 48: Have you looked at `getContractJSON`?

milapsheth, line 48: could also modify getContractPath to accept bytecode instead of contractName and check if it starts with 0x

axelar-chains-config/src/utils/sourcify.js
blockchainguyy: Using printError or printInfo in axelar-chains-config introduces circular dependency problem, should I declare it again in axelar-chains-config?
Also I think it is better to just use the normal console as done in other place like [here](https://github.com/axelarnetwork/axelar-contract-deployments/blob/dc8e9d6acbd6e26938006f5d273ca178fc14fa50/axelar-chains-config/src/utils/verifyContract.js#L35) .

axelar-chains-config/src/utils/sourcify.js
blockchainguyy: Using printError or printInfo in axelar-chains-config introduces circular dependency problem, should I declare it again in axelar-chains-config?
Also I think it is better to just use the normal console as done in other place like [here](https://github.com/axelarnetwork/axelar-contract-deployments/blob/dc8e9d6acbd6e26938006f5d273ca178fc14fa50/axelar-chains-config/src/utils/verifyContract.js#L35) .

axelar-chains-config/src/utils/sourcify.js
blockchainguyy: same case of circular import.

axelar-chains-config/src/utils/sourcify.js
blockchainguyy: Its a method for fetching bytecode hash, and bytecode is needed here.

axelar-chains-config/src/utils/sourcify.js
blockchainguyy: Milap advised against it here: https://github.com/axelarnetwork/axelar-contract-deployments/pull/162#discussion_r1472814043

axelar-chains-config/src/utils/sourcify.js
blockchainguyy: prettier removes these changes.

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,161 @@
...
+        address,
+        chainId,
+    };
+
+    const apiUrl = 'https://sourcify.dev/server/verify/etherscan';
+    const res = await axios.post(apiUrl, payload, { headers });
+
+    switch (res.data.result[0].status) {
+        case 'perfect':
+            console.log(`Sourcify: contract verified perfectly., ${address}`);
```
deanamiel, line 151:
```suggestion
            console.log(`Sourcify: contract verified perfectly at address ${address}`);
```

axelar-chains-config/src/utils/sourcify.js
```diff
@@ -0,0 +1,161 @@
...
+            break;
+        case 'partial':
+            console.log(`Sourcify: contract verified partially., ${address}`);
```
deanamiel, line 154:
```suggestion
            console.log(`Sourcify: contract verified partially at address ${address}`);
```

deanamiel (approved):

re1ro (approved):

PR #163, opened 2024-02-04
feat: fraxtal testnet configurations

canhtrinh:


axelar-chains-config/info/testnet.json
```diff
@@ -2094,6 +2094,92 @@
         "url": ""
       }
     },
+    "fraxtal": {
+      "name": "Fraxtal",
+      "id": "fraxtal",
+      "axelarId": "fraxtal",
+      "chainId": 2522,
+      "rpc": "https://rpc.testnet.frax.com",
+      "tokenSymbol": "frxETH",
```
blockchainguyy, line 2103: We should also add confirmations too. 2 maybe?

milapsheth (approved):

PR #164, opened 2024-02-07
fix: remove remaining instances of id

deanamiel:
Remove lingering instances of `chain.id` to `chain.axelarId`

### DIFF
--- a/cosmwasm/deploy-contract.js
+++ b/cosmwasm/deploy-contract.js
@@ -331,8 +331,8 @@ const upload = (client, wallet, chainName, config, options) => {
                 }
 
                 if (chainConfig) {
-                    contractConfig[chainConfig.id] = {
-                        ...contractConfig[chainConfig.id],
+                    contractConfig[chainConfig.axelarId] = {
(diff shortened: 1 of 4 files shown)


cosmwasm/deploy-contract.js
```diff
@@ -360,8 +360,8 @@ const instantiate = (client, wallet, chainName, config, options) => {
     const initMsg = makeInstantiateMsg(contractName, chainName, config);
     return instantiateContract(client, wallet, initMsg, config, options).then((contractAddress) => {
         if (chainConfig) {
-            contractConfig[chainConfig.id] = {
-                ...contractConfig[chainConfig.id],
+            contractConfig[chainConfig.axelarId] = {
```
milapsheth, line 363: @jcs47 do we refer to `id` in the wasm deployment action? can we change it to axelarId there as well if so?

jcs47: IIRC I already fixed something related to this in the orchestration. But I am not sure if this PR will break anything else. I'll keep an eye on the pipelines.

jcs47 (approved):

PR #170, opened 2024-02-15
feat: add polygon zkEVM cardona testnet config

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-3155

### DIFF
--- a/axelar-chains-config/info/testnet.json
+++ b/axelar-chains-config/info/testnet.json
@@ -2126,6 +2126,23 @@
         "url": "https://sepolia-explorer.base.org",
         "api": "https://api-sepolia.basescan.org/api"
       }
+    },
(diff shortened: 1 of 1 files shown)


axelar-chains-config/info/testnet.json
```diff
@@ -2126,6 +2126,23 @@
         "url": "https://sepolia-explorer.base.org",
         "api": "https://api-sepolia.basescan.org/api"
       }
+    },
+    "polygon-cardona": {
+      "name": "Polygon ZKEVM Cardona",
+      "id": "Polygon-zkevm-cardona",
```
milapsheth, line 2132: axelar registered chain id should be lower case

milapsheth:
not adding support for now

PR #173, opened 2024-02-20
feat: add fraxtal contracts

milapsheth:


milapsheth:
merging without ITS deployment to fix axelar config generation

PR #180, opened 2024-02-29
feat(info): configuration for devnet-integrators

jcs47:


talalashraf (approved):
I don't think savings these and committing in code is a sustainable approach. We should either store in configmaps within the namespace or find some other solution for this. We will most likely have a lot more devnets 12 months from now than we have today.

PR #186, opened 2024-03-06
feat: deploy L2 testnet and blast governance and ITS contracts

milapsheth:


evm/nonces.json
```diff
@@ -147,4 +155,4 @@
       "0x027c1882B975E2cd771AE068b0389FA38B9dda73": 0
     }
   }
-}
+}
```
milapsheth, line 158:
```suggestion
}
```

PR #189, opened 2024-03-13
feat(wasm): flag for setting contract admin

jcs47:


cosmwasm/utils.js
```diff
@@ -66,8 +66,9 @@ const instantiateContract = (client, wallet, initMsg, config, { contractName, sa
                       initMsg,
                       contractName,
                       initFee,
+                      { admin: admin || account.address },
```
milapsheth, line 69: we should default the admin to the governance address if it's not overridden, not the deployer wallet

jcs47: done

cosmwasm/deploy-contract.js
```diff
@@ -425,6 +425,7 @@ const programHandler = () => {
     program.addOption(new Option('-n, --chainNames <chainNames>', 'chain names').default('none'));
     program.addOption(new Option('-r, --reuseCodeId', 'reuse code Id'));
     program.addOption(new Option('-s, --salt <salt>', 'salt for instantiate2. defaults to contract name').env('SALT'));
+    program.addOption(new Option('--admin <address>', 'when instantiating contract, set an admin address other than the creator'));
```
milapsheth, line 428:
```suggestion
    program.addOption(new Option('--admin <address>', 'when instantiating contract, set an admin address other than the creator. Defaults to governance module account').default(governanceAddress));
```
and import it here

milapsheth (approved):

cosmwasm/deploy-contract.js
jcs47: done

PR #192, opened 2024-03-20
feat(GasService): script with gas service commands

re1ro:
[AXE-3447]
* [x] `AxelarGasService`: new  script with gas service commands
* [x] Depends on release https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/133
* [x] Depends on release https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/146
* [x] Depends on release https://github.com/axelarnetwork/axelar-cgp-solidity/pull/273

[AXE-3447]: https://axelarnetwork.atlassian.net/browse/AXE-3447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/axelar-chains-config/info/mainnet.json
+++ b/axelar-chains-config/info/mainnet.json
@@ -95,22 +95,22 @@
           "salt": "Operators"
         },
         "InterchainTokenService": {
-          "salt": "ITS v1.2.1",
+          "salt": "ITS v1.2.4",
           "deployer": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
           "proxySalt": "ITS v1.0.0",
           "tokenManagerDeployer": "0x121b0e54Cd7ad2BBCb4c4C9275697978EBaF3653",
           "interchainToken": "0x7F9F70Da4af54671a6abAc58e705b5634cac8819",
           "interchainTokenDeployer": "0x58667c5f134420Bf6904C7dD01fDDcB4Fea3a760",
-          "tokenManager": "0xC1B09c9c16117417A1B414A52Dd92CF1f634e786",
-          "tokenHandler": "0x9Ef1f24EF4Ed0520DC1Da282a0fe76271A183b36",
-          "implementation": "0x30DEFb998D7FdDC75ffb8D002fC15BbB75115758",
+          "tokenManager": "0x81a0545091864617E7037171FdfcBbdCFE3aeb23",
+          "tokenHandler": "0x07715674F74c560200c7C95430673180812fCE73",
+          "implementation": "0x6d59D9360BDAe406614b7E61c53F43a03198A4ef",
           "predeployCodehash": "0x08a4a556c4db879b4f24104d13a8baf86915d58b12c81b382dfea2a82d2856cf",
           "address": "0xB5FB4BE02232B1bBA4dC8f81dc24C26980dE9e3C"
         },
         "InterchainTokenFactory": {
           "salt": "ITS Factory v1.0.0",
           "deployer": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
-          "implementation": "0x440B118f34d6224B20b4641835AC9161BD4f0994",
+          "implementation": "0xA852412D2d91fE05d790934B6E64C3C351fAB54f",
           "address": "0x83a93500d23Fbc3e82B410aD07A6a9F7A0670D66"
         }
       },
@@ -220,22 +220,22 @@
           "salt": "Operators"
         },
         "InterchainTokenService": {
-          "salt": "ITS v1.2.1",
+          "salt": "ITS v1.2.4",
           "deployer": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
           "proxySalt": "ITS v1.0.0",
           "tokenManagerDeployer": "0x121b0e54Cd7ad2BBCb4c4C9275697978EBaF3653",
           "interchainToken": "0x7F9F70Da4af54671a6abAc58e705b5634cac8819",
           "interchainTokenDeployer": "0x58667c5f134420Bf6904C7dD01fDDcB4Fea3a760",
-          "tokenManager": "0xC1B09c9c16117417A1B414A52Dd92CF1f634e786",
-          "tokenHandler": "0x9Ef1f24EF4Ed0520DC1Da282a0fe76271A183b36",
-          "implementation": "0x30DEFb998D7FdDC75ffb8D002fC15BbB75115758",
+          "tokenManager": "0x81a0545091864617E7037171FdfcBbdCFE3aeb23",
+          "tokenHandler": "0x07715674F74c560200c7C95430673180812fCE73",
+          "implementation": "0x6d59D9360BDAe406614b7E61c53F43a03198A4ef",
           "predeployCodehash": "0x08a4a556c4db879b4f24104d13a8baf86915d58b12c81b382dfea2a82d2856cf",
           "address": "0xB5FB4BE02232B1bBA4dC8f81dc24C26980dE9e3C"
         },
         "InterchainTokenFactory": {
           "salt": "ITS Factory v1.0.0",
           "deployer": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
-          "implementation": "0x440B118f34d6224B20b4641835AC9161BD4f0994",
+          "implementation": "0xA852412D2d91fE05d790934B6E64C3C351fAB54f",
           "address": "0x83a93500d23Fbc3e82B410aD07A6a9F7A0670D66"
         }
       },
@@ -342,22 +342,22 @@
           "salt": "Operators"
         },
         "InterchainTokenService": {
-          "salt": "ITS v1.2.1",
+          "salt": "ITS v1.2.4",
           "deployer": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
           "proxySalt": "ITS v1.0.0",
           "tokenManagerDeployer": "0x121b0e54Cd7ad2BBCb4c4C9275697978EBaF3653",
           "interchainToken": "0x7F9F70Da4af54671a6abAc58e705b5634cac8819",
           "interchainTokenDeployer": "0x58667c5f134420Bf6904C7dD01fDDcB4Fea3a760",
-          "tokenManager": "0xC1B09c9c16117417A1B414A52Dd92CF1f634e786",
-          "tokenHandler": "0x9Ef1f24EF4Ed0520DC1Da282a0fe76271A183b36",
-          "implementation": "0x30DEFb998D7FdDC75ffb8D002fC15BbB75115758",
+          "tokenManager": "0x81a0545091864617E7037171FdfcBbdCFE3aeb23",
+          "tokenHandler": "0x07715674F74c560200c7C95430673180812fCE73",
+          "implementation": "0x6d59D9360BDAe406614b7E61c53F43a03198A4ef",
           "predeployCodehash": "0x08a4a556c4db879b4f24104d13a8baf86915d58b12c81b382dfea2a82d2856cf",
           "address": "0xB5FB4BE02232B1bBA4dC8f81dc24C26980dE9e3C"
         },
         "InterchainTokenFactory": {
           "salt": "ITS Factory v1.0.0",
           "deployer": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
-          "implementation": "0x440B118f34d6224B20b4641835AC9161BD4f0994",
+          "implementation": "0xA852412D2d91fE05d790934B6E64C3C351fAB54f",
           "address": "0x83a93500d23Fbc3e82B410aD07A6a9F7A0670D66"
         }
       },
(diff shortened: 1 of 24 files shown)


evm/gas-service.js
```diff
@@ -0,0 +1,414 @@
...
+                            gas_price_in_units: { value: gasPrice },
+                            decimals,
+                        },
+                        execute_gas_multiplier: multiplier = 1.1,
+                    } = result;
+
+                    const axelarBaseFee = Math.ceil(parseFloat(sourceBaseFee) * Math.pow(10, decimals));
+                    const relativeGasPrice = Math.ceil(parseFloat(gasPrice) * parseFloat(multiplier));
+
+                    return [feeType, axelarBaseFee, relativeGasPrice, blobBaseFee * relativeGasPrice];
```
milapsheth, line 79:
```suggestion
                    return [feeType, axelarBaseFee, relativeGasPrice, blobBaseFee * destinationTokenPrice / srcTokenPrice];
```
blobBaseFee itself is the blob gas price, e.g. 1 gwei, so we just need to normalize it, and not multiply regular gas price/base fee

evm/gas-service.js
```diff
@@ -0,0 +1,414 @@
...
+        }
+
+        case 'updateGasInfo': {
+            validateParameters({
+                isNonEmptyStringArray: { chains },
+            });
+
+            const gasUpdates = await getGasInfo(env, chain, chains);
+
+            // Adding lowecase chain names for case insensitivity
```
milapsheth, line 176:
```suggestion
            // Adding lowercase chain names for case insensitivity
```

evm/gas-service.js
```diff
@@ -0,0 +1,414 @@
...
+
+            break;
+        }
+
+        case 'updateGasInfo': {
+            validateParameters({
+                isNonEmptyStringArray: { chains },
+            });
+
+            const gasUpdates = await getGasInfo(env, chain, chains);
```
milapsheth, line 174: If we're running this script for multiple chains `-n`, we should cache the info and reuse the same one across all chains. You can use a global variable for now. But we should introduce a state object that we pass around between `processCommands`

evm/gas-service.js
```diff
@@ -0,0 +1,414 @@
...
+            let feeType = 0;
+            let blobBaseFee = 0;
+
+            switch (destinationChain.toLowerCase()) {
+                case 'ethereum-sepolia':
+
+                // eslint-disable-next-line no-fallthrough
+                case 'ethereum': {
+                    feeType = 0;
+                    blobBaseFee = env === 'testnet' ? 50187563959 : 1;
```
milapsheth, line 40: query this dynamically from an RPC (we can do it in a separate task)

evm/gas-service.js
```diff
@@ -0,0 +1,414 @@
...
+                    break;
+                }
+
+                case 'base':
+
+                // eslint-disable-next-line no-fallthrough
+                case 'optimism': {
+                    feeType = 1; // Optimism Ecotone
+                    blobBaseFee = 0;
```
milapsheth, line 49:
```suggestion
```
doesn't need to set again since it's 0 by default

evm/gas-service.js
```diff
@@ -0,0 +1,414 @@
...
+const { addBaseOptions } = require('./cli-utils');
+const IAxelarGasService = require('@axelar-network/axelar-gmp-sdk-solidity/interfaces/IAxelarGasService.json');
+const { getWallet } = require('./sign-utils');
+
+async function getGasInfo(env, chain, destinationChains) {
+    const apiUrl = env === 'testnet' ? 'https://testnet.api.axelarscan.io/gmp/getFees' : 'https://api.axelarscan.io/gmp/getFees';
+
+    return Promise.all(
+        destinationChains.map((destinationChain) => {
+            let feeType = 0;
```
milapsheth, line 31:
```suggestion
            let gasEstimationType = 0;
```
to match the contract name

evm/gas-service.js
```diff
@@ -0,0 +1,414 @@
...
+            let blobBaseFee = 0;
+
+            switch (destinationChain.toLowerCase()) {
+                case 'ethereum-sepolia':
```
milapsheth, line 35: L1 name might not be ethereum on the network. Push ethereum's config under the `ethereum` key as well if it wasn't done

evm/gas-service.js
re1ro: It's different info for each source chain, and a different API call.

evm/gas-service.js
re1ro: I moved this to the config, so we can have an upper boundary there.

evm/gas-service.js
re1ro: I moved everything to config, no hardcoded chain names anymore.

axelar-chains-config/info/mainnet.json
```diff
@@ -124,7 +124,10 @@
         "gasPrice": 150000000000
       },
       "finality": "finalized",
-      "approxFinalityWaitTime": 40
+      "approxFinalityWaitTime": 40,
+      "gasOptions": {
+        "blobBaseFee": 1
```
milapsheth, line 129: gasOptions are fed directly into evm txs, so unexpected values can cause it to fail. create a different field instead `onchainGasEstimate`. otherwise you'd need to delete it in the utils:
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/evm/utils.js#L1052

axelar-chains-config/info/mainnet.json
```diff
@@ -1367,7 +1370,10 @@
         "gasPrice": 10000000000
       },
       "finality": "finalized",
-      "approxFinalityWaitTime": 25
+      "approxFinalityWaitTime": 25,
+      "gasOptions": {
+        "gasEstimationType": 1
```
milapsheth, line 1375: same as above. set the corresponding L1 name as well

axelar-chains-config/info/stagenet.json
```diff
@@ -1162,6 +1168,7 @@
     "rpc": "",
     "lcd": "",
     "grpc": "",
-    "tokenSymbol": "AXL"
+    "tokenSymbol": "AXL",
+    "api": "https://testnet.api.axelarscan.io"
```
milapsheth, line 1172:
```suggestion
    "axelarscanApi": "https://testnet.api.axelarscan.io"
```

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+            const relativeGasPrice = Math.ceil(parseFloat(gasPrice) * parseFloat(multiplier));
+            const gasPriceRatio = parseFloat(destinationTokenPrice) / parseFloat(srcTokenPrice);
+            const relativeBlobBaseFee = Math.ceil(blobBaseFee * gasPriceRatio);
+
+            return [gasEstimationType, axelarBaseFee, relativeGasPrice, relativeBlobBaseFee];
+        }),
+    );
+}
+
+async function processCommand(_, chain, options) {
```
milapsheth, line 68:
```suggestion
async function processCommand(config, chain, options) {
```

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+
+            break;
+        }
+
+        case 'updateGasInfo': {
+            validateParameters({
+                isNonEmptyStringArray: { chains },
+            });
+
+            let gasUpdates = await getGasUpdates(env, chain, chains);
```
milapsheth, line 141: pass in `config` here instead of loading again

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+    isValidAddress,
+    validateParameters,
+    httpPost,
+    loadConfig,
+} = require('./utils');
+const { addBaseOptions } = require('./cli-utils');
+const { getWallet } = require('./sign-utils');
+
+async function getGasUpdates(env, chain, destinationChains) {
+    const config = loadConfig(env);
```
milapsheth, line 24: you can get this from the caller

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+
+            if (!destinationConfig) {
+                printInfo(`Error: chain ${destinationChain} not found in config.`);
+                printInfo(`Skipping ${destinationChain}.`);
+                return null;
+            }
+
+            const { gasEstimationType = 0, blobBaseFee = 0 } = destinationConfig;
+
+            const data = await httpPost(`${api}/gmp/getFees`, {
```
milapsheth, line 39: add a try/catch in case it errors out, so we don't skip other chains?

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+async function getGasUpdates(env, chain, destinationChains) {
+    const config = loadConfig(env);
+    const api = config.axelar.api;
+
+    return Promise.all(
+        destinationChains.map(async (destinationChain) => {
+            const destinationConfig = config.chains[destinationChain];
+
+            if (!destinationConfig) {
+                printInfo(`Error: chain ${destinationChain} not found in config.`);
```
milapsheth, line 32:
```suggestion
                printError(`Error: chain ${destinationChain} not found in config.`);
```

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+                printInfo(`Skipping ${destinationChain}.`);
```
milapsheth, line 33:
```suggestion
                printError(`Skipping ${destinationChain}.`);
```

evm/gas-service.js
milapsheth: ah, yes, we can still do it, but we'd have to normalize to the src token price ourselves by src chain. we can add it as a follow up optimization (my concern is that we might hit a rate limit due to n^2 API calls for n chains).

evm/gas-service.js
milapsheth: we still need to post L1 info under `ethereum` [key](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/gas/InterchainGasEstimation.sol#L61) even if ethereum's axelarId is not `ethereum`

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        default:
+            throw new Error(`Unknown action: ${action}`);
+    }
+}
+
+async function main(options) {
+    await mainProcessor(options, processCommand, false);
+}
```
milapsheth, line 177: if any gasUpdates were null for any chains, throw an error here with the src chain/destination chain combination, so the caller knows the script failed and can alert

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+                printError(`Skipping ${destinationChain}.`);
+                return null;
+            }
+
+            const { gasEstimationType = 0, blobBaseFee = 0 } = destinationConfig;
+
+            const data = await httpPost(`${api}/gmp/getFees`, {
+                sourceChain: chain.axelarId,
+                destinationChain,
+                sourceTokenAddress: '0x0000000000000000000000000000000000000000',
```
deanamiel, line 42:
```suggestion
                sourceTokenAddress: ethers.utils.AddressZero,
```

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+                    token_price: { usd: srcTokenPrice },
+                    decimals,
+                },
+                destination_native_token: {
+                    token_price: { usd: destinationTokenPrice },
+                },
+                execute_gas_multiplier: multiplier = 1.1,
+            } = data.result;
+
+            const axelarBaseFee = Math.ceil(parseFloat(sourceBaseFee) * Math.pow(10, decimals));
```
deanamiel, line 58: Are you sure you don't need to use BigNumbers here? (18 decimals)

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+    if (prompt(`Proceed with action ${action} on chain ${chain.name}?`, yes)) {
+        return;
+    }
+
+    switch (action) {
+        case 'estimateGasFee': {
+            validateParameters({
+                isNonEmptyString: { destinationChain },
+                isValidAddress: { destinationAddress },
+                isNumber: { executionGasLimit },
```
deanamiel, line 124: Is this expected to be passed in as wei or eth? If eth then it needs to be converted to wei when calling the gas service right?

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+    program.addOption(
+        new Option('--action <action>', 'GasService action').choices(['estimateGasFee', 'updateGasInfo']).makeOptionMandatory(true),
+    );
+    program.addOption(new Option('--offline', 'run script in offline mode'));
+    program.addOption(new Option('--nonceOffset <nonceOffset>', 'The value to add in local nonce if it deviates from actual wallet nonce'));
+
+    // options for estimateGasFee
+    program.addOption(new Option('--destinationChain <destinationChain>', 'Destination chain name').makeOptionMandatory(false));
+    program.addOption(new Option('--destinationAddress <destinationAddress>', 'Destination contract address').makeOptionMandatory(false));
+    program.addOption(new Option('--payload <payload>', 'Payload for the contract call').makeOptionMandatory(false));
```
deanamiel, line 198: Sometimes helpful to add the env option for long payloads

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+
+if (require.main === module) {
+    const program = new Command();
+
+    program.name('GasService').description('Script to manage GasService actions');
+
+    addBaseOptions(program, { address: true });
+
+    program.addOption(
+        new Option('-c, --contractName <contractName>', 'contract name').default('AxelarGasService').makeOptionMandatory(false),
```
deanamiel, line 187:
```suggestion
        new Option('-c, --contractName <contractName>', 'contract name').default('AxelarGasService'),
```
I think providing a default implies that it is not mandatory

evm/gas-service.js
```diff
@@ -0,0 +1,211 @@
...
+    wasEventEmitted,
+    isValidAddress,
+    validateParameters,
+    httpPost,
+    loadConfig,
+} = require('./utils');
+const { addBaseOptions } = require('./cli-utils');
+const { getWallet } = require('./sign-utils');
+
+async function getGasUpdates(env, chain, destinationChains) {
```
deanamiel, line 23: Should probably add this check here as well since you export this function to be used in other files:
```
validateParameters({
                isNonEmptyStringArray: { destinationChains },
});
```

axelar-chains-config/info/mainnet.json
re1ro: Added canonical chain names to the `onchainGasEstimate`

evm/gas-service.js
re1ro: It's in gas units

evm/operators.js
```diff
@@ -229,6 +242,49 @@ async function processCommand(_, chain, options) {
...
+                if (chainName.toLowerCase() !== chainName) {
+                    gasUpdates.push({
+                        chainName: chainName.toLowerCase(),
+                        gasInfo,
+                    });
+                }
+            });
+
+            const filteredChains = gasUpdates.map(({ chainName }) => chainName);
+            const gasInfoUpdates = gasUpdates.map(({ gasInfo }) => gasInfo);
```
milapsheth, line 268: move this into the shared util? it can return the chains and gasInfo to be used for the update

axelar-chains-config/info/mainnet.json
```diff
@@ -1367,7 +1370,10 @@
         "gasPrice": 10000000000
       },
       "finality": "finalized",
-      "approxFinalityWaitTime": 25
+      "approxFinalityWaitTime": 25,
+      "onchainGasEstimate": {
+        "gasEstimationType": 1
```
milapsheth, line 1375: add L1 chain name here

evm/gas-service.js
```diff
@@ -0,0 +1,252 @@
...
+                .round();
+            // destinationTokenPrice / srcTokenPrice
+            const gasPriceRatio = FixedNumber.from(parseFloat(destinationTokenPrice).toFixed(10)).divUnsafe(
+                FixedNumber.from(parseFloat(srcTokenPrice).toFixed(10)),
+            );
+            // blobBaseFee * gasPriceRatio
+            const relativeBlobBaseFee = FixedNumber.from(parseFloat(blobBaseFee)).mulUnsafe(gasPriceRatio);
+
+            return {
+                chainName,
```
milapsheth, line 95: it needs to be stored under both the `axelarId` and `ethereum`, since if the destination is ethereum, then `axelarId` version should use. But for L2s `ethereum` key is used

axelar-chains-config/info/mainnet.json
re1ro: @milapsheth should we also make an api call for the L1 chain for the right base fee?
Right now blob base fee is calculated via source/destination price ratio, not L1.

re1ro: Disregard most L2s have Ether as gas token.

evm/gas-service.js
re1ro: I was able to implement it with FixedNumber instead, but they have lower precision only of 10 decimals.
I think it's actually better to use floats for the calculation. And rounded float is converted to BN automatically.

milapsheth (approved):

PR #202, opened 2024-03-28
ci(GasService): update GasInfo every 2 hours

re1ro:
* [x] CI: adding a job to update gas info on testnet every 2 hours

### DIFF
--- a/.github/workflows/update-gas-info.yaml
+++ b/.github/workflows/update-gas-info.yaml
@@ -52,4 +52,4 @@ jobs:
           echo "ENV=$ENV" >> .env
           echo "CHAINS=all" >> .env
 
-          node evm/operators.js --action updateGasInfo -y --chains $(echo $(jq -r '.chains | keys | join(" ")' axelar-chains-config/info/$ENV.json))
+          node evm/operators.js --action updateGasInfo --chains all -y
--- a/.github/workflows/update-gas-info.yaml
+++ b/.github/workflows/update-gas-info.yaml
@@ -52,4 +52,4 @@ jobs:
           echo "ENV=$ENV" >> .env
           echo "CHAINS=all" >> .env
 
-          node evm/operators.js --action updateGasInfo --chains $(echo $(jq -r '.chains | keys | join(" ")' axelar-chains-config/info/$ENV.json))
+          node evm/operators.js --action updateGasInfo -y --chains $(echo $(jq -r '.chains | keys | join(" ")' axelar-chains-config/info/$ENV.json))
--- a/.github/workflows/update-gas-info.yaml
+++ b/.github/workflows/update-gas-info.yaml
@@ -1,6 +1,8 @@
 name: 'Update Gas Info on Testnet'
 
 on:
+  pull_request: #delete before merge
+
   workflow_dispatch:
     inputs:
       environment:
@@ -28,10 +30,10 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
       - run: |
-          if [ "${{ github.event_name }}" == "schedule" ]; then
-            ENV="testnet"
-          else
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
             ENV="${{ github.event.inputs.environment }}"
+          else
+            ENV="testnet"
           fi
 
           case "$ENV" in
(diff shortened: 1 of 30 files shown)


.github/workflows/update-gas-info.yaml
```diff
@@ -0,0 +1,53 @@
...
+            stagenet)
+              PRIVATE_KEY="${{ secrets.STAGENET_PRIVATE_KEY }}"
+              ;;
+          esac
+
+          echo "PRIVATE_KEY=$PRIVATE_KEY" >> .env
+          echo "ENV=$ENV" >> .env
+          echo "CHAINS=all" >> .env
+
+          node evm/operators.js --action updateGasInfo --chains $(echo $(jq -r '.chains | keys | join(" ")' axelar-chains-config/info/$ENV.json))
```
milapsheth: support "all" for `--chains`

axelar-chains-config/info/testnet.json
```diff
@@ -1010,6 +1011,9 @@
       "gasOptions": {
         "gasLimit": 300000000
       },
+      "onchainGasEstimate": {
+        "chainName": "filecoin"
```
milapsheth, line 1015: use `axelarId` field for the chain name instead of adding another one

.github/workflows/update-gas-info.yaml
```diff
@@ -0,0 +1,53 @@
...
+        default: testnet
+        type: choice
+        options:
+          - testnet
+          - mainnet
+          - stagenet
+
+  schedule:
+    - cron: '0 */2 * * *'
+
```
milapsheth, line 18:
```suggestion
```
thinking of removing the cron for now until we add some optimizations and make it a bit more reliable. we can update manually until then if needed

axelar-chains-config/info/testnet.json
```diff
@@ -1981,6 +2007,11 @@
       "gasOptions": {
         "gasLimit": 50000000000
       },
+      "onchainGasEstimate": {
+        "chainName": "mantle",
+        "l1ChainName": "ethereum",
+        "gasEstimationType": 1
```
milapsheth, line 2013: change the estimation type

axelar-chains-config/info/testnet.json
```diff
@@ -1872,6 +1893,11 @@
       "gasOptions": {
         "gasLimit": 8000000
       },
+      "onchainGasEstimate": {
+        "chainName": "blast",
+        "l1ChainName": "ethereum",
+        "gasEstimationType": 1
```
milapsheth, line 1899: which params are appropriate for blast?

axelar-chains-config/info/testnet.json
```diff
@@ -1872,6 +1893,11 @@
       "gasOptions": {
         "gasLimit": 8000000
       },
+      "onchainGasEstimate": {
```
milapsheth, line 1896: also check for fraxtal

axelar-chains-config/info/testnet.json
re1ro: But `axelarId` is `filecoin-2`, so wanted to have a canonical name there too.

axelar-chains-config/info/testnet.json
re1ro: We haven't released/deployed the arbitrum estimation yet.

axelar-chains-config/info/testnet.json
re1ro: We don't have the bedrock implemented in there, so just using OP as an upper bound.

axelar-chains-config/info/testnet.json
re1ro: Yeah it's still using bedrock, but we haven't deployed it yet.

axelar-chains-config/info/testnet.json
milapsheth: that refers to a different testnet. The gateway requires apps to use the axelar registered Id otherwise it'll not be recognized, they should pass the same value to the gas service

milapsheth (approved):

axelar-chains-config/info/testnet.json
re1ro: gotcha, will remove that

PR #203, opened 2024-04-01
feat: add flag to execute scripts in parallel w.r.t. chains

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-3324
https://axelarnetwork.atlassian.net/browse/AXE-4177

### DIFF
--- a/cosmwasm/README.md
+++ b/cosmwasm/README.md
@@ -41,6 +41,7 @@ This folder contains deployment scripts for cosmwasm contracts needed for amplif
       },
       "VotingVerifier": {
         "ethereum-2": {
+          "governanceAddress": "axelar10d07y265gmmuvt4z0w9aw880jnsr700j7v9daj",
           "serviceName": "validators",
           "sourceGatewayAddress": "0xe432150cce91c13a887f7D836923d5597adD8E31",
           "votingThreshold": [
@@ -51,6 +52,7 @@ This folder contains deployment scripts for cosmwasm contracts needed for amplif
           "confirmationHeight": 1
         },
         "Avalanche": {
+          "governanceAddress": "axelar10d07y265gmmuvt4z0w9aw880jnsr700j7v9daj",
           "serviceName": "validators",
           "sourceGatewayAddress": "0xe432150cce91c13a887f7D836923d5597adD8E31",
           "votingThreshold": [
--- a/cosmwasm/deploy-contract.js
+++ b/cosmwasm/deploy-contract.js
@@ -85,7 +85,7 @@ const makeVotingVerifierInstantiateMsg = (
     { id: chainId },
 ) => {
     const {
-        [chainId]: { serviceName, sourceGatewayAddress, votingThreshold, blockExpiry, confirmationHeight },
+        [chainId]: { governanceAddress, serviceName, sourceGatewayAddress, votingThreshold, blockExpiry, confirmationHeight },
     } = contractConfig;
 
     if (!validateAddress(serviceRegistryAddress)) {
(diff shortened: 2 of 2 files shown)


evm/utils.js
```diff
@@ -852,6 +861,76 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
...
+                        successfullChains.push(chainName);
+                        printInfo(`Finished running script for chain`, chainName);
+                    }
+
+                    resolve();
+                });
+            });
+        };
+
+        await (async () => {
```
re1ro, line 918: Maybe just?
```
await Promise.all(chains.map(executeChain));
```

re1ro (approved):

evm/utils.js
```diff
@@ -668,10 +669,18 @@ function loadConfig(env) {
...
 
+function loadParallelExecutionConfig(env, chain) {
+    return require(`${__dirname}/../chains-info/${env}-${chain}.json`);
+}
+
 function saveConfig(config, env) {
     writeJSON(config, `${__dirname}/../axelar-chains-config/info/${env}.json`);
 }
 
+function saveParallelExecutionConfig(config, env, chain) {
```
milapsheth, line 680: Note that the implementation of this function is not tied to a notion of parallel execution, so the naming isn't quite right. Furthermore, it's pretty much the same as existing functions. So instead of a new helper, generalize `loadConfig`/`saveConfig` to take a `filename`. It can default to null meaning the current setting, but otherwise allows customizing the name. Using a different info directory is confusing as well since it's still saving the config.

evm/utils.js
```diff
@@ -852,6 +861,73 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
         }
     }
 
+    if (options.parallel && chains.length > 1) {
```
milapsheth, line 864: Please avoid making functions so complex. Use clear and concise helper functions to organize the code

evm/utils.js
```diff
@@ -852,6 +861,73 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
...
+        const cmds = process.argv.filter((command) => command);
+        let chainCommandIndex = -1;
+        let skipPrompt = false;
+
+        for (let commandIndex = 0; commandIndex < cmds.length; commandIndex++) {
```
milapsheth, line 869: this is a weird way to run it parallely (like what you'd need to do in bash). Javascript supports `async` natively, so you can just run `processCommand` without an `await` for each chain and wait at the end.

evm/utils.js
```diff
@@ -852,6 +861,73 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
...
+            for (const chainName of successfullChains) {
+                config.chains[chainName.toLowerCase()] = loadParallelExecutionConfig(options.env, chainName);
+            }
+
+            saveConfig(config, options.env);
+        }
+
+        return;
+    }
+
```
milapsheth, line 930: i don't see a need for separate code like above to handle parallel execution. Create a chainProcessor function that handles most of the `for` loop body. Depending on sequential/parallel mode, it can be run with an await or the promise can be collected, and we await at the end.

To deal with configs in parallel mode, you can give a copy of it to each call so they don't write to the same object. And then you can merge the chain specific changes at the end. No need to save configs separately.

milapsheth: An initial complex solution is usually not the right one, I recommend spending more time to come up with a more elegant approach and discuss it during syncs if needed, otherwise codebase can easily become unmaintainable.

evm/utils.js
milapsheth: I imagine it was to separate stdout capture between parallel runs. I'm wondering if we can capture the console.log separately or reassign .log field to something that stores the output in a list so we can output, or use a process library to do the above more cleanly.

PR #207, opened 2024-04-09
chore: read env variables

blockchainguyy:
Add `require('dotenv').config();` to be able to use env in script.

### DIFF
--- a/evm/sign-message.js
+++ b/evm/sign-message.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { ethers } = require('hardhat');
+require('dotenv').config();
 const {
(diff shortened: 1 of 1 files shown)


kalidax (approved):

evm/sign-message.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const { ethers } = require('hardhat');
+require('dotenv').config();
```
milapsheth, line 4: Import `cli-utils` instead which does this already. you can use it to add the private key option as well

milapsheth (changes requested):

PR #208, opened 2024-04-11
feat(GasService): caching GMP fee results

re1ro:
* [x] GasService: caching requests to AxelarScan API

### DIFF
--- a/evm/gas-service.js
+++ b/evm/gas-service.js
@@ -119,7 +119,7 @@ async function getGasUpdates(config, env, chain, destinationChains) {
             const expressFee = parseFloat(expressFeeUsd) / parseFloat(srcTokenPrice);
             const gasPriceRatio = parseFloat(destTokenPrice) / parseFloat(srcTokenPrice);
             const relativeGasPrice =
-                parseFloat(multiplier) * parseFloat(estimateMultiplier) * (parseFloat(gasPriceInDestToken) / gasPriceRatio);
+                parseFloat(multiplier) * parseFloat(onchainGasVolatilityMultiplier) * (parseFloat(gasPriceInDestToken) / gasPriceRatio);
             const relativeBlobBaseFee = blobBaseFee * gasPriceRatio;
 
             return {
--- a/evm/gas-service.js
+++ b/evm/gas-service.js
@@ -45,6 +45,22 @@ function printFailedChainUpdates() {
     }
 }
 
+const feesCache = {};
+
+async function getFeeData(api, chain) {
+    if (!feesCache[chain]) {
+        const data = await httpPost(`${api}/gmp/getFees`, {
+            sourceChain: 'ethereum',
+            destinationChain: chain,
+            sourceTokenAddress: AddressZero,
+        });
+
+        feesCache[chain] = data.result;
+    }
+
+    return feesCache[chain];
+}
+
 async function getGasUpdates(config, env, chain, destinationChains) {
     const api = config.axelar.axelarscanApi;
 
@@ -70,14 +86,12 @@ async function getGasUpdates(config, env, chain, destinationChains) {
             const { axelarId, onchainGasEstimate: { gasEstimationType = 0, blobBaseFee = 0, multiplier: estimateMultiplier = 1.1 } = {} } =
                 destinationConfig;
 
-            let data;
+            let destinationFeeData;
+            let sourceFeeData;
 
             try {
-                data = await httpPost(`${api}/gmp/getFees`, {
-                    sourceChain: chain.axelarId,
-                    destinationChain: axelarId,
-                    sourceTokenAddress: AddressZero,
-                });
+                destinationFeeData = await getFeeData(api, axelarId);
+                sourceFeeData = await getFeeData(api, chain.axelarId);
             } catch (e) {
                 printError(`Error getting gas info for ${chain.axelarId} -> ${axelarId}`);
                 printError(e);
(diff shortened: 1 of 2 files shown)


evm/gas-service.js
```diff
@@ -45,6 +45,22 @@ function printFailedChainUpdates() {
     }
 }
 
+const feesCache = {};
+
+async function getFeeData(api, chain) {
+    if (!feesCache[chain]) {
+        const data = await httpPost(`${api}/gmp/getFees`, {
+            sourceChain: 'ethereum',
```
milapsheth, line 53: You're assuming that `ethereum` is the valid name of the chain on that network. I'd suggest just picking the first chain name from the config as the one to use as the reference.

evm/gas-service.js
```diff
@@ -70,14 +86,12 @@ async function getGasUpdates(config, env, chain, destinationChains) {
...
+            let destinationFeeData;
+            let sourceFeeData;
 
             try {
-                data = await httpPost(`${api}/gmp/getFees`, {
-                    sourceChain: chain.axelarId,
-                    destinationChain: axelarId,
-                    sourceTokenAddress: AddressZero,
-                });
+                destinationFeeData = await getFeeData(api, axelarId);
```
milapsheth, line 93:
```suggestion
                destinationFeeData = await getFeeData(api, destinationAxelarId);
```

evm/gas-service.js
```diff
@@ -86,23 +100,26 @@ async function getGasUpdates(config, env, chain, destinationChains) {
...
                 },
                 execute_gas_multiplier: multiplier = 1.1,
-            } = data.result;
+            } = destinationFeeData;
 
-            const axelarBaseFee = parseFloat(sourceBaseFee) * Math.pow(10, decimals);
-            const expressFee = parseFloat(sourceExpressFee) * Math.pow(10, decimals);
-            const relativeGasPrice = parseFloat(gasPrice) * parseFloat(multiplier) * parseFloat(estimateMultiplier);
-            const gasPriceRatio = parseFloat(destinationTokenPrice) / parseFloat(srcTokenPrice);
+            const axelarBaseFee = parseFloat(baseFeeUsd) / parseFloat(srcTokenPrice);
```
milapsheth, line 118: This value will be in terms of the source token (e.g. 10 FTM etc.). We still need to convert it to wei

Same below

evm/gas-service.js
```diff
@@ -86,23 +100,26 @@ async function getGasUpdates(config, env, chain, destinationChains) {
...
+
+            const {
+                base_fee_usd: baseFeeUsd,
+                destination_express_fee: { total_usd: expressFeeUsd },
                 destination_native_token: {
-                    token_price: { usd: destinationTokenPrice },
+                    token_price: { usd: destTokenPrice },
+                    gas_price_in_units: { value: gasPriceInDestToken },
                 },
                 execute_gas_multiplier: multiplier = 1.1,
```
milapsheth, line 115:
```suggestion
                execute_gas_multiplier: executeGasMultiplier = 1.1,
```
to better distinguish it

evm/gas-service.js
```diff
@@ -86,23 +100,26 @@ async function getGasUpdates(config, env, chain, destinationChains) {
...
 
-            const axelarBaseFee = parseFloat(sourceBaseFee) * Math.pow(10, decimals);
-            const expressFee = parseFloat(sourceExpressFee) * Math.pow(10, decimals);
-            const relativeGasPrice = parseFloat(gasPrice) * parseFloat(multiplier) * parseFloat(estimateMultiplier);
-            const gasPriceRatio = parseFloat(destinationTokenPrice) / parseFloat(srcTokenPrice);
+            const axelarBaseFee = parseFloat(baseFeeUsd) / parseFloat(srcTokenPrice);
+            const expressFee = parseFloat(expressFeeUsd) / parseFloat(srcTokenPrice);
+            const gasPriceRatio = parseFloat(destTokenPrice) / parseFloat(srcTokenPrice);
+            const relativeGasPrice =
+                parseFloat(multiplier) * parseFloat(estimateMultiplier) * (parseFloat(gasPriceInDestToken) / gasPriceRatio);
```
milapsheth, line 122:
```suggestion
                parseFloat(multiplier) * parseFloat(onchainGasVolatilityMultiplier) * (parseFloat(gasPriceInDestToken) / gasPriceRatio);
```
perhaps

evm/gas-service.js
```diff
@@ -72,6 +74,8 @@ async function getGasUpdates(config, env, chain, destinationChains) {
         destinationChains = Object.keys(config.chains);
     }
 
+    const l1Chain = Object.values(config.chains)[0];
```
milapsheth, line 77: it's not guaranteed to be the L1 chain, just the first valid chain. And we don't care if it's L1 or not, right?

evm/gas-service.js
```diff
@@ -47,18 +47,20 @@ function printFailedChainUpdates() {
 
 const feesCache = {};
 
-async function getFeeData(api, chain) {
-    if (!feesCache[chain]) {
+async function getFeeData(api, sourceChain, destinationChain) {
```
milapsheth, line 50:
```suggestion
async function getFeeData(api, referenceSourceChain, destinationChain) {
```
maybe?

milapsheth (approved):

evm/gas-service.js
re1ro: This function is not aware if it's used with reference chain or real source -> dest. It's an API client and from it's perspective it's just source and dest

evm/gas-service.js
re1ro: We don't care but kinda expecting it to be Ethereum, so the token price is more stable for conversions.

milapsheth: But we can't assume it to be the case. Also, since we're using base_fee_usd, the source chain's token market price isn't really relevant anymore.

evm/gas-service.js
milapsheth: sure

evm/gas-service.js
re1ro: Yeah the algorithm is not dependent for it to be ETH or L1 I will change the naming.
Good point about USD. I though we use ETH price for conversion as well, but that changes when I switched to USD

PR #211, opened 2024-04-15
fix: contract deployment testing script

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-3715

### DIFF
--- a/axelar-chains-config/info/devnet-amplifier.json
+++ b/axelar-chains-config/info/devnet-amplifier.json
@@ -20,21 +20,21 @@
           "startingKeyIDs": [
             null
           ],
-          "address": "0xCa85f85C72df5f8428a440887CA7c449D94e0D0c",
-          "implementation": "0xB42a3c5Bf7fDaf253c0C26Bad3227Ac91594A6D9",
-          "implementationCodehash": "0xc867f56619b2b142b2a2dfe45eb42abab7a6cc21ac2df1035f365d8b10b22789",
-          "authModule": "0x569fE174ecd2a42E9463335F1f1B2FcF760aaAD8",
-          "tokenDeployer": "0xE427A8aBB85D90E652CBd92F8DB8EF102687785D",
+          "address": "0xeA3D3979ffa953FB209FcF6D13EA49AFC0A05604",
+          "implementation": "0xfF99ED85b998bD7e2DD973cEFCb738c0C8F560EA",
+          "implementationCodehash": "0x57eb84f26ff5fab154371ae86f74219e19497e6c2c70ba1dc4116a8dc85c4eaf",
+          "authModule": "0x8d0Fbce6554ef89BBB2b200389C5e7B0Ef8Aa54b",
+          "tokenDeployer": "0xBf428fb6B2831EC30d7c098A3c3dC4d6972686d9",
           "deploymentMethod": "create"
         },
         "InterchainGovernance": {
           "governanceChain": "Axelarnet",
           "governanceAddress": "axelar10d07y265gmmuvt4z0w9aw880jnsr700j7v9daj",
           "minimumTimeDelay": 100,
-          "address": "0xbdC2f1ab89Cef197DF3061Eb4af91d2700e881ce",
+          "address": "0xd7A0e641Bfbb9AA83aAb9f3e89bf83e128d321c4",
           "deployer": "0xC5C73C3032e0577662000887D60d279B5Cc1C1AB",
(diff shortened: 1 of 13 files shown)


evm/contracts-deployment-test.js
```diff
@@ -8,19 +8,48 @@ const {
...
-async function processCommand(_, chain, options) {
+async function processCommand(config, chain, options) {
     const wallet = new Wallet(options.privateKey, new JsonRpcProvider(chain.rpc));
 
+    if (
+        !config.chains[options.chainNames].contracts.Multisig ||
+        !config.chains[options.chainNames].contracts.Multisig.signers ||
+        !config.chains[options.chainNames].contracts.Multisig.threshold
+    ) {
+        config.chains[options.chainNames].contracts.Multisig = { signers: [wallet.address], threshold: 1 };
```
milapsheth, line 22: add support for parsing `--args` for these contracts to avoid having to do this manipulation. could simplify our onboarding commands as well

README.md
```diff
@@ -3,4 +3,23 @@
 Install dependencies via
 `npm ci`
 
-[EVM deployment instructions](./evm/README.md)
\ No newline at end of file
+[EVM deployment instructions](./evm/README.md)
+
+## Scripts Usage and Example
+
+1. Test mock deployment of contracts using the `contracts-deployment-test.js` script:
```
milapsheth, line 10: Should be added to evm/README.md, see above

evm/deploy-contract.js
```diff
@@ -352,6 +353,9 @@ if (require.main === module) {
         new Option('-m, --deployMethod <deployMethod>', 'deployment method').choices(['create', 'create2', 'create3']).default('create2'),
     );
     program.addOption(new Option('--ignoreError', 'ignore errors during deployment for a given chain'));
+    program.addOption(new Option('--signers <signers>', 'signer addresses for multisig').env('SIGNERS'));
+    program.addOption(new Option('--threshold <threshold>', 'multisig threshold').env('THRESHOLD'));
+    program.addOption(new Option('--minTimeDelay <minTimeDelay>', 'minimum time delay').env('MIN_DELAY'));
```
milapsheth, line 358: since it's not scalable to have separate flag for every arg, this should be passed generically. Specifically, the flag can just be `--args`, and if provided, it can override the `contractConfig` values. E.g.
`--args '{"signers": ["0x1...", "0x2.."], "threshold": 3}'`
This can be parsed as JSON, and then used by the script and also update the config object so it gets saved locally.

evm/deploy-upgradable.js
```diff
@@ -38,21 +38,22 @@ async function getImplementationArgs(contractName, config, options) {
...
 
-            const refundIssuer = contractConfig.refundIssuer;
+            const refundIssuer = args && Array.isArray(args) && args[1] ? args[1] : contractConfig.refundIssuer;
 
             if (!isAddress(refundIssuer)) {
                 throw new Error(`${config.name} | Missing AxelarDepositService.refundIssuer in the chain info.`);
             }
 
-            const gateway = config.AxelarGateway?.address;
+            const gateway = args && Array.isArray(args) && args[2] ? args[2] : config.AxelarGateway?.address;
```
milapsheth, line 56: duplicated logic, and pretty unreadable. You can use the same approach as in the above comment.

evm/deploy-contract.js
```diff
@@ -31,6 +31,7 @@ const { addExtendedOptions } = require('./cli-utils');
 async function getConstructorArgs(contractName, config, wallet, options) {
     const args = options.args ? JSON.parse(options.args) : {};
```
milapsheth, line 32: maybe add a try/catch to print a readable error. easy to get args encoding incorrect. also print the JSON prettified args so user can double check

evm/deploy-upgradable.js
```diff
@@ -52,7 +50,8 @@ async function getImplementationArgs(contractName, config, options) {
                 throw new Error(`${config.name} | Missing AxelarDepositService.refundIssuer in the chain info.`);
             }
 
-            const gateway = config.AxelarGateway?.address;
+            const gateway = contractConfig.gatewayAddress ?? config.AxelarGateway?.address;
+            config.AxelarGateway.address = gateway;
```
milapsheth, line 54: doesn't need to be made configurable i think. at least we shouldn't overwrite the actual AxelarGateway config (as that is reserved to be modified by the gateway deployment script)

milapsheth (approved):

PR #212, opened 2024-04-15
fix: contract verification script failing for fantom and scroll

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-3546
https://axelarnetwork.atlassian.net/browse/AXE-3156

### DIFF
--- a/axelar-chains-config/info/mainnet.json
+++ b/axelar-chains-config/info/mainnet.json
@@ -2283,6 +2283,17 @@
       "approxFinalityWaitTime": 25
     }
   },
+  "stellar": {
+    "id": "stellar",
+    "rpc": "https://rpc.ankr.com/stellar_soroban",
+    "horizonRpc": "https://horizon.stellar.org",
+    "networkType": "mainnet",
+    "explorer": {
+      "name": "Soroban Nownodes",
+      "url": "https://soroban.nownodes.io"
+    },
+    "tokenSymbol": "XLM"
+  },
   "axelar": {
     "id": "Axelarnet",
     "axelarId": "Axelarnet",
--- a/axelar-chains-config/info/testnet.json
+++ b/axelar-chains-config/info/testnet.json
@@ -904,123 +904,6 @@
       "approxFinalityWaitTime": 60,
       "timeout": 180000
     },
-    "linea": {
-      "name": "Linea",
-      "id": "linea",
-      "axelarId": "linea",
-      "chainId": 59140,
-      "rpc": "https://linea-goerli.infura.io/v3/a4812158fbab4a2aaa849e6f4a6dc605",
-      "tokenSymbol": "ETH",
-      "contracts": {
-        "AxelarGateway": {
-          "address": "0xe432150cce91c13a887f7D836923d5597adD8E31",
-          "implementation": "0x99B5FA03a5ea4315725c43346e55a6A6fbd94098",
-          "authModule": "0x50D84DE83c089184E8EbA074ab38887B0CA30A0d",
-          "tokenDeployer": "0xb28478319B64f8D47e19A120209A211D902F8b8f",
-          "deployer": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
-          "implementationCodehash": "0x327eb5816918332c7b6787ef777587c8f9b39af4ec7b8710cdd037d976b65d5f",
-          "deploymentMethod": "create3",
-          "salt": "AxelarGateway v6.2"
-        },
-        "AxelarGasService": {
-          "salt": "AxelarGasService",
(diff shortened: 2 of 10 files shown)


axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -27,7 +27,12 @@ const verifyContract = (env, chain, contract, args, options = {}) => {
     console.log(`Verifying contract ${contract} with args '${stringArgs.join(',')}'`);
     console.log(cmd);
 
-    execSync(cmd, { stdio: 'inherit' });
+    try {
+        execSync(cmd, { stdio: 'inherit' });
```
milapsheth, line 31: we already catch the error per chain in the verify script. what's the reason for this? If there's a failure for verification for a given chain, we'll have to rerun that anyways?

milapsheth: > If for a new deployment there are 5 contracts that will be verified and if verifying the 1st one fails then all other 4 contracts too will not be verified.
This is not as big of a concern since you have to rerun it for that chain again anyways if there's any failures. I'm generally not a fan of hiding errors implicitly, and allow the caller to determine.

> From a list of 5 contracts if lets say 3rd one is upgraded and needs to be verified again. Then running the script should verify 3rd one and other 4 should say as already verified. But for fantom tetsnet the other contracts which are already verified throws error.
I see. Can we make the verify method more robust then? `verifyContract` should only fail on actual errors, not if the contract is already verified, so it needs to better handle the scenario for Fantom (e.g. can we check the error string, or can we check verification status explicitly first if chain is fantom), instead of ignoring all failures.

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -27,9 +27,26 @@ const verifyContract = (env, chain, contract, args, options = {}) => {
...
     console.log(cmd);
 
-    execSync(cmd, { stdio: 'inherit' });
+    await new Promise((resolve, reject) => {
+        exec(cmd, { stdio: 'inherit' }, (error, stdout, stderr) => {
+            console.log(stdout);
 
-    console.log('Verified!');
+            if (error) {
+                if (stderr && stderr.includes('Reason: Already Verified')) {
```
milapsheth, line 35: the reason i usually get is this. e.g. when verifying gas service v6.3.1 upgrade on testnet
`Reason: Fail - Unable to verify. Compiled contract deployment bytecode does NOT match the transaction deployment bytecode.`

can you test it on that?

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -12,7 +12,7 @@ const { writeFileSync } = require('fs');
  * @param {any[]} args
  * @returns {void}
  */
-const verifyContract = (env, chain, contract, args, options = {}) => {
+const verifyContract = async (env, chain, contract, args, options = {}) => {
```
milapsheth, line 15: `async` switch requires adding an `await` everywhere this is used. Although, I'd prefer to handle this without having to make this method async.

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -27,9 +27,26 @@ const verifyContract = (env, chain, contract, args, options = {}) => {
     console.log(`Verifying contract ${contract} with args '${stringArgs.join(',')}'`);
     console.log(cmd);
 
-    execSync(cmd, { stdio: 'inherit' });
```
milapsheth, line 30: pipe stderr to another place, and use a try/catch to read this stderr output and avoid re-raising the error in that case?

axelar-chains-config/src/utils/verifyContract.js
milapsheth: ok, that seems to be the bigger issue then on fantom. Have you tried investigating it? Do you just verify it with the flattened contract in that case? Maybe reach out to etherscan team about this issue?

axelar-chains-config/src/utils/verifyContract.js
```diff
@@ -27,9 +27,16 @@ const verifyContract = (env, chain, contract, args, options = {}) => {
...
-
-    console.log('Verified!');
+    try {
+        execSync(cmd, { stdio: ['inherit', 'pipe', 'pipe'] });
+        console.log('Verified!');
+    } catch (error) {
+        if (error.message.includes('Reason: Already Verified')) {
+            console.log(`Contract ${contract} is already verified on ${chain.toLowerCase()}.`);
+        } else {
+            throw new Error(`An error occurred while trying to verify ${contract} on ${chain.toLowerCase()}: ${error}`);
```
milapsheth, line 37:
```suggestion
            throw new Error(`An error occurred while trying to verify ${contract} on ${chain.toLowerCase()}:\n${error}`);
```
to show on a newline perhaps

milapsheth (approved):

PR #216, opened 2024-04-23
chore(cosmwasm): use gas estimation when uploading contracts

jcs47:


cosmwasm/utils.js
```diff
@@ -28,12 +28,7 @@ const uploadContract = async (client, wallet, config, options) => {
         .getAccounts()
         .then(([account]) => {
             const wasm = readFileSync(`${artifactPath}/${pascalToSnake(contractName)}${aarch64 ? '-aarch64' : ''}.wasm`);
-
-            const {
-                axelar: { gasPrice, gasLimit },
-            } = config;
-            const uploadFee = calculateFee(gasLimit, GasPrice.fromString(gasPrice));
-            return client.upload(account.address, wasm, uploadFee).then(({ checksum, codeId }) => ({ checksum, codeId, account }));
+            return client.upload(account.address, wasm, 'auto').then(({ checksum, codeId }) => ({ checksum, codeId, account }));
```
milapsheth, line 31: Maybe update the logic to use `auto` if `gasLimit` override is undefined or set to `auto` from the config, otherwise calculate the fee? Need to remove it from the config as well. Can do this in another PR as well to not block the e2e fix

milapsheth (approved):

cosmwasm/utils.js
jcs47: done

PR #217, opened 2024-04-24
feat(cosmwasm): save salt in config file

jcs47:


cosmwasm/utils.js
```diff
@@ -35,20 +35,21 @@ const uploadContract = async (client, wallet, config, options) => {
             return client.upload(account.address, wasm, uploadFee).then(({ checksum, codeId }) => ({ checksum, codeId, account }));
         })
         .then(({ account, checksum, codeId }) => {
+            const usedSalt = salt || contractName.concat(chainNames);
```
milapsheth, line 38:
```suggestion
            let salt = options.salt || contractName.concat(chainNames);
```
to reuse `salt` name

cosmwasm/utils.js
```diff
@@ -35,20 +35,21 @@ const uploadContract = async (client, wallet, config, options) => {
...
             const address = instantiate2
                 ? instantiate2Address(
                       fromHex(checksum),
                       account.address,
-                      fromHex(getSaltFromKey(salt || contractName.concat(chainNames))),
+                      fromHex(getSaltFromKey(usedSalt)),
```
milapsheth, line 43:
```suggestion
                      fromHex(getSaltFromKey(salt)),
```

cosmwasm/utils.js
```diff
@@ -35,20 +35,21 @@ const uploadContract = async (client, wallet, config, options) => {
...
                       'axelar',
                   )
                 : null;
 
-            return { codeId, address };
+            return { codeId, address, usedSalt };
```
milapsheth, line 48:
```suggestion
            return { codeId, address, salt };
```

cosmwasm/utils.js
milapsheth, line 43: Does the salt need to bytes32, or can it be a string? If the latter, then `getSaltFromKey` isn't needed. We could use the readable string as is

PR #218, opened 2024-04-25
feat(amplifier): add amplifier gateway deployment script

milapsheth:
[AXE-3486]

[AXE-3486]: https://axelarnetwork.atlassian.net/browse/AXE-3486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evm/deploy-amplifier-gateway.js
```diff
@@ -0,0 +1,427 @@
+'use strict';
```
jcs47, line 1: for the sake of orchestrations, it comes in handy to have the major and minor parts of the version number as part ofthe file name

milapsheth: We don't really plan to change this gateway. I'm also not a big fan of having different scripts per version due to a lot of duplication. The current deployment repo uses a single fixed dependency version as well. Maybe a better option is to pass in the version being deployed as a flag. But currently there is no plan to support multiple versions, so I'd prefer to add this if needed in the future

jcs47 (approved):

PR #228, opened 2024-05-09
chore: add case for appproving execute tx using batch data

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-3869

### DIFF
--- a/evm/gateway.js
+++ b/evm/gateway.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const axios = require('axios');
 const chalk = require('chalk');
 const { ethers } = require('hardhat');
 const {
@@ -21,6 +20,7 @@ const {
     mainProcessor,
     printError,
     getGasOptions,
+    httpGet,
 } = require('./utils');
 const { addBaseOptions } = require('./cli-utils');
 const { getWallet } = require('./sign-utils');
@@ -58,14 +58,10 @@ const getSignedWeightedExecuteInput = async (data, operators, weights, threshold
 
 const fetchBatchData = async (apiUrl, batchId) => {
     try {
-        const response = await axios.post(apiUrl, { batchId });
-        const data = response.data?.data;
+        const response = await httpGet(`${apiUrl}/${batchId}`);
+        const data = response?.execute_data;
 
-        if (!Array.isArray(data) || !data[0] || !data[0].execute_data) {
-            throw new Error('Invalid data format or execute_data is empty.');
-        }
-
-        return '0x' + data[0].execute_data;
+        return '0x' + data;
     } catch (error) {
         throw new Error(`Failed to fetch batch data: ${error.message}`);
     }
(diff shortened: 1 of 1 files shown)


re1ro (approved):

evm/gateway.js
```diff
@@ -183,6 +195,38 @@ async function processCommand(config, chain, options) {
             break;
         }
 
+        case 'approveWithBatch': {
+            const { batchID, api } = options;
```
milapsheth, line 199: there's no flag for the api. Reading it from the config should be fine, don't have to add an override for options where we don't have a need for it yet

evm/gateway.js
```diff
@@ -183,6 +195,38 @@ async function processCommand(config, chain, options) {
...
+
+            if (!batchID) {
+                throw new Error('Batch ID is required for the approve action');
+            }
+
+            const batchId = batchID.startsWith('0x') ? batchID.substring(2) : batchID;
+            let apiUrl = api || `${config.axelar.lcd}/axelar/evm/v1beta1/batched_commands/${chain.name.toLowerCase()}`;
```
milapsheth, line 206: `chain.axelarId`. use axelar registered chain names when interacting with the protocol

evm/gateway.js
```diff
@@ -55,6 +56,17 @@ const getSignedWeightedExecuteInput = async (data, operators, weights, threshold
     );
 };
 
+const fetchBatchData = async (apiUrl, batchId) => {
+    try {
+        const response = await httpGet(`${apiUrl}/${batchId}`);
```
milapsheth, line 61: the helper isn't very useful rn since it's still making the caller construct the apiUrl. Instead, the inputs should be chain and batchId, and it should handle the call

evm/gateway.js
```diff
@@ -183,6 +195,38 @@ async function processCommand(config, chain, options) {
...
+        case 'approveWithBatch': {
+            const { batchID, api } = options;
+
+            if (!batchID) {
+                throw new Error('Batch ID is required for the approve action');
+            }
+
+            const batchId = batchID.startsWith('0x') ? batchID.substring(2) : batchID;
+            let apiUrl = api || `${config.axelar.lcd}/axelar/evm/v1beta1/batched_commands/${chain.name.toLowerCase()}`;
+            apiUrl = apiUrl.endsWith('/') ? apiUrl.slice(0, -1) : apiUrl;
```
milapsheth, line 207: this logic is an example of duplication due to a lack of a strong guarantee, and every usage of the api will cause this. Maybe add a test for the info files here to check it's not added: https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/axelar-chains-config/tests/info.test.js#L20

we could probably use trimPrefix, trimSuffix helpers as well in our utils

evm/gateway.js
```diff
@@ -55,6 +56,17 @@ const getSignedWeightedExecuteInput = async (data, operators, weights, threshold
     );
 };
 
+const fetchBatchData = async (apiUrl, batchId) => {
+    try {
+        const response = await httpGet(`${apiUrl}/${batchId}`);
+        const data = response?.execute_data;
```
milapsheth, line 62: also check the status of the signing session in the output. If signing hasn't completed, then the execute data value won't work

PR #242, opened 2024-05-29
feat(GasService): using internal TX relay API

re1ro:


evm/gas-service.js
```diff
@@ -71,6 +71,16 @@ async function getFeeData(api, sourceChain, destinationChain) {
     return feesCache[key];
 }
 
+async function relayTransaction(chain, address, calldata, nativeValue, gasLimit) {
+    return httpPost(process.env.RELAYER_API_ENDPOINT, {
```
milapsheth, line 75: avoid reading env var directly from process. This should be read via the CLI args

evm/gas-service.js
```diff
@@ -71,6 +71,16 @@ async function getFeeData(api, sourceChain, destinationChain) {
     return feesCache[key];
 }
 
+async function relayTransaction(chain, address, calldata, nativeValue, gasLimit) {
```
milapsheth, line 74: move this to utils since it's generic

evm/gas-service.js
```diff
@@ -304,24 +314,42 @@ async function processCommand(config, chain, options) {
...
-                    chain.timeout || 60000,
-                    new Error(`Timeout updating gas info for ${chain.name}`),
-                );
-
-                const eventEmitted = wasEventEmitted(receipt, gasService, 'GasInfoUpdated');
-
-                if (!eventEmitted) {
-                    printWarn('Event not emitted in receipt.');
+                if (options.useRelay) {
+                    const gasLimit = Number(await gasService.estimateGas.updateGasInfo(chainsToUpdate, gasInfoUpdates));
```
milapsheth, line 318: does gasLimit need to be set? will the API auto-estimate if we don't set it? That would be ideal, otherwise use the chain's gasLimit from config or a high value like 5M as a default. Otherwise we're relying on the RPC in our configs which might not be robust for this

evm/gas-service.js
```diff
@@ -304,24 +314,42 @@ async function processCommand(config, chain, options) {
...
+
+                    const tx = await timeout(
```
milapsheth, line 320: This API is asynchronous I believe, so we don't need a timeout. you can double check

evm/gas-service.js
```diff
@@ -304,24 +314,42 @@ async function processCommand(config, chain, options) {
...
+                            0,
+                            gasLimit,
+                        ),
+                        chain.timeout || 60000,
+                        new Error(`Timeout updating gas info for ${chain.name}`),
+                    );
+
+                    printInfo('Relayed TX', tx.hash);
+                } else {
+                    const tx = await timeout(
```
milapsheth, line 334: this function is getting quite complicated, move this into a helper function?

evm/operators.js
```diff
@@ -268,18 +268,36 @@ async function processCommand(config, chain, options) {
...
+                    const gasLimit = Number(await operatorsContract.estimateGas.executeContract(target, updateGasInfoCalldata, 0));
+
+                    const tx = await timeout(
+                        relayTransaction(
+                            chain.axelarId,
+                            operatorsContract.address,
+                            operatorsContract.interface.encodeFunctionData('executeContract', [target, updateGasInfoCalldata, 0]),
+                            0,
+                            gasLimit,
+                        ),
```
milapsheth, line 281: same comments here

evm/gas-service.js
```diff
@@ -368,6 +396,7 @@ if (require.main === module) {
 
     // options for updateGasInfo
     program.addOption(new Option('--chains <chains...>', 'Chain names'));
+    program.addOption(new Option('--useRelay', 'Relaying the transaction through the internal API'));
```
milapsheth, line 399:
```suggestion
    program.addOption(new Option('--relayerAPI <relayerAPI>', 'Relay the tx through an external relayer API'));
```

evm/gas-service.js
ilyalukyanov: It should autoestimate if unset, yeah

evm/gas-service.js
ilyalukyanov: yep, it should return quickly and do relaying in the background. It also won't return a receipt so `tx.hash` will be `undefined`. The only field it returns is `relay_id`

evm/gas-service.js
```diff
@@ -304,24 +314,42 @@ async function processCommand(config, chain, options) {
             }
```
milapsheth, line 315: we can run the script in `--offline` mode in the cron. This will skip connecting to the config RPC. A dummy private key can be set in the cron

evm/utils.js
```diff
@@ -1196,6 +1196,39 @@ function timeout(prom, time, exception) {
...
+            call_data: contract.interface.encodeFunctionData(method, params),
+            value: nativeValue,
+        });
+
+        printInfo('Relay ID', result.relay_id);
+    } else {
+        await timeout(
+            (async () => {
+                const tx = await contract[method](...params, gasOptions);
+                printInfo('TX', tx.hash);
```
milapsheth, line 1213:
```suggestion
                printInfo('Tx hash', tx.hash);
```

evm/utils.js
```diff
@@ -1196,6 +1196,39 @@ function timeout(prom, time, exception) {
...
+    if (options.relayerAPI) {
+        const result = await httpPost(options.relayerAPI, {
+            chain: chain.axelarId,
+            to: contract.address,
+            call_data: contract.interface.encodeFunctionData(method, params),
+            value: nativeValue,
+        });
+
+        printInfo('Relay ID', result.relay_id);
+    } else {
```
milapsheth, line 1209: return in the if branch and avoid the extra else nesting

evm/gas-service.js
```diff
@@ -396,7 +359,7 @@ if (require.main === module) {
 
     // options for updateGasInfo
     program.addOption(new Option('--chains <chains...>', 'Chain names'));
-    program.addOption(new Option('--useRelay', 'Relaying the transaction through the internal API'));
+    program.addOption(new Option('--relayerAPI <relayerAPI>', 'Relay the tx through an external relayer API').env('RELAYER_API_ENDPOINT'));
```
milapsheth, line 362:
```suggestion
    program.addOption(new Option('--relayerAPI <relayerAPI>', 'Relay the tx through an external relayer API').env('RELAYER_API'));
```
match the flag name

milapsheth (approved):

PR #246, opened 2024-05-31
fix(GasService): request and response of relay API

ilyalukyanov:


evm/utils.js
```diff
@@ -1201,11 +1201,16 @@ async function relayTransaction(options, chain, contract, method, params, native
...
-            value: nativeValue,
+            calldata: contract.interface.encodeFunctionData(method, params),
+            value: nativeValue.toString(),
         });
 
-        printInfo('Relay ID', result.relay_id);
+        if (!result.error) {
+            printInfo('Relay ID', result.relayId);
+        } else {
+            printInfo('Relay Error', result.error);
```
milapsheth, line 1211:
```suggestion
            throw new Error(`Relay Error: ${result.error}`);
```
The caller/script catches errors and exits with non zero return code at the end if there was a failure for any chain

ilyalukyanov: SG, but I'd like to also do this https://github.com/axelarnetwork/axelar-contract-deployments/pull/246/commits/da8be7aeb8bb24e7b819037e132e9224d6101e41

For how I wrapped the script, it's convenient not to exit with non-zero code. I also don't want an isolated failure on a single chain to prevent updates on the rest of them.

re1ro (approved):

evm/utils.js
milapsheth: yeah, the script already [catches the error](https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/evm/gas-service.js#L317) and continues running it for other chains. But yeah, it'll exit with an error code at the end if there was any failure so `--ignoreError` support works.

PR #249, opened 2024-05-31
ci(GasService): remove gas info updates

re1ro:
* [x] GitHub CI: removing cron updates for the gas info


.github/workflows/update-gas-info.yaml
milapsheth, line 1: only remove the cron. we can still use the action for manual updates if needed

milapsheth (approved):

PR #250, opened 2024-06-03
fix(evm): domain separator calculation

jcs47:


milapsheth (approved):

evm/deploy-amplifier-gateway.js
```diff
@@ -77,13 +77,13 @@ async function getDomainSeparator(config, chain, options) {
...
     const {
-        axelar: { contracts, chainId, axelarId },
+        axelar: { contracts, chainId },
     } = config;
     const {
         Router: { address: routerAddress },
     } = contracts;
     const domainSeparator = hexlify((await getContractConfig(config, chain.axelarId)).domain_separator);
-    const expectedDomainSeparator = calculateDomainSeparator(axelarId, routerAddress, chainId);
+    const expectedDomainSeparator = calculateDomainSeparator(chain.axelarId, routerAddress, chainId);
```
milapsheth, line 86: add checks in this method to ensure none of the args are undefined

evm/deploy-amplifier-gateway.js
```diff
@@ -77,13 +77,13 @@ async function getDomainSeparator(config, chain, options) {
     }
```
milapsheth, line 77: also needs to be fixed here
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/cosmwasm/deploy-contract.js#L204

PR #252, opened 2024-06-03
fix(amplifier): salt derivation and chain.axelarId

eguajardo:
For the salt derivation to be correct, this other PR needs to be merged: https://github.com/axelarnetwork/axelar-contract-deployments/pull/251

https://axelarnetwork.atlassian.net/browse/AXE-4153

### DIFF
--- a/evm/deploy-amplifier-gateway.js
+++ b/evm/deploy-amplifier-gateway.js
@@ -82,7 +82,7 @@ async function getDomainSeparator(config, chain, options) {
     const {
         Router: { address: routerAddress },
     } = contracts;
-    const domainSeparator = hexlify((await getContractConfig(config, chain.name)).domain_separator);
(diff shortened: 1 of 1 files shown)


evm/deploy-amplifier-gateway.js
```diff
@@ -184,7 +184,7 @@ async function deploy(config, chain, options) {
     contractConfig.deployer = wallet.address;
     const domainSeparator = await getDomainSeparator(config, chain, options);
     const minimumRotationDelay = Number(options.minimumRotationDelay);
-    const salt = options.salt || '';
+    const salt = options.salt || `AxelarAmplifierGateway ${config.axelar.chainId}`;
```
milapsheth, line 187:
```suggestion
    const salt = options.salt || 'AxelarAmplifierGateway';
```
Let's use this as the default, so testnet/mainnet derive the same addresses. For devnets, our orchestration can pass in an appropriate salt perhaps

milapsheth (approved):

PR #255, opened 2024-06-04
feat: add script to merge and split coins

npty:
# Description

[AXE-4095](https://axelarnetwork.atlassian.net/browse/AXE-4095)

Added a script to manage coins including:
- Merge coins for every coin types or give type into 1 object
- Split all coins or given specific coin by provided amount
- Shows total balance for each coin

### 1. Shows total balance for each coin

```bash
node sui/tokens.js list 
```

Output:
```bash
Action: List Coins

Wallet Address: 0xebafd8ef2ee5084a73f36eda4a6d0d36fe6b6a4e4a93e5ecda9bf337a7a1e7fa

Coin Type: 0x2::sui::SUI

Total Balance: 1999977577440 (1999.97757744)

Total Objects: 2

Coin Type: 0x3569c9e7452a68d7bfeffabc7f6343cbe48ff3dda607c9a1249d156d661b16e5::my_coin::MY_COIN

Total Balance: 1000000000000 (1000000.0)

Total Objects: 1

Coin Type: 0x5d6e299d0b1315fa1ee74605476e8635c8b7ad1de580c7e2a23a35640096c1bb::my_coin::MY_COIN

Total Balance: 1000000000000 (1000000.0)

Total Objects: 1
```

### 2. Merge coins

Merge all coins
```bash
node sui/tokens.js merge
```

Merge only specific coin type
```bash
 node sui/tokens.js merge --coin-type 0x3569c9e7452a68d7bfeffabc7f6343cbe48ff3dda607c9a1249d156d661b16e5::my_coin::MY_COIN
```

Output:
```bash
Action: Merge Coins

Merge Coins: 0x3569c9e7452a68d7bfeffabc7f6343cbe48ff3dda607c9a1249d156d661b16e5::my_coin::MY_COIN
```

### 3. Split coins

Split all coins to given amount
```bash
node sui/tokens.js split 100000
```

Split only specific coin type
```bash
 node sui/tokens.js split 100000 --coin-type 0xbb713fb5bb92d32d45754981d01942e0f1b191ea7b2ad12d745d0c9f4dd7ea10::my_coin::MY_COIN
```

Output:
> will update later

[AXE-4095]: https://axelarnetwork.atlassian.net/browse/AXE-4095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+    const program = new Command();
+    program.name('manage-coins').description('Merge or split coins for an account');
+
+    addBaseOptions(program);
+
+    program
+        .option('--merge', 'Merge all coins')
+        .option('--split <amount>', 'Split coins')
+        .option('--coin-type <coinType>', 'Coin type to merge/split')
+        .option('--transfer <recipientAddress>', 'Used with --split to transfer the split coins to the recipient address')
```
milapsheth, line 213: In other scripts, we use the `--action` pattern for the main action, instead of separate cli flags.

Although, I'd like to switch CLI groups going forward, which is the right way to handle CLI cmds (e.g. how `axelard` nests commands). Can you try that out here?
https://github.com/tj/commander.js?tab=readme-ov-file#commands

We can still use the `mainProcessor` pattern, although it can also forward the keypair and client and processCommand can be action specific maybe? broadcast can be a shared helper method.

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+
+async function mainProcessor(options, processor) {
+    const config = loadSuiConfig(options.env);
+    await processor(config, config.sui, options);
+    saveConfig(config, options.env);
+}
+
+if (require.main === module) {
+    const program = new Command();
+    program.name('manage-coins').description('Merge or split coins for an account');
```
milapsheth, line 205:
```suggestion
    const program = new Command();

    program.name('tokens').description('Token management tool (e.g. show balance, merge, split etc.)');
```
Let's standardize `tokens.js` for all chain folders?

sui/manage-coins.js
milapsheth: cc @blockchainguyy (similar to the Click refactor in python)

we should refactor other scripts to follow this pattern as well

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
+const { TransactionBlock } = require('@mysten/sui.js/transactions');
+const { saveConfig } = require('../evm/utils');
+const { Command } = require('commander');
+const { addBaseOptions } = require('./cli-utils');
+const { getWallet } = require('./sign-utils');
+const { printInfo } = require('../evm/utils');
+const chalk = require('chalk');
+const { loadSuiConfig } = require('./utils');
+
+const SUI_COIN_ID = '0x2::sui::SUI';
```
milapsheth, line 10: export this in utils maybe (it might be defined in Sui SDK as well)

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+
+class CoinManager {
```
milapsheth, line 12: using a class here feels a bit overkill since we never instantiate it, or export it. But fine to keep it and see how it goes

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+        } while (cursor);
+
+        return coinTypeToCoins;
+    }
+
+    static async splitCoins(tx, coinTypeToCoins, options) {
+        CoinManager.checkSplitAmount(options);
+        const splitAmount = BigInt(options.split);
+
+        console.log('\n==== Splitting Coins ====');
```
milapsheth, line 49: use `printInfo` over console.log from utils

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+            console.log(`\nDone`);
+        }
+    }
+
+    static printAllCoins(coinTypeToCoins) {
+        console.log('==== Coins Info ====');
+
+        for (const coinType in coinTypeToCoins) {
+            const coins = coinTypeToCoins[coinType];
+            console.log(`Coin Type: ${chalk.green(coinType)}`);
```
milapsheth, line 170: `printInfo` handles colouring

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+        const firstCoin = coinObjectIds.shift();
+        const remainingCoins = coinObjectIds.map((id) => tx.object(id));
+
+        tx.mergeCoins(firstCoin, remainingCoins);
+        console.log(`Merge ${coins.data.length} coins of type '${chalk.green(coins.data[0].coinType)}'`);
+    }
+
+    static async processCommand(config, chain, options) {
+        const [keypair, client] = getWallet(chain, options);
+        printInfo('Wallet Address', keypair.toSuiAddress());
```
milapsheth, line 135: see printWalletInfo, it prints the balance as well

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+
+const SUI_COIN_ID = '0x2::sui::SUI';
+
+class CoinManager {
+
+    static async getAllCoins(client, account) {
+        let cursor;
+        const coinTypeToCoins = {};
+
+        do {
```
milapsheth, line 18: Create a generic util for paginated queries? It can take a lambda that accepts a cursor and makes the query (`async (cursor) => { return ...query... }`), and the util can iteratively call it and returns all items.

This is assuming that sui SDK uses `hasNextPage`, `nextCursor` and `cursor` fields for other paginated queries as well.

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+        tx.mergeCoins(firstCoin, remainingCoins);
+        console.log(`Merge ${coins.data.length} coins of type '${chalk.green(coins.data[0].coinType)}'`);
+    }
+
+    static async processCommand(config, chain, options) {
+        const [keypair, client] = getWallet(chain, options);
+        printInfo('Wallet Address', keypair.toSuiAddress());
+
+        const coinTypeToCoins = await CoinManager.getAllCoins(client, keypair.toSuiAddress());
+        CoinManager.printAllCoins(coinTypeToCoins);
```
milapsheth, line 138: Looking into the Sui SDK, we could use `getAllBalances` method for printing balances or `getBalances` if we want to print it for a given coin instead of `getAllCoins`

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+if (require.main === module) {
+    const program = new Command();
+    program.name('manage-coins').description('Merge or split coins for an account');
+
+    addBaseOptions(program);
+
+    program
+        .option('--merge', 'Merge all coins')
+        .option('--split <amount>', 'Split coins')
+        .option('--coin-type <coinType>', 'Coin type to merge/split')
```
milapsheth, line 212: add a balance action as well that can print total balance, or by provided coin

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+
+    static isGasToken(coin) {
+        return coin.coinType === '0x2::sui::SUI';
+    }
+
+    static checkSplitAmount(options) {
+        try {
+            parseInt(options.split);
+        } catch (e) {
+            console.error('\nError: Please specify a valid split amount');
```
milapsheth, line 191: we have printError in utils as well

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+
+    static doSplitCoins(tx, coins, splitAmount) {
+        const firstObjectId = this.isGasToken(coins.data[0]) ? tx.gas : coins.data[0].coinObjectId;
+        const response = tx.splitCoins(firstObjectId, [splitAmount]);
+        console.log(`Split coins of type '${chalk.green(coins.data[0].coinType)}' with amount ${splitAmount}`);
+        return response;
+    }
+
+    static async mergeCoin(tx, coinTypeToCoins, options) {
+        const coinType = options.coinType;
```
milapsheth, line 96:
```suggestion
        const coinTypes = options.coinType ? [options.coinType] : Object.keys(coinTypeToCoins);
```

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+        console.log('\n==== Merging Coins ====');
+
+        // Throw an error if the coin type is specified but no coins are found
+        CoinManager.checkCoinType(coinType, coinTypeToCoins);
+
+        if (coinType) {
+            const coins = coinTypeToCoins[coinType];
+            await CoinManager.doMergeCoin(tx, coins);
+        } else {
+            for (const coinType in coinTypeToCoins) {
```
milapsheth, line 106:
```suggestion
            for (const coinType in coinTypes) {
```

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+                const coins = coinTypeToCoins[coinType];
```
milapsheth, line 107: throw error if coins is undefined. Don't think a check helper is needed for it

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+            if (coinsAtCursor.hasNextPage) {
+                cursor = coinsAtCursor.nextCursor;
+            }
+        } while (cursor);
+
+        return coinTypeToCoins;
+    }
+
+    static async splitCoins(tx, coinTypeToCoins, options) {
+        CoinManager.checkSplitAmount(options);
```
milapsheth, line 46: use this helper instead
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/evm/its.js#L251

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+        const splitAmount = BigInt(options.split);
+
+        console.log('\n==== Splitting Coins ====');
+
+        // Set coin type to given coin type or the first coin type if there's only one if it's a SUI token.
+        const hasOnlyGasToken = Object.keys(coinTypeToCoins).length === 1 && coinTypeToCoins[SUI_COIN_ID];
```
milapsheth, line 52: split should only split for one coin type i think. If coin type isn't provided, Sui is assumed

sui/manage-coins.js
```diff
@@ -0,0 +1,218 @@
...
+        tx.setGasBudget(1e8);
+    }
+
+    static doTransfer(tx, coin, recipient) {
+        tx.transferObjects([coin], recipient);
+    }
+
+    static doSplitCoins(tx, coins, splitAmount) {
+        const firstObjectId = this.isGasToken(coins.data[0]) ? tx.gas : coins.data[0].coinObjectId;
+        const response = tx.splitCoins(firstObjectId, [splitAmount]);
```
milapsheth, line 90: splitAmount should be decimal aware. Script user will provide 1 as the amount to mean 1 SUI. does the Sui SDK handle this? If not, we should do it ourselves (by retrieving coin metadata)

milapsheth (changes requested):

sui/manage-coins.js
npty: I'd like to group these three lines for each coin close together like this.

```
==== Coins Info ====
Coin Type: 0x2::sui::SUI
Total Balance: 1000000000000
Total Objects: 5
````

But the `printInfo` function adds an extra line like this.

```
==== Coins Info ====
Coin Type:: 0x2::sui::SUI

Total Balance:: 1000000000000

Total Objects:: 5
```

I guess you prefer the latter? I'll update to printInfo.

File: sui/manage-coins.js (hunk 7 above)
Comment (npty) [reply sui/manage-coins.js:135]: I don't use printWalletInfo because this program will print balance for all coins anyway which i think it's a bit redundant.

File: sui/manage-coins.js (hunk 14 above)
Comment (npty) [reply sui/manage-coins.js:107]: i think `coinType` cannot be undefined here, as it is being iterated over the defined keys in `coinTypeToCoins`.

File: sui/manage-coins.js (hunk 17 above)
Comment (npty) [reply sui/manage-coins.js]: can't find it. so i'll do it manually using sui sdk to get decimals from coin metadata and use ethers.js to handle amount conversion.

File: sui/manage-coins.js (hunk 4 above)
Comment (npty) [reply sui/manage-coins.js]: yeah, it looks like an overkill by using a class. but my purpose is only for grouping coins-management functions together (there're too many functions) without create a separate file for them. it's easier for me to see overall code, especially when fold or expand code in editor.

File: sui/manage-coins.js (hunk 9 above)
Comment (npty) [reply sui/manage-coins.js]: yes, it's exactly the same. however, we still need the data from `getAllCoins` api for merge and split mode, so i think we could just reuse the data from it to print coin's balance details.

File: sui/tokens.js [hunk 18]
```diff
@@ -43,48 +43,23 @@ class CoinManager {
     }
 
     static async splitCoins(tx, coinTypeToCoins, options) {
-        CoinManager.checkSplitAmount(options);
-        const splitAmount = BigInt(options.split);
-
-        console.log('\n==== Splitting Coins ====');
+        printInfo('\n==== Splitting Coins ====');
```
milapsheth, line 46: printInfo already prints an extra newline at the end
```suggestion
        printInfo('==== Splitting Coins ====');
```

milapsheth, line 46: print the action name in `processCommand` instead of printing in each helper function. For consistency, use `printInfo('Action', actionType)` print format

sui/tokens.js
```diff
@@ -43,48 +43,23 @@ class CoinManager {
...
         tx.setGasBudget(1e8);
     }
 
-    static doTransfer(tx, coin, recipient) {
-        tx.transferObjects([coin], recipient);
-    }
-
     static doSplitCoins(tx, coins, splitAmount) {
         const firstObjectId = this.isGasToken(coins.data[0]) ? tx.gas : coins.data[0].coinObjectId;
         const response = tx.splitCoins(firstObjectId, [splitAmount]);
```
milapsheth, line 65: there's still some console.log's remaining

sui/tokens.js
```diff
@@ -43,48 +43,23 @@ class CoinManager {
...
 
         // The transaction will fail if the gas budget is not set for splitting coins transaction
         tx.setGasBudget(1e8);
     }
 
-    static doTransfer(tx, coin, recipient) {
-        tx.transferObjects([coin], recipient);
-    }
-
     static doSplitCoins(tx, coins, splitAmount) {
```
milapsheth, line 63: merge this with splitCoins. There seem to be too many helpers rn. splitCoins is much smaller now

sui/tokens.js
```diff
@@ -43,48 +43,23 @@ class CoinManager {
...
-        const hasOnlyGasToken = Object.keys(coinTypeToCoins).length === 1 && coinTypeToCoins[SUI_COIN_ID];
-        const coinType = options.coinType || hasOnlyGasToken ? SUI_COIN_ID : undefined;
-
-        if (coinType) {
-            // Throw an error if the coin type is specified but no coins are found
-            CoinManager.checkCoinType(coinType, coinTypeToCoins);
-            const coins = coinTypeToCoins[coinType];
-            const [coin] = CoinManager.doSplitCoins(tx, coins, splitAmount);
+        const splitAmount = BigInt(options.split);
+        const coinType = options.coinType || SUI_COIN_ID;
```
milapsheth, line 49: coinType has already been processed before. No need to process it again

sui/tokens.js
```diff
@@ -142,7 +114,27 @@ class CoinManager {
...
+                printError(`No metadata found for coin type ${coinType}`);
+                process.exit(0);
+            }
+
+            const splitAmount = parseUnits(options.split, metadata.decimals);
+
+            await CoinManager.splitCoins(tx, coinTypeToCoins, {
+                ...options,
+                split: splitAmount.toString(),
+                coinType,
```
milapsheth, line 136: pass these as positional args. `options` should be passed as-is, we should avoid modifying it like this, it becomes harder to reason about then (ideally an implementor can just refer to CLI options and know exactly what to expect)

It also avoids needing weird logic like converting the amount back to a string for e.g.

sui/tokens.js
```diff
@@ -93,20 +68,17 @@ class CoinManager {
...
-        } else {
-            for (const coinType in coinTypeToCoins) {
-                const coins = coinTypeToCoins[coinType];
-                await CoinManager.doMergeCoin(tx, coins);
+
+            if (!coins) {
+                throw new Error(`No coins found for coin type ${coinType}`);
             }
+
+            await CoinManager.doMergeCoin(tx, coins);
```
milapsheth, line 81: merge doMergeCoin and this method

npty:
@milapsheth i've updated the pr to use new subcommands pattern. please take another look 🙏🏻

sui/tokens.js
```diff
@@ -0,0 +1,219 @@
...
+        mainProcessor(options, processListCommand);
+    });
+
+    // Add sub-programs to the main program
+    program.addCommand(mergeProgram);
+    program.addCommand(splitProgram);
+    program.addCommand(listProgram);
+
+    // Add base options to all sub-programs
+    addOptionsToCommands(program, addBaseOptions);
```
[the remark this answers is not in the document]
npty: can you try again? it works for me 🤔

milapsheth (approved):

sui/tokens.js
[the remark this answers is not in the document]
npty: fixed

PR #261, opened 2024-06-05
feat: script to deploy gas service

npty:
- feat: deploy gas service
- chore: update default faucet url in readme and faucet.js

### DIFF
--- a/sui/README.md
+++ b/sui/README.md
@@ -33,6 +33,7 @@ If you want to run against a local Sui network, then create a `axelar-chains-con
         "networkType": "localnet",
         "tokenSymbol": "SUI",
         "rpc": "http://127.0.0.1:9000",
+        "faucetUrl": "http://127.0.0.1:9123",
         "contracts": {}
     }
 }
@@ -50,7 +51,7 @@ node sui/faucet.js
 
 2. Deploy the gateway package:
 
-- By querying the signer set from the Amplifier contract (this only works if Amplifier contracts have been setup):
+-   By querying the signer set from the Amplifier contract (this only works if Amplifier contracts have been setup):
 
 ```bash
 node sui/deploy-gateway.js
@@ -58,13 +59,13 @@ node sui/deploy-gateway.js
 
 Use `--help` flag to see other setup params that can be overridden.
 
-- For testing convenience, you can use the secp256k1 wallet as the signer set for the gateway.
+-   For testing convenience, you can use the secp256k1 wallet as the signer set for the gateway.
 
 ```bash
 node sui/deploy-gateway.js --signers wallet
 ```
 
-- You can also provide a JSON object with a full signer set:
+-   You can also provide a JSON object with a full signer set:
 
 ```bash
 node sui/deploy-gateway.js -e testnet --signers '{"signers": [{"pubkey": "0x020194ead85b350d90472117e6122cf1764d93bf17d6de4b51b03d19afc4d6302b", "weight": 1}], "threshold": 1, "nonce": "0x0000000000000000000000000000000000000000000000000000000000000000"}'
--- a/sui/deploy-gas-service.js
+++ b/sui/deploy-gas-service.js
@@ -1,0 +1,55 @@
+const { saveConfig, printInfo } = require('../evm/utils');
+const { Command } = require('commander');
+const { publishPackage, updateMoveToml } = require('@axelar-network/axelar-cgp-sui/scripts/publish-package');
+
+const { addBaseOptions } = require('./cli-utils');
+const { getWallet, printWalletInfo } = require('./sign-utils');
+const { loadSuiConfig } = require('./utils');
+
+async function processCommand(config, chain, options) {
+    const [keypair, client] = getWallet(chain, options);
+    const packageName = 'gas_service';
+
+    await printWalletInfo(keypair, client, chain, options);
+
+    if (!chain.contracts.axelar_gas_service) {
+        chain.contracts.axelar_gas_service = {};
+    }
+
+    const contractConfig = chain.contracts.axelar_gas_service;
+
+    const published = await publishPackage(packageName, client, keypair);
+    const packageId = published.packageId;
+
+    updateMoveToml(packageName, packageId);
+
+    const gasService = published.publishTxn.objectChanges.find(
+        (change) => change.objectType === `${packageId}::${packageName}::GasService`,
+    );
+    contractConfig.address = packageId;
+    contractConfig.objects = {
+        gasService: gasService.objectId,
+    };
+
+    printInfo('\nGas Service deployed', JSON.stringify(contractConfig, null, 2));
+}
+
+async function mainProcessor(options, processor) {
+    const config = loadSuiConfig(options.env);
+    await processor(config, config.sui, options);
+    saveConfig(config, options.env);
+}
+
+if (require.main === module) {
+    const program = new Command();
+
+    program.name('deploy-gas-service').description('Deploys/publishes the Axelar Gas Service package');
+
+    addBaseOptions(program);
+
+    program.action((options) => {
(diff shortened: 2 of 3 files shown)


sui/deploy-gas-service.js
```diff
@@ -0,0 +1,55 @@
...
+        (change) => change.objectType === `${packageId}::${packageName}::GasService`,
+    );
+
+    const contractConfig = chain.contracts.axelar_gas_service;
+    contractConfig.address = packageId;
+    contractConfig.objects = {
+        gasService: gasService.objectId,
+    };
+
+    printInfo('\nGas Service deployed', JSON.stringify(contractConfig, null, 2));
```
milapsheth, line 34:
```suggestion
    printInfo('Gas Service deployed', JSON.stringify(contractConfig, null, 2));
```
not needed

sui/deploy-gas-service.js
```diff
@@ -0,0 +1,55 @@
...
+async function mainProcessor(options, processor) {
+    const config = loadSuiConfig(options.env);
+    await processor(config, config.sui, options);
+    saveConfig(config, options.env);
+}
+
+if (require.main === module) {
+    const program = new Command();
+
+    program.name('deploy-gas-service').description('Deploys/publishes the Axelar Gas Service package');
```
milapsheth, line 46: make this into a generic `deploy-contract` (see EVM). Most contracts are straightforward to deploy

milapsheth (approved):

sui/deploy-gas-service.js
npty: should I also include the gateway deployment into one of the options for this generic as well?

sui/deploy-contract.js
```diff
@@ -0,0 +1,78 @@
...
+    const published = await publishPackage(contract.packageName, client, keypair);
+    const packageId = published.packageId;
+
+    updateMoveToml(contract.packageName, packageId);
+
+    const contractObject = published.publishTxn.objectChanges.find(
+        (change) => change.objectType === `${packageId}::${contract.packageName}::${contract.contractName}`,
+    );
+    const gasCollectorCapObject = published.publishTxn.objectChanges.find(
+        (change) => change.objectType === `${packageId}::${contract.packageName}::GasCollectorCap`,
```
milapsheth, line 39: create a util to find the object from the published tx and other inputs

sui/README.md
```diff
@@ -33,6 +33,7 @@ If you want to run against a local Sui network, then create a `axelar-chains-con
         "networkType": "localnet",
         "tokenSymbol": "SUI",
         "rpc": "http://127.0.0.1:9000",
+        "faucetUrl": "http://127.0.0.1:9123",
         "contracts": {}
     }
```
milapsheth, line 38: add readme example for gas service deployment

sui/deploy-contract.js
```diff
@@ -0,0 +1,78 @@
...
+    const contractConfig = chain.contracts[contract.chainConfigKey];
+    contractConfig.address = packageId;
+    contractConfig.objects = {
+        [contractName]: contractObject.objectId,
+    };
+
+    switch (contractName) {
+        case contractMap.gas_service.packageName:
+            contractConfig.objects.gas_collector_cap = gasCollectorCapObject.objectId;
+            break;
```
milapsheth, line 51: add default case where an error is thrown for completeness

sui/deploy-contract.js
```diff
@@ -0,0 +1,78 @@
...
+
+    const contractConfig = chain.contracts[contract.chainConfigKey];
+    contractConfig.address = packageId;
+    contractConfig.objects = {
+        [contractName]: contractObject.objectId,
+    };
+
+    switch (contractName) {
+        case contractMap.gas_service.packageName:
+            contractConfig.objects.gas_collector_cap = gasCollectorCapObject.objectId;
```
milapsheth, line 50:
```suggestion
            contractConfig.objects.gasCollectorCap = gasCollectorCapObject.objectId;
```
let's stick to camel case? we should change the contract keys to it for consistency as well

milapsheth (approved):

sui/deploy-contract.js
```diff
@@ -0,0 +1,76 @@
...
+const { getWallet, printWalletInfo } = require('./sign-utils');
+const { loadSuiConfig, findPublishedObject } = require('./utils');
+
+// Add more contracts here to support more modules deployment
+const contractMap = {
+    gasService: {
+        packageName: 'gas_service',
+        contractName: 'GasService',
+        displayName: 'Gas Service',
+        chainConfigKey: 'axelar_gas_service',
```
milapsheth, line 15: this is redundant since we should use GasService for the key. Let's just set the `packageName` in the config itself (it can be provided as an arg override too). I really don't like having multiple configs that we have to maintain.

sui/deploy-contract.js
```diff
@@ -0,0 +1,76 @@
...
+const { addBaseOptions } = require('./cli-utils');
+const { getWallet, printWalletInfo } = require('./sign-utils');
+const { loadSuiConfig, findPublishedObject } = require('./utils');
+
+// Add more contracts here to support more modules deployment
+const contractMap = {
+    gasService: {
+        packageName: 'gas_service',
+        contractName: 'GasService',
+        displayName: 'Gas Service',
```
milapsheth, line 14: not really needed, `GasService` is fine. If we do need display name in the future, it should go in the main config

sui/deploy-contract.js
```diff
@@ -0,0 +1,76 @@
...
+    contractConfig.objects = {
+        [contractName]: contractObject.objectId,
+    };
+
+    switch (contractName) {
+        case 'gasService':
+            contractConfig.objects.gasCollectorCap = gasCollectorCapObject.objectId;
+            break;
+        default:
+            break;
```
milapsheth, line 49: error should be thrown here

milapsheth (approved):

sui/deploy-contract.js
npty: this line is impossible to reach since it'll throw earlier by the commander lib as available contract names are defined as choices.

sui/README.md
```diff
@@ -51,26 +51,32 @@ node sui/faucet.js
 
 Deploy the gateway package:
 
-- By querying the signer set from the Amplifier contract (this only works if Amplifier contracts have been setup):
+-   By querying the signer set from the Amplifier contract (this only works if Amplifier contracts have been setup):
```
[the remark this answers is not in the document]
npty: it's accidentally formatted by vscode. will revert.

sui/deploy-contract.js
```diff
@@ -8,48 +8,46 @@ const { loadSuiConfig, findPublishedObject } = require('./utils');
...
         packageName: 'gas_service',
-        contractName: 'GasService',
-        displayName: 'Gas Service',
-        chainConfigKey: 'axelar_gas_service',
     },
 };
 
 async function processCommand(contractName, config, chain, options) {
     const contract = contractMap[contractName];
+    const packageName = options.packageName || contract.packageName;
```
milapsheth, line 18: I think we should remove `contractMap` altogether. `Move.toml` for each of our packages contains
```
[package]
name = "AxelarGateway"
```
Could just recursively find the right package name from by searching (we do something similar to get the abi/bytecode for EVM builds by contract name, `getContractJSON` util).

You can add this in a separate PR though

sui/deploy-contract.js
milapsheth: i understand but we shouldn't rely on that. For e.g. what if the choices are extended or renamed but this switch case isn't updated. It would be a less readable error

milapsheth (approved):

sui/deploy-contract.js
npty: Ok, will do this separately.

PR #263, opened 2024-06-06
feat(amplifier): store setupParams when deploying an external amplifier gateway

eguajardo:
https://axelarnetwork.atlassian.net/browse/AXE-4195


evm/deploy-amplifier-gateway.js
```diff
@@ -240,6 +240,7 @@ async function deploy(config, chain, options) {
         printInfo('Proxy deployment args', `${implementation.address}, ${params}`);
 
         contractConfig.operator = operator;
+        contractConfig.setupParams = params;
```
milapsheth, line 243:
```suggestion
        const proxyDeploymentArgs = [implementation.address, owner, params];
        contractConfig.proxyDeploymentArgs = proxyDeploymentArgs;
```
this handles the full source verification requirement

milapsheth: reuse `proxyDeploymentArgs` below as well

axelar-chains-config/info/devnet-amplifier.json
```diff
@@ -25,7 +25,8 @@
           "salt": "AxelarAmplifierGateway devnet-amplifier",
           "gasOptions": {
             "gasLimit": 5000000
-          }
+          },
+          "setupParams": "0x000000000000000000000000ba76c6980428a0b10cfc5d8ccb61949677a612330000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000cb9e1000000000000000000000000000000000000000000000000000000000000000300000000000000000000000051380cbf0777990e197a3e498ffafd26143e35f8000000000000000000000000000000000000000000000000000000000000000100000000000000000000000061373485594010371dfbf3a8f8bd1736bfda7c090000000000000000000000000000000000000000000000000000000000000001000000000000000000000000c70aa87b38e6ab3c0df6e56338f96a5c00e653080000000000000000000000000000000000000000000000000000000000000001"
```
milapsheth, line 29: the reason I never added this before was that it gets quite long for testnet/mainnet deployments (25+ verifiers) and worsens readability a bit. But still seems better to add it at this point.

Regardless, is there a way to fetch the initial verifier set though? Can we also persist a verifier set id or something in the config to fetch it later? We didn't have this issue in core due to key ids

eguajardo: https://axelarnetwork.atlassian.net/browse/AXE-4199

milapsheth (approved):

PR #267, opened 2024-06-07
feat: add verify support for axelar amplifier gateway

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-4196


Foivos (approved):

evm/verify-contract.js
```diff
@@ -320,6 +330,46 @@ async function processCommand(config, chain, options) {
...
+            if (args) {
+                args = JSON.parse(args);
+                options = { ...options, ...args };
+            }
+
+            const AxelarAmplifierGateway = require('@axelar-network/axelar-gmp-sdk-solidity/artifacts/contracts/gateway/AxelarAmplifierGateway.sol/AxelarAmplifierGateway.json');
+            const amplifierGatewayFactory = await getContractFactoryFromArtifact(AxelarAmplifierGateway, wallet);
+            const amplifierGateway = amplifierGatewayFactory.attach(options.address || chain.contracts.AxelarGateway.address);
+
+            const implementation = await amplifierGateway.implementation();
```
milapsheth, line 345: see latest amplifier deployment script / devnet-amplifier config. `proxyDeploymentArgs` are checked in now, switch to using that. The queried value can't be used for proxy verification (but only for the implementation verification)

evm/utils.js
```diff
@@ -1235,6 +1236,37 @@ async function relayTransaction(options, chain, contract, method, params, native
     );
 }
 
+async function getWeightedSigners(config, chain, options) {
+    printInfo(`Retrieving verifier addresses for ${chain.name} from Axelar network`);
+
+    let signers;
+
+    if (isValidAddress(options.keyID)) {
```
milapsheth, line 1244: why is this being duplicated? import from the amplifier gateway deployment script in verify-contract. Or use this in deployment as well

evm/verify-contract.js
```diff
@@ -320,6 +330,46 @@ async function processCommand(config, chain, options) {
...
+            const previousSignersRetention = (await amplifierGateway.previousSignersRetention()).toNumber();
+            const domainSeparator = await amplifierGateway.domainSeparator();
+            const minimumRotationDelay = (await amplifierGateway.minimumRotationDelay()).toNumber();
+
+            options.keyID =
+                chain.contracts.AxelarGateway.startingKeyIDs[0] || options.keyID || `evm-${chain.axelarId.toLowerCase()}-genesis`;
```
milapsheth, line 351: keyIDs aren't used for amplifier gateway when retrieving signers from axelar, see the deployment script

evm/verify-contract.js
```diff
@@ -320,6 +330,46 @@ async function processCommand(config, chain, options) {
...
+            const signerSets = await getWeightedSigners(config, chain, options);
+            const operator = options.operator || chain.contracts.AxelarGateway.operator || wallet.address;
+            const owner = options.owner || chain.contracts.InterchainGovernance?.address || wallet.address;
+
+            const setupParams = defaultAbiCoder.encode([`address`, `${WEIGHTED_SIGNERS_TYPE}[]`], [operator, signerSets]);
```
milapsheth, line 356: not needed due to proxy args in the config

evm/verify-contract.js
```diff
@@ -320,6 +330,46 @@ async function processCommand(config, chain, options) {
...
+
+            verifyOptions.contractPath = 'contracts/gateway/AxelarAmplifierGateway.sol:AxelarAmplifierGateway';
```
milapsheth, line 358: shouldn't be needed

evm/verify-contract.js
```diff
@@ -320,6 +330,46 @@ async function processCommand(config, chain, options) {
             break;
         }
 
+        case 'AxelarAmplifierGateway': {
+            let args = options.args;
+
+            if (args) {
+                args = JSON.parse(args);
+                options = { ...options, ...args };
+            }
```
milapsheth, line 339: avoid merging the two. read from `args` directly if that's where a value is defined. This becomes quite confusing for readability (if I see options.xyz later, I expect an `--xyz` flag to be defined)

milapsheth (changes requested):

evm/utils.js
```diff
@@ -1238,6 +1239,37 @@ async function relayTransaction(options, chain, contract, method, params, native
     );
 }
 
+async function getWeightedSigners(config, chain, options) {
+    printInfo(`Retrieving verifier addresses for ${chain.name} from Axelar network`);
```
milapsheth, line 1243: logging in helper wasn't removed

evm/verify-contract.js
```diff
@@ -47,7 +47,7 @@ async function processCommand(config, chain, options) {
         }
 
         case 'InterchainGovernance': {
-            const InterchainGovernance = require('@axelar-network/axelar-gmp-sdk-solidity/artifacts/contracts/governance/InterchainGovernance.sol/InterchainGovernance.json');
+            const InterchainGovernance = getContractJSON('InterchainGovernance');
```
milapsheth, line 50: duplicate code for calling getContractJSON for every case. Instead call it once at the top, we already have a variable containing the contractName. the factory can be created generically as well similarly

evm/verify-contract.js
milapsheth: @blockchainguyy That indicates an issue with the gmp-sdk repo. Can you modify the test contract so it's bytecode is different from the main one?

evm/verify-contract.js
```diff
@@ -320,6 +320,32 @@ async function processCommand(config, chain, options) {
...
+            verifyOptions.contractPath = 'contracts/gateway/AxelarAmplifierGateway.sol:AxelarAmplifierGateway';
+            await verifyContract(
+                env,
+                chain.name,
+                implementation,
+                [previousSignersRetention, domainSeparator, minimumRotationDelay],
+                verifyOptions,
+            );
+
+            verifyOptions.contractPath = 'contracts/gateway/AxelarAmplifierGatewayProxy.sol:AxelarAmplifierGatewayProxy';
```
milapsheth, line 343: let's add an identifier in this contract as well to distinguish it from the base proxy

blockchainguyy: There is already a contractId() function in this proxy. @milapsheth

evm/verify-contract.js
blockchainguyy: Seems to be a issue with Hardhat plugin. Getting this error `NomicLabsHardhatPluginError: More than one contract was found to match the deployed bytecode.`  on fantom but no issue in verifying on other chains like Avalanche, Optimism.
Instead of changing the test code as bytecode is different then the original contract, will raise issue with Hardhat team

evm/verify-contract.js
milapsheth: it's an internal method that's only used in the constructor of Proxy, and thus only gets included in the init code and not contract bytecode

evm/README.md
```diff
@@ -206,6 +206,12 @@ Verify TokenManagerProxy contract for ITS. `--tokenId` must be specified and `--
 node evm/verify-contract.js -e [env] -n [chain] -c TokenManagerProxy --dir /path/to/interchain-token-service --tokenId [tokenId]
 ```
 
+Verify AxelarAmplifierGateway contract. `--address` can be optionally specified (otherwise will default to the value from config).
+
+```bash
+node evm/verify-contract.js -e [env] -n [chain] -c AxelarAmplifierGateway --dir /path/to/axelar-gmp-sdk-solidity
```
milapsheth, line 212: missing instructions for how to prepare local gmp-sdk-solidity repo

evm/verify-contract.js
```diff
@@ -33,26 +33,22 @@ async function processCommand(config, chain, options) {
     printInfo('Contract address', options.address || chain.contracts[contractName]?.address);
 
     const contractConfig = chain.contracts[contractName];
+    const contractJson = getContractJSON(contractName);
+    let contract;
+    const contractFactory = await getContractFactoryFromArtifact(contractJson, wallet);
+    const excludedContracts = ['InterchainProposalSender', 'BurnableMintableCappedERC20', 'TokenManagerProxy', 'AxelarAmplifierGateway'];
```
milapsheth, line 39: this causes duplication of case-specific handling and should be avoided (otherwise we need to think of updating 2 places)

evm/verify-contract.js
```diff
@@ -33,26 +33,22 @@ async function processCommand(config, chain, options) {
...
+
+    if (!excludedContracts.includes(contractName)) {
+        contract = contractFactory.attach(options.address || contractConfig.address);
```
milapsheth, line 42:
```suggestion
        const contractAddress = options.address || contractConfig.address;
```
instead create a var, and create contract object only for the cases where necessary

re1ro (approved):

milapsheth (approved):

PR #272, opened 2024-06-17
feat(sui): new operators script

re1ro:
[AXE-4128]
* [x] `sui/operators`: new script for calling operators contract

### DIFF
--- a/sui/operators.js
+++ b/sui/operators.js
@@ -1,0 +1,63 @@
+const { saveConfig, printInfo } = require('../evm/utils');
+const { Command, Option } = require('commander');
+const { TransactionBlock } = require('@mysten/sui.js/transactions');
+
+const { addBaseOptions } = require('./cli-utils');
+const { getWallet, printWalletInfo, broadcast } = require('./sign-utils');
+const { loadSuiConfig } = require('./utils');
+
+async function callContractWithCapability(keypair, client, config, chain, args, options) {
+    if (!chain.contracts.operators) {
+        throw new Error('Operators package not found.');
+    }
+
+    const contractConfig = chain.contracts.operators;
+    const packageId = contractConfig.address;
+    const [contractId, functionName, ...functionArgs] = args;
+
+    const tx = new TransactionBlock();
+
+    const [operatorCap] = tx.moveCall({
+        target: `${packageId}::operators::borrow_cap`,
+        arguments: [tx.object(contractConfig.objects.operators), tx.object(contractConfig.objects.operator_cap), tx.pure(options.capId)],
+    });
+
+    tx.moveCall({
+        target: `${contractId}::${functionName}`,
+        arguments: [operatorCap, ...functionArgs.map((arg) => tx.pure(arg))],
+    });
+
+    await broadcast(client, keypair, tx);
+
+    printInfo('Contract called with borrowed capability');
+}
+
+async function mainProcessor(processor, args, options) {
+    const config = loadSuiConfig(options.env);
+
+    const [keypair, client] = getWallet(config.sui, options);
+    await printWalletInfo(keypair, client, config.sui, options);
+
+    await processor(keypair, client, config, config.sui, args, options);
+
+    saveConfig(config, options.env);
+}
+
+if (require.main === module) {
+    const program = new Command();
+
+    program.name('operators').description('Operators contract operations.');
+
+    const callContractCmd = program
+        .command('call-contract <contractId> <functionName> [functionArgs...]')
+        .description('Call a contract with a borrowed capability')
+        .addOption(new Option('--capId <capId>', 'ID of the capability to borrow'))
+        .action((contractId, functionName, functionArgs, options) =>
+            mainProcessor(callContractWithCapability, [contractId, functionName, ...functionArgs], options),
+        );
+
+    addBaseOptions(program);
+    addBaseOptions(callContractCmd);
+
+    program.parse();
+}


sui/operators.js
```diff
@@ -0,0 +1,141 @@
...
+        .addOption(new Option('--capId <capId>', 'ID of the GasCollectorCap to borrow'))
+        .action((receiver, amount, options) => mainProcessor(collectGas, [receiver, amount], options));
+
+    const refundGasCmd = program
+        .command('refund-gas <messageId> <receiver> <amount>')
+        .description('Refund gas from the gas service')
+        .addOption(new Option('--capId <capId>', 'ID of the GasCollectorCap to borrow'))
+        .action((messageId, receiver, amount, options) => mainProcessor(refundGas, [messageId, receiver, amount], options));
+
+    addBaseOptions(program);
```
blockchainguyy, line 135: This will give error that -p option is not specified when providing options via CLI instead of an env. As you are calling `addBaseOptions` with the program, as well as with other commands, it is trying to fetch `privateKey` options twice from CLI when running a command, this will not be a problem when base options are provided via an env file, as it can fetch the same options any number of times from env. So I believe `addBaseOptions` call to the program can be removed if it does not create a problem, so the base options can be provided via CLI as well.

npty:
I'll take over this PR

sui/operators.js
```diff
@@ -0,0 +1,280 @@
...
+async function removeCap(keypair, client, config, operatorsConfig, args, options) {
+    const [capId] = args;
+
+    const gasServiceAddress = config.sui.contracts.GasService.address;
+    const operatorsObjectId = operatorsConfig.objects.Operators;
+    const ownerCapObjectId = options.ownerCapId || operatorsConfig.objects.OwnerCap;
+
+    const tx = new Transaction();
+
+    tx.moveCall({
```
milapsheth, line 181: the removed cap needs to be transferred somewhere, since it can't be dropped at the end of the Tx

sui/operators.js
```diff
@@ -0,0 +1,280 @@
...
+    const gasCollectorCapId = gasCollectorCapObject.data.content.fields.value.fields.id.id;
+    return gasCollectorCapId;
+}
+
+// TODO: Fix `InvalidPublicFunctionReturnType { idx: 0 } in command 0` error
+async function collectGas(keypair, client, config, chain, args, options) {
+    const [amount] = args;
+    const receiver = options.receiver || keypair.toSuiAddress();
+    const gasServiceConfig = config.sui.contracts.GasService;
+    const operatorsConfig = config.sui.contracts.Operators;
```
milapsheth, line 53: can be passed in since it's common everywhere

sui/operators.js
milapsheth: let me know if this fixes it

npty: that works. thanks @milapsheth

sui/operators.js
```diff
@@ -0,0 +1,283 @@
...
+        throw new Error('Operators package not found.');
+    }
+
+    const operatorId = operatorsConfig.objects.Operators;
+    const gasCollectorCapId = await getGasCollectorCapId(client, gasServiceConfig, operatorsConfig);
+    const operatorCapId = await findOwnedObjectId(client, keypair.toSuiAddress(), `${operatorsConfig.address}::operators::OperatorCap`);
+
+    const tx = new Transaction();
+
+    const borrowedCap = tx.moveCall({
```
milapsheth, line 111:
```suggestion
    const [borrowedCap] = tx.moveCall({
```
do you need to explicitly destructure?

sui/operators.js
```diff
@@ -0,0 +1,283 @@
...
+        target: `${operatorsConfig.address}::operators::borrow_cap`,
+        arguments: [tx.object(operatorId), tx.object(operatorCapId), tx.pure(bcs.Address.serialize(gasCollectorCapId).toBytes())],
+        typeArguments: [`${gasServiceConfig.address}::gas_service::GasCollectorCap`],
+    });
+
+    tx.moveCall({
+        target: `${gasServiceConfig.address}::gas_service::refund`,
+        arguments: [
+            tx.object(gasServiceConfig.objects.GasService),
+            borrowedCap,
```
milapsheth, line 121:
```suggestion
            tx.object(borrowedCap),
```
maybe

sui/operators.js
```diff
@@ -0,0 +1,301 @@
...
+        ],
+        typeArguments: [`${gasServiceConfig.address}::gas_service::GasCollectorCap`],
+    });
+
+    const receipt = await broadcast(client, keypair, tx);
+
+    printInfo('Gas refunded', receipt.digest);
+}
+
+async function storeCap(keypair, client, config, chain, args, options) {
```
milapsheth, line 150:
```suggestion
async function storeCap(keypair, client, config, contactConfig, args, options) {
```
naming convention from evm scripts

sui/operators.js
```diff
@@ -0,0 +1,301 @@
...
+    const [capId] = args;
+    const gasCollectorCapConfig = config.sui.contracts.GasService;
```
milapsheth, line 152:
```suggestion
    const gasServiceConfig = config.sui.contracts.GasService;
```
This can be passed in as an arg as well

sui/operators.js
```diff
@@ -0,0 +1,301 @@
...
+    const gasServiceConfig = config.sui.contracts.GasService;
+    const operatorsConfig = config.sui.contracts.Operators;
+
+    if (!gasServiceConfig) {
+        throw new Error('Gas service package not found.');
+    }
+
+    if (!operatorsConfig) {
+        throw new Error('Operators package not found.');
+    }
```
milapsheth, line 58: check this in the caller method (so it applies to all calls in this file)

sui/operators.js
```diff
@@ -0,0 +1,301 @@
...
+        .action((options) => mainProcessor(storeCap, [], options));
+
+    const removeCapCmd = new Command('removeCap')
+        .command('removeCap <capId>')
+        .description('Remove a capability')
+        .addOption(new Option('--ownerCap <ownerCapId>', 'ID of the owner capability'))
+        .addOption(new Option('--receiver <receiver>', 'The removed cap receiver address'))
+        .action((capId, options) => mainProcessor(removeCap, [capId], options));
+
+    const refundCmd = new Command('refund')
```
milapsheth, line 284:
```suggestion
    const refundCmd = program.addCommand(new Command('refund')
```
Wrap each of this inside an addCommand to avoid the statements below

sui/operators.js
```diff
@@ -0,0 +1,301 @@
...
+    // Get the actual cap ID from the bag ID
+    const gasCollectorCapObject = await client.getObject({
+        id: gasCollectorBagId,
+        options: {
+            showContent: true,
+        },
+    });
+
+    // Extract and return the gas collector cap ID
+    const gasCollectorCapId = gasCollectorCapObject.data.content.fields.value.fields.id.id;
```
milapsheth, line 42: hmmmm, isn't there a structured helper to retrieve bag items?

sui/operators.js
```diff
@@ -0,0 +1,301 @@
+const { Command, Option } = require('commander');
+const { Transaction } = require('@mysten/sui/transactions');
+const { bcs } = require('@mysten/sui/bcs');
+const { printInfo, loadConfig } = require('../common/utils');
+const { operatorsStruct } = require('./types-utils');
+const { addBaseOptions, addOptionsToCommands, parseSuiUnitAmount } = require('./cli-utils');
+const { getWallet, printWalletInfo, broadcast } = require('./sign-utils');
+const { getBcsBytesByObjectId, findOwnedObjectId } = require('./utils');
+
+async function getGasCollectorCapId(client, gasServiceConfig, operatorsConfig) {
```
milapsheth, line 10: Do we need this? We store the gas collector cap id when deploying the gas service. Just read that? We could add a separate cmd in this script that checks if the gas collector (or user provided) cap id is actually stored in the Operators though

sui/operators.js
```diff
@@ -0,0 +1,301 @@
...
+        .addOption(new Option('--ownerCap <ownerCapId>', 'ID of the owner capability'))
+        .action((newOperatorAddress, options) => mainProcessor(addOperator, [newOperatorAddress], options));
+
+    const removeCmd = new Command('remove')
+        .command('remove <operatorAddress>')
+        .description('Remove an operator')
+        .addOption(new Option('--ownerCap <ownerCapId>', 'ID of the owner capability'))
+        .action((operatorAddress, options) => mainProcessor(removeOperator, [operatorAddress], options));
+
+    const collectGasCmd = new Command('collectGas')
```
milapsheth, line 264: Add an example to README, such as for collect gas

sui/operators.js
npty: do you mean in the sui sdk? no, i haven't seen any utils for this.

let me create one for it

sui/operators.js
milapsheth: @npty is this still an issue?

sui/README.md
```diff
@@ -147,6 +147,16 @@ The syntax is `node sui/gas-service.js payGas --amount <amount> <destinationChai
 node sui/gas-service.js payGas --amount 0.1 ethereum 0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05 0xba76c6980428A0b10CFC5d8ccb61949677A61233 0x1234
 ```
 
+Collect gas:
+
+Conditions:
+- The `GasCollectorCap` must be stored in the operator contract.
+- The sender must holds the `Operator` object.
```
milapsheth, line 154: `Operators` is a shared object so everyone can refer it. What you might have meant is that the sender needs to be an operator, i.e hold an `OperatorCap`
```suggestion
- The sender must be a whitelisted operator and hold the `OperatorCap` capability.
```

sui/README.md
```diff
@@ -147,6 +147,16 @@ The syntax is `node sui/gas-service.js payGas --amount <amount> <destinationChai
 node sui/gas-service.js payGas --amount 0.1 ethereum 0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05 0xba76c6980428A0b10CFC5d8ccb61949677A61233 0x1234
 ```
 
+Collect gas:
+
+Conditions:
+- The `GasCollectorCap` must be stored in the operator contract.
```
milapsheth, line 153:
```suggestion
- The `GasCollectorCap` must be stored in the gas service object config.
```
This should be saved in the config under the GasService objects (if it's not already being saved), and operators script should read from it instead of needing a copy.

sui/operators.js
milapsheth: i see, that's weird. Can you ask in the mysten channel if there's a shortcut/conventional way to do this?

sui/operators.js
```diff
@@ -0,0 +1,297 @@
...
+    const gasCollectorCapId = gasCollectorCapObject.data.content.fields.value.fields.id.id;
+    return gasCollectorCapId;
+}
+
+async function collectGas(keypair, client, gasServiceConfig, contractConfig, args, options) {
+    const [amount] = args;
+    const receiver = options.receiver || keypair.toSuiAddress();
+
+    const operatorId = contractConfig.objects.Operators;
+    const gasCollectorCapId = await getGasCollectorCapId(client, gasServiceConfig, contractConfig);
```
milapsheth, line 51: Since it can be read from the gas service config, helper use is not needed?

sui/operators.js
```diff
@@ -0,0 +1,297 @@
...
+        target: `${contractConfig.address}::operators::restore_cap`,
+        arguments: [
+            tx.object(operatorId),
+            tx.object(operatorCapId),
+            tx.pure(bcs.Address.serialize(gasCollectorCapId).toBytes()),
+            cap,
+            loanedCap,
+        ],
+        typeArguments: [`${gasServiceConfig.address}::gas_service::GasCollectorCap`],
+    });
```
milapsheth, line 130:
```suggestion
    operatorMoveCall(
        tx,
        `${gasServiceConfig.address}::gas_service::GasCollectorCap`,
        gasCollectorCapId,
        (cap) => {
            target: `${gasServiceConfig.address}::gas_service::refund`,
            arguments: [
                tx.object(gasServiceConfig.objects.GasService),
                cap,
                tx.pure.string(messageId),
                tx.pure.address(receiver),
                tx.pure.u64(amount),
            ],
        });
```
Perhaps create a helper method, say `operatorMoveCall(tx, capTypeName, capId, callWithLoanedCap)` to generalize executing arbitrary move call using operators contract.

sui/operators.js
```diff
@@ -0,0 +1,297 @@
...
+
+    const operatorId = contractConfig.objects.Operators;
+    const gasCollectorCapId = await getGasCollectorCapId(client, gasServiceConfig, contractConfig);
+    const operatorCapId = await findOwnedObjectId(client, keypair.toSuiAddress(), `${contractConfig.address}::operators::OperatorCap`);
+
+    const tx = new Transaction();
+
+    const [cap, loanedCap] = tx.moveCall({
+        target: `${contractConfig.address}::operators::loan_cap`,
+        arguments: [tx.object(operatorId), tx.object(operatorCapId), tx.pure(bcs.Address.serialize(gasCollectorCapId).toBytes())],
```
milapsheth, line 105: why doesn't tx.object work for gasCollectorCapId like the other fields?

sui/operators.js
```diff
@@ -0,0 +1,297 @@
...
+    });
+
+    tx.transferObjects([cap], capReceiver);
+
+    try {
+        const receipt = await broadcast(client, keypair, tx);
+
+        printInfo('Capability Removed', receipt.digest);
+    } catch (e) {
+        console.log('Error', e.message);
```
milapsheth, line 197: use printError instead?

sui/types-utils.js
```diff
@@ -83,12 +83,18 @@ const discoveryStruct = bcs.struct('Discovery', {
...
 });
 
-const tableStruct = bcs.struct('Table', {
+const bagStruct = bcs.struct('Bag', {
     id: UID,
     size: bcs.U64,
 });
 
-const bagStruct = bcs.struct('Bag', {
+const operatorsStruct = bcs.struct('Operators', {
```
milapsheth, line 91: this should be defined in cgp-sui? And we should remove duplicates here and use the cgp-sui ones

sui/README.md
npty: Sorry, I meant the `store_cap` in Operators contract must be called first.

sui/operators.js
npty: yes, we can read it from the config. do you think we can remove the `getGasCollectorCapId ` function or we can convert it to a util function to read the bag content id?

sui/operators.js
npty: I've asked him in the `mysten` channel. Here's Damir answer:

```
Frontend wise, while we still haven’t launched GraphQL fully, most of the solutions won’t be optimal. You’ll have to query the object, find the bag id, and then do either a direct request of a field, or query all dynamic fields and use this information to query the ones you need.

When not using dynamic fields, if you want to avoid long property chains with fields.xxx, you can add the `showBcs: true”` and then parse bcs bytes “manually”. I’ll be able to share an example of this later today.
```

sui/operators.js
npty: Yes, this is still an issue. will investigate.

sui/types-utils.js
npty: done https://github.com/axelarnetwork/axelar-cgp-sui/pull/105/files

sui/operators.js
npty: I've moved this into a util function

sui/README.md
```diff
@@ -150,8 +150,8 @@ node sui/gas-service.js payGas --amount 0.1 ethereum 0x6f24A47Fc8AE5441Eb47EFfC3
 Collect gas:
 
 Conditions:
-- The `GasCollectorCap` must be stored in the operator contract.
-- The sender must holds the `Operator` object.
+- The `GasCollectorCap` must be stored in the gas service object config.
```
milapsheth, line 153:
```suggestion
- The `GasCollectorCap` object id is read from the chain config, under gas service objects.
- The `GasCollectorCap` object must have been stored in the Operators contract via `store_cap` by the previous owner.
```
since that's what you were going for before

sui/operators.js
```diff
@@ -1,79 +1,44 @@
...
     const operatorId = contractConfig.objects.Operators;
+    const gasCollectorCapId = gasServiceConfig.objects.GasCollectorCap;
 
-    // Get and parse operator data
-    const operatorBytes = await getBcsBytesByObjectId(client, operatorId);
-    const parsedOperator = operatorsStruct.parse(operatorBytes);
-
-    // Get the capabilities bag ID
-    const bagId = parsedOperator.caps.id;
+    const tx = new Transaction();
```
milapsheth, line 12: Modify an existing tx instead? Makes it easier to add batching for operator calls if we want later

sui/operators.js
npty: Fixed https://github.com/axelarnetwork/axelar-contract-deployments/pull/272/commits/27f1db3542e75f4344fda6e80c975de13987e0b4

milapsheth (approved):

PR #278, opened 2024-06-20
feat(amplifier): store initial verifier set id when deploying amplifier gateway

eguajardo:
https://axelarnetwork.atlassian.net/browse/AXE-4289


evm/deploy-amplifier-gateway.js
```diff
@@ -243,6 +246,7 @@ async function deploy(config, chain, options) {
 
         const proxyDeploymentArgs = [implementation.address, owner, params];
         contractConfig.proxyDeploymentArgs = proxyDeploymentArgs;
+        contractConfig.initialVerifierSetId = verifierSetId;
```
milapsheth, line 249: can you set this value for the current devnet configs as well?

milapsheth (approved):

PR #291, opened 2024-07-10
chore(cosmwasm): update Multisig instantiation message

jcs47:
The amplifier Multisig contract now requires and `admin_address` field and a new way to encode the `blockExpiry` field. This PR updates the deployment script accordingly

### DIFF
--- a/cosmwasm/deploy-contract.js
+++ b/cosmwasm/deploy-contract.js
@@ -3,7 +3,17 @@
 require('dotenv').config();
 const { isNil } = require('lodash');
 
-const { printInfo, loadConfig, saveConfig, isString, isStringArray, isKeccak256Hash, isNumber, prompt, toBigNumberString } = require('../evm/utils');
+const {
+    printInfo,
+    loadConfig,
+    saveConfig,
+    isString,
+    isStringArray,
+    isKeccak256Hash,
+    isNumber,
+    prompt,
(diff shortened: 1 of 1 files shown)


cosmwasm/deploy-contract.js
```diff
@@ -36,7 +46,11 @@ const makeServiceRegistryInstantiateMsg = ({ governanceAccount }) => {
     return { governance_account: governanceAccount };
 };
 
-const makeMultisigInstantiateMsg = ({ governanceAddress, blockExpiry }, { Rewards: { address: rewardsAddress } }) => {
+const makeMultisigInstantiateMsg = ({ adminAddress, governanceAddress, blockExpiry }, { Rewards: { address: rewardsAddress } }) => {
+    if (!validateAddress(adminAddress)) {
```
milapsheth, line 50: you can simplify validation of multiple params via this approach
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/evm/gas-service.js#L252

milapsheth (approved):

cosmwasm/deploy-contract.js
jcs47: in this particular case `validateParameters` cannot be used because it is unable to validate a cosmos address. And since the function is part of the evm package, I don't think it makes sense to extend it for cosmos

PR #296, opened 2024-07-12
feat: add gas service commands

npty:
[AXE-4393](https://axelarnetwork.atlassian.net/browse/AXE-4393)

## Todo Items
- [x] Implement `pay_gas` command
- [x] Implement `add_gas` command
- [x] Implement `refund` command
- [x] Implement `collect_gas` command

[AXE-4393]: https://axelarnetwork.atlassian.net/browse/AXE-4393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sui/amount-utils.js
+++ b/sui/amount-utils.js
@@ -1,4 +1,4 @@
-import { ethers } from 'ethers';
+const { ethers } = require('ethers');
 
 // Convert formatted amount to atomic units (e.g. 1000000000). Default decimals is 9 for SUI
 function getAtomicAmount(amount, decimals = 9) {
--- a/sui/amount-utils.js
+++ b/sui/amount-utils.js
@@ -1,7 +1,7 @@
 const { ethers } = require('ethers');
 
 // Convert formatted amount to atomic units (e.g. 1000000000). Default decimals is 9 for SUI
-function getAtomicAmount(amount, decimals = 9) {
+function getUnitAmount(amount, decimals = 9) {
     return ethers.utils.parseUnits(amount, decimals).toBigInt();
 }
 
@@ -11,6 +11,6 @@ function getFormattedAmount(amount, decimals = 9) {
 }
 
 module.exports = {
-    getAtomicAmount,
+    getUnitAmount,
     getFormattedAmount,
 };
--- a/sui/amount-utils.js
+++ b/sui/amount-utils.js
@@ -1,0 +1,16 @@
+import { ethers } from 'ethers';
+
+// Convert formatted amount to atomic units (e.g. 1000000000). Default decimals is 9 for SUI
+function getAtomicAmount(amount, decimals = 9) {
+    return ethers.utils.parseUnits(amount, decimals).toBigInt();
+}
+
+// Convert atomic amount to formatted units (e.g. 1.0) with decimals. Default decimals is 9 for SUI
+function getFormattedAmount(amount, decimals = 9) {
+    return ethers.utils.formatUnits(amount, decimals);
+}
+
+module.exports = {
+    getAtomicAmount,
+    getFormattedAmount,
+};
--- a/sui/cli-utils.js
+++ b/sui/cli-utils.js
@@ -51,16 +51,20 @@ const addExtendedOptions = (program, options = {}) => {
     return program;
 };
 
-const addBaseOptionsToCommands = (program, options = {}) => {
+// `optionMethod` is a method such as `addBaseOptions`
+// `options` is an option object for optionMethod
+const addOptionsToCommands = (program, optionMethod, options) => {
     if (program.commands.length > 0) {
         program.commands.forEach((command) => {
-            addBaseOptions(command, options);
+            optionMethod(command, options);
         });
     }
+
+    optionMethod(program, options);
 };
 
 module.exports = {
     addBaseOptions,
     addExtendedOptions,
-    addBaseOptionsToCommands,
+    addOptionsToCommands,
 };
--- a/sui/cli-utils.js
+++ b/sui/cli-utils.js
@@ -2,7 +2,7 @@
 
 require('dotenv').config();
 
-const { Option } = require('commander');
+const { Option, InvalidArgumentError } = require('commander');
 const { getUnitAmount } = require('./amount-utils');
 
 const addBaseOptions = (program, options = {}) => {
@@ -68,7 +68,11 @@ const addOptionsToCommands = (program, optionMethod, options) => {
 // The user is expected to pass a full amount (e.g. 1.0), and this option parser will convert it to smallest units (e.g. 1000000000).
 // Note that this function will use decimals of 9 for SUI. So, other tokens with different decimals will not work.
 const parseSuiUnitAmount = (value, previous) => {
-    return getUnitAmount(value);
+    try {
+        return getUnitAmount(value);
+    } catch (error) {
+        throw new InvalidArgumentError('Please use the correct format (e.g. 1.0)');
+    }
 };
 
 module.exports = {
--- a/sui/cli-utils.js
+++ b/sui/cli-utils.js
@@ -3,6 +3,7 @@
 require('dotenv').config();
 
 const { Option } = require('commander');
+const { getUnitAmount } = require('./amount-utils');
 
 const addBaseOptions = (program, options = {}) => {
     program.addOption(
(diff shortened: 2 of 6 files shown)


sui/types-utils.js
```diff
@@ -21,6 +22,11 @@ const bytes32Struct = bcs.fixedArray(32, bcs.u8()).transform({
     output: (id) => hexlify(id),
 });
 
+const UID = bcs.fixedArray(32, bcs.u8()).transform({
```
milapsheth, line 25: also add types to cgp-sui repo since we're migrating to it

sui/gas-service.js
```diff
@@ -0,0 +1,251 @@
...
+            break;
+        case 'collect_gas':
+            printInfo('Action', 'Collect gas');
+            await collectGas(config, chain, args, options);
+            break;
+        case 'refund':
+            printInfo('Action', 'Refund gas');
+            await refund(config, chain, args, options);
+            break;
+    }
```
milapsheth, line 191: the function itself should be provided as input to mainProcessor to avoid this switch case

sui/gas-service.js
```diff
@@ -0,0 +1,251 @@
...
+    await processor(command, config, config.sui, args, options);
+    saveConfig(config, options.env);
+}
+
+if (require.main === module) {
+    const program = new Command();
+
+    program.name('gas-service').description('Interact with the gas service contract.');
+
+    const payGasProgram = new Command()
```
milapsheth, line 205:
```suggestion
    const payGasCmd = new Command()
```

sui/types-utils.js
npty: https://github.com/axelarnetwork/axelar-cgp-sui/pull/74

sui/gas-service.js
```diff
@@ -0,0 +1,251 @@
...
+        .option('--refund_address <refundAddress>', 'Refund address.')
+        .option('--params <params>', 'Params. Default is empty.')
+        .action((messageId, amount, options) => {
+            mainProcessor('add_gas', options, [messageId, amount], processCommand);
+        });
+
+    const collectGasProgram = new Command()
+        .command('collect_gas <amount>')
+        .description('Collect gas from the gas service contract.')
+        .option('--receiver <receiver>', 'Receiver address. Default is the sender address.')
```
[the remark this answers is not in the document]
npty: this one should be optional since we can use sender address as a default receiver?

sui/types-utils.js
```diff
@@ -46,6 +52,11 @@ const proofStruct = bcs.struct('Proof', {
     signatures: bcs.vector(bcs.vector(bcs.u8())),
 });
 
+const gasServiceStruct = bcs.struct('GasService', {
```
milapsheth, line 55: add this to cgp sui PR as well?

sui/gas-service.js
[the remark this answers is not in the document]
milapsheth: ah, sure. we can make it required if people forget to supply it

sui/amount-utils.js
```diff
@@ -0,0 +1,16 @@
+const { ethers } = require('ethers');
+
+// Convert formatted amount to atomic units (e.g. 1000000000). Default decimals is 9 for SUI
+function getAtomicAmount(amount, decimals = 9) {
```
milapsheth, line 4:
```suggestion
function getUnitAmount(amount, decimals = 9) {
```
we usually refer to it as unit

sui/cli-utils.js
```diff
@@ -51,7 +51,16 @@ const addExtendedOptions = (program, options = {}) => {
     return program;
 };
 
+const addBaseOptionsToCommands = (program, options = {}) => {
+    if (program.commands.length > 0) {
```
milapsheth, line 55: add to the program itself as well, so you can see these generic options when doing `node ./program --help`

Although, I was wondering if commander could do this for us instead of us manually having to add to each cmd

sui/gas-service.js
```diff
@@ -0,0 +1,220 @@
...
+    const walletAddress = keypair.toSuiAddress();
+
+    const gasServicePackageId = gasServiceConfig.address;
+
+    const refundAddress = options.refundAddress || walletAddress;
+    const params = options.params || '0x';
+
+    const [amount, destinationChain, destinationAddress, channelId, payload] = args;
+
+    const atomicAmount = getAtomicAmount(amount);
```
milapsheth, line 26:
```suggestion
    const unitAmount = getUnitAmount(amount);
```

sui/gas-service.js
```diff
@@ -0,0 +1,220 @@
...
+async function refund(keypair, client, gasServiceConfig, args, options) {
+    const walletAddress = keypair.toSuiAddress();
+
+    const gasServicePackageId = gasServiceConfig.address;
+    const gasServiceObjectId = gasServiceConfig.objects.GasService;
+
+    const [messageId, amount] = args;
+    const receiver = options.receiver || walletAddress;
+
+    const atomicAmount = getAtomicAmount(amount);
```
milapsheth, line 127: commander provides option processing, so we can do the conversion there instead of duplicating it
https://github.com/tj/commander.js?tab=readme-ov-file#custom-option-processing

we can use this for validation as well, and any other processing

sui/gas-service.js
```diff
@@ -0,0 +1,220 @@
...
+if (require.main === module) {
+    const program = new Command();
+
+    program.name('gas-service').description('Interact with the gas service contract.');
+
+    const payGasCmd = new Command()
+        .command('payGas <amount> <destinationChain> <destinationAddress> <channelId> <payload>')
+        .description('Pay gas for the new contract call.')
+        .option('--refundAddress <refundAddress>', 'Refund address. Default is the sender address.')
+        .option('--params <params>', 'Params. Default is empty.')
```
milapsheth, line 182: is the default empty, or undefined, i forget? if it's undefined we should set it explicitly to empty bytes

sui/cli-utils.js
```diff
@@ -51,7 +51,16 @@ const addExtendedOptions = (program, options = {}) => {
     return program;
 };
 
+const addBaseOptionsToCommands = (program, options = {}) => {
```
milapsheth, line 54:
```suggestion
// `options` is a method such as `addBaseOptions`
const addOptionsToCommands = (program, options) => {
```
make it more generic

milapsheth (approved):

sui/gas-service.js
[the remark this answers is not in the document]
npty: so, you think the receiver is better to always be specified? I thought the sender will be the receiver most of the time because you need to setup the right wallet that has gas collector cap.

sui/types-utils.js
npty: should we move the whole file to `axelar-cgp-sui`?

sui/cli-utils.js
npty: will apply base options to the main program as well. i didn't check the commander lib because it seems easier to just implement it than digging their doc. anyway, will do a bit more read.

sui/gas-service.js
npty: amount is positional param, so custom option processing doesn't work. If we convert amount into option, then it's currently hard to define default value for e.g. addGas function.

milapsheth: hmm, no need to define a default. it can just be a required option. i prefer reusing the validation over keeping it as a positional arg

sui/types-utils.js
milapsheth: we already have a bunch of these defined there, need to add the missing ones

sui/gas-service.js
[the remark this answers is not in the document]
milapsheth: sorry, i meant we can wait and see. i'm ok with the sender as default for now

milapsheth (approved):

sui/cli-utils.js
```diff
@@ -51,7 +52,28 @@ const addExtendedOptions = (program, options = {}) => {
...
+    }
+
+    optionMethod(program, options);
+};
+
+// Custom option processing for amount. https://github.com/tj/commander.js?tab=readme-ov-file#custom-option-processing
+// The user is expected to pass a full amount (e.g. 1.0), and this option parser will convert it to smallest units (e.g. 1000000000).
+// Note that this function will use decimals of 9 for SUI. So, other tokens with different decimals will not work.
+const parseSuiUnitAmount = (value, previous) => {
+    return getUnitAmount(value);
```
milapsheth, line 71: also validate?

PR #298, opened 2024-07-12
chore: update nonces for arbitrum mainnet

talalashraf:


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "lint": "eslint --fix '**/*.js'",
-    "prettier": "prettier --write '**/*.js' 'axelar-chains-config/info/*.json' 'package.json' 'evm/**/*.json' '.github/**/*.yaml'"
+    "prettier": "prettier --write '**/*.js' 'axelar-chains-config/info/*.json' 'package.json' 'evm/**/*.json' '.github/**/*.yaml' && find axelar-chains-config/info evm -name \"*.json\" -type f -print0 | xargs -0 -I {} sh -c 'echo \"\" >> \"{}\"'"
```
milapsheth, line 8: Check if prettierrc can be configured to do this instead

milapsheth (approved):

package.json
talalashraf: already did, tried other json parsers too, can't find anything that works. Officially prettier should be adding newlines EOF for everything but for some reason it fails for json files.

PR #306, opened 2024-07-18
feat: add script to upgrade sui gateway

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-3062


sui/upgrade-gateway.js
milapsheth, line 1: similar to EVM, add an `--upgrade` flag to gateway deploy script instead of a separate upgrade script?

sui/upgrade-gateway.js
```diff
@@ -0,0 +1,93 @@
...
+        ticket,
+    });
+
+    tx.moveCall({
+        target: '0x2::package::commit_upgrade',
+        arguments: [cap, receipt],
+    });
+
+    if (offline) {
+        sender ? tx.setSender(sender) : tx.setSender(keypair.toSuiAddress());
```
milapsheth, line 50: this conflicts with Foivos's PR where helpers have been created for upgrading. we should use that since it'll be a common pattern between contracts

re1ro (approved):

package.json
```diff
@@ -24,7 +24,7 @@
   "dependencies": {
     "@0xpolygonhermez/zkevm-commonjs": "github:0xpolygonhermez/zkevm-commonjs#v1.0.0",
     "@axelar-network/axelar-cgp-solidity": "6.3.1",
-    "@axelar-network/axelar-cgp-sui": "https://github.com/axelarnetwork/axelar-cgp-sui.git",
+    "@axelar-network/axelar-cgp-sui": "^0.2.0",
```
milapsheth, line 27:
```suggestion
    "@axelar-network/axelar-cgp-sui": "0.2.0",
```
we upgrade our repo dependencies deliberately, to avoid including an unintentional change in the contracts

sui/deploy-gateway.js
```diff
@@ -107,6 +106,40 @@ async function processCommand(config, chain, options) {
...
+        await upgradeGateway(chain, options, keypair, client);
+    }
+
+    if (options.offline) {
+        const { txFilePath } = options;
+        validateParameters({ isNonEmptyString: { txFilePath } });
+
+        const txB64Bytes = toB64(options.txBytes);
+
+        writeJSON({ status: 'PENDING', bytes: txB64Bytes }, txFilePath);
```
milapsheth, line 138: file content fields should be consistent with EVM upgrade. What about the signed tx?

sui/deploy-gateway.js
```diff
@@ -107,6 +106,40 @@ async function processCommand(config, chain, options) {
     printInfo('Gateway deployed', JSON.stringify(contractConfig, null, 2));
 }
 
+async function upgradeGateway(chain, options, keypair, client) {
+    const contractsConfig = chain.contracts;
+    const packageName = 'axelar_gateway';
+    const builder = new TxBuilder(client);
+    const chainConfig = contractsConfig[packageName];
+    chainConfig.objects = chainConfig.objects || {};
+    chainConfig.objects.upgradeCap = options.upgradeCap || chainConfig.objects.UpgradeCap;
```
milapsheth, line 115: this is confusing. why are there two cases for the same object key? why isn't the cap id stored during deployment to avoid requiring upgradeCap to be passed in?

sui/deploy-gateway.js
```diff
@@ -126,6 +159,14 @@ if (require.main === module) {
     program.addOption(new Option('--minimumRotationDelay <minimumRotationDelay>', 'minium delay for signer rotations (in ms)').default(0));
     program.addOption(new Option('--domainSeparator <domainSeparator>', 'domain separator').default(HashZero));
     program.addOption(new Option('--nonce <nonce>', 'nonce for the signer (defaults to HashZero)'));
+    program.addOption(new Option('--previousSigners <previousSigners>', 'number of previous signers to retain').default(0));
```
milapsheth, line 162: see the new commander syntax we're using in `sui/gateway.js`. Split deploy/upgrade into two commands and use positional args for mandatory args. We want to switch to a nested CLI command structure

sui/deploy-gateway.js
```diff
@@ -126,6 +159,14 @@ if (require.main === module) {
...
+    program.addOption(new Option('--upgradeCap <upgradeCap>', 'gateway UpgradeCap id'));
+    program.addOption(new Option('--upgrade', 'upgrade a deployed contract'));
+    program.addOption(new Option('--policy <policy>', 'new policy to upgrade'));
```
milapsheth, line 165: not clear what policy arg looks like, add example in help text and use readable .choices() restriction?

sui/deploy-gateway.js
```diff
@@ -87,6 +85,7 @@ async function processCommand(config, chain, options) {
             tx.pure.address(operator),
             separator,
             tx.pure(minimumRotationDelay),
+            tx.pure(options.previousSigners),
```
milapsheth, line 88: deployment needs to support specifying upgrade policy, created a [task](https://axelarnetwork.atlassian.net/browse/AXE-4489)

sui/deploy-gateway.js
milapsheth, line 162:
```suggestion
    program.addOption(new Option('--previousSigners <previousSigners>', 'number of previous signers to retain').default(15));
```
our evm default

milapsheth: Could also tackle this as part of merging gateway deploy functionality into deploy-contract, so it can be deploy/upgrade any contract

sui/deploy-gateway.js
milapsheth: can you fix this deployment script to save the object?

sui/deploy-gateway.js
milapsheth: Yeah, forgot that signing will need to be a separate step. Regardless, the combined signed tx can be stored within this same file, and we can mimic the following layout, instead of creating a separate file.
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/evm/deploy-gateway-v6.2.x.js#L488-L493

This should also work for normal non-multisig accounts, since we might not use a multisig for everything

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
-    });
+    program.name('deploy-contract').description('Deploy/Upgrade packages');
+
+    const deployCMD = program
+        .name('deploy')
+        .description('Deploy SUI modules')
+        .command('deploy <contractName>')
+        .addOption(new Option('--signers <signers>', 'JSON with the initial signer set').env('SIGNERS'))
+        .addOption(new Option('--operator <operator>', 'operator for the gateway (defaults to the deployer address)').env('OPERATOR'))
+        .addOption(new Option('--minimumRotationDelay <minimumRotationDelay>', 'minium delay for signer rotations (in ms)').default('0'))
```
milapsheth, line 203: switch to evm amplifier default. can you make this use seconds as well? mention this arg in the readme

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
-    addBaseOptions(program);
-
-    program.action((contractName, options) => {
-        mainProcessor(contractName, options, processCommand);
-    });
+    program.name('deploy-contract').description('Deploy/Upgrade packages');
+
+    const deployCMD = program
+        .name('deploy')
+        .description('Deploy SUI modules')
```
milapsheth, line 199:
```suggestion
        .description('Deploy a Sui package')
```
Modules exist within a package

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
+        .command('deploy <contractName>')
+        .addOption(new Option('--signers <signers>', 'JSON with the initial signer set').env('SIGNERS'))
+        .addOption(new Option('--operator <operator>', 'operator for the gateway (defaults to the deployer address)').env('OPERATOR'))
+        .addOption(new Option('--minimumRotationDelay <minimumRotationDelay>', 'minium delay for signer rotations (in ms)').default('0'))
+        .addOption(new Option('--domainSeparator <domainSeparator>', 'domain separator').default(HashZero))
+        .addOption(new Option('--nonce <nonce>', 'nonce for the signer (defaults to HashZero)'))
+        .addOption(new Option('--previousSigners <previousSigners>', 'number of previous signers to retain').default('15'))
+        .addOption(
+            new Option('--policy <policy>', 'upgrade policy for upgrade cap: For example, use "0" to allow all types of upgrades')
+                .choices(['0', '128', '192'])
```
milapsheth, line 209: use strings for readable policy names. Convert to an int internally (maybe sui sdk has a helper for conversion?)

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
+                .default('0'),
+        )
+        .action((contractName, options) => {
+            mainProcessor([contractName], options, deploy);
+        });
+
+    const upgradeCMD = program
+        .name('upgrade')
+        .description('Upgrade SUI modules')
```
milapsheth, line 218:
```suggestion
        .description('Upgrade a Sui package')
```

sui/utils.js
```diff
@@ -77,15 +77,20 @@ const deployPackage = async (packageName, client, keypair, options = {}) => {
...
-    return published.publishTxn.objectChanges.find((change) => change.objectType === `${packageId}::${packageName}::${contractName}`);
+const getObjectIdsByObjectTypes = (txn, objectTypes) => {
+    const objectIds = [];
+
+    for (const objectType of objectTypes) {
+        objectIds.push(txn.objectChanges.find((change) => change.objectType === objectType).objectId);
+    }
+
+    return objectIds;
 };
```
milapsheth, line 88:
```suggestion
const getObjectIdsByObjectTypes = (txn, objectTypes) => 
  objectTypes.map((objectType) => {
    const objectId = txn.objectChanges.find(change => change.objectType === objectType)?.objectId;
    if (!objectId) {
      throw new Error(`No object found for type: ${objectType}`);
    }
  });
```
handle undefined cases. And use `map` to make it more functional

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
+            tx.moveCall({
+                target: `${packageId}::gateway::setup`,
+                arguments: [
+                    tx.object(creatorCap),
+                    tx.pure.address(operator),
+                    separator,
+                    tx.pure.u64(minimumRotationDelay),
+                    tx.pure.u64(options.previousSigners),
+                    tx.pure(bcs.vector(bcs.u8()).serialize(encodedSigners).toBytes()),
+                    tx.object('0x6'),
```
milapsheth, line 111: magic constant
```suggestion
                    tx.object(suiClockId),
```

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
+
+                tx.moveCall({
+                    target: `0x2::package::${upgradeType}`,
+                    arguments: [tx.object(upgradeCap)],
+                });
+            }
+
+            const result = await broadcast(client, keypair, tx);
+
+            printInfo('Setup transaction digest: ', result.digest);
```
milapsheth, line 126:
```suggestion
            printInfo('Setup transaction digest', result.digest);
```
printInfo handles this already

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
     printInfo(`${contractName} deployed`, JSON.stringify(contractConfig, null, 2));
 }
 
-async function mainProcessor(contractName, options, processor) {
+async function upgrade(contractName, policy, config, chain, options) {
+    const [keypair, client] = getWallet(chain, options);
+    const { packageDependencies } = options;
+    options.policy = policy;
+
+    await printWalletInfo(keypair, client, chain, options);
```
milapsheth, line 154: duplication, do this in the mainProcessor, and pass in the keypair to deploy/upgrade

sui/deploy-contract.js
```diff
@@ -23,49 +54,179 @@ async function processCommand(contractName, config, chain, options) {
...
+        }
+    }
+
+    const builder = new TxBuilder(client);
+    await upgradePackage(client, keypair, contractName, packageConfig, builder, options);
+}
+
+async function mainProcessor(args, options, processor) {
     const config = loadSuiConfig(options.env);
-    await processor(contractName, config, config.sui, options);
```
milapsheth, line 51: perform contractName validation here, since it'll be provided for any processor method

PR #323, opened 2024-07-29
feat(sui): support all contract deployment in deploy-contract script

npty:
# Description

[AXE-4558](https://axelarnetwork.atlassian.net/browse/AXE-4558)

This PR expands contracts support for both deployment and upgrade including `Test`, `AxelarGateway`, and `GasService` contracts. As a generalized script to deploy and upgrade contract, it's mostly focusing on providing a clear structure to support additional contracts. 

There're few different sections that needs to modify/add when we need to support new contracts:

#### 1. Package Dirs

This constant contains a list of folder names for each contract

```js
const PACKAGE_DIRS = ['gas_service', 'test', 'axelar_gateway'];
```

The package names will be read from the `Move.toml` file programmatically based on given `PACKAGE_DIRS`. They will be used as a command name for the sub-commands in the deploy and upgrade commands dynamically.

#### 2. Post-Deployment Functions
These functions serve the purposes such as updating chain config, submitting additional setup transactions. Each function will be called based on the package name in a switch-case condition as follows:

```js
switch (packageName) {
    case 'GasService':
        await postDeployGasService(published, chain);
        break;
    case 'AxelarGateway':
        await postDeployAxelarGateway(published, keypair, client, config, chain, options);
        break;
    case 'Test':
        await postDeployTest(published, config, chain, options);
        break;
    default:
        throw new Error(`${packageName} is not supported.`);
}
```

#### 3. Deploy Options

Each contract has its own set of arguments. They're defined in the `cli-utils.js` file and added to the program conditionally based on the `packageName` as follows:

```js
const getDeployGatewayOptions = () => {
    return [
        new Option('--signers <signers>', 'JSON with the initial signer set').env('SIGNERS'),
        new Option('--operator <operator>', 'operator for the gateway (defaults to the deployer address)').env('OPERATOR'),
        new Option('--minimumRotationDelay <minimumRotationDelay>', 'minium delay for signer rotations (in second)')
            .argParser((val) => parseInt(val) * 1000)
            .default(24 * 60 * 60),
        new Option('--domainSeparator <domainSeparator>', 'domain separator'),
        new Option('--nonce <nonce>', 'nonce for the signer (defaults to HashZero)'),
        new Option('--previousSigners <previousSigners>', 'number of previous signers to retain').default('15'),
        new Option('--policy <policy>', 'upgrade policy for upgrade cap: For example, use "any_upgrade" to allow all types of upgrades')
            .choices(['any_upgrade', 'code_upgrade', 'dep_upgrade'])
            .default('any_upgrade'),
    ];
};

const addDeployOptions = (program) => {
    switch (program.name()) {
        case 'AxelarGateway':
            getDeployGatewayOptions().forEach((option) => program.addOption(option));
            break;
        case 'GasService':
        case 'Test':
            break;
        default:
            throw new Error(`Unsupported package: ${program.name()}. `);
    }

    return program;
};
```

## Changes

- Refactor the deployment scripts structure.
- Moved `getSigners` and `getChannelId` methods from `sui/deploy-contract.js` to `sui/utils.js` to reduce the complexity.
- Moved `getDomainSeparator` from `evm/utils` to `common/utils` as we have to use it when deploys the AxelarGateway contract on Sui network as well. Also modifying a bit to allow zero hex as a default value when the env is `local`

[AXE-4558]: https://axelarnetwork.atlassian.net/browse/AXE-4558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Usage

1. Deploy Axelar Gateway contract:
```bash
node sui/deploy-contract.js deploy AxelarGateway --signers wallet --nonce test
```

2. Deploy Gas Service contract:
```bash
node sui/deploy-contract.js deploy GasService
```

3. Upgrade Axelar Gateway contract:
```bash
node sui/deploy-contract.js upgrade AxelarGateway <policy>
```


sui/deploy-contract.js
```diff
@@ -5,156 +5,206 @@ const { toB64 } = require('@mysten/sui/utils');
...
+ * `node_modules/@axelar-network/axelar-cgp-sui/move`
+ *
+ * Each string in this array corresponds to a folder name within that path.
+ *
+ * To deploy a new package:
+ * 1. Add the new package's folder name to this array
+ * 2. Ensure the corresponding folder exists in the specified path
+ *
+ */
+const PACKAGE_DIRS = ['gas_service', 'test', 'axelar_gateway'];
```
milapsheth, line 39: read this dynamically by going over all dirs instead? we do something kind of similar for EVM contracts
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/evm/utils.js#L811

sui/deploy-contract.js
```diff
@@ -5,156 +5,206 @@ const { toB64 } = require('@mysten/sui/utils');
...
+        case 'GasService':
+            await postDeployGasService(published, chain);
             break;
-        }
-
-        default: {
-            throw new Error(`${contractName} is not supported.`);
-        }
+        case 'AxelarGateway':
+            await postDeployAxelarGateway(published, keypair, client, config, chain, options);
```
milapsheth, line 183: define a map from contract name to postDeploy script. Avoids the switch case, and standardize input args

sui/deploy-contract.js
```diff
@@ -5,156 +5,206 @@ const { toB64 } = require('@mysten/sui/utils');
...
+            separator,
+            tx.pure.u64(minimumRotationDelay),
+            tx.pure.u64(options.previousSigners),
+            tx.pure(bcs.vector(bcs.u8()).serialize(encodedSigners).toBytes()),
+            tx.object(suiClockAddress),
+        ],
+    });
+
+    if (policy !== 'any_upgrade') {
+        const upgradeType = UPGRADE_POLICIES[policy];
```
milapsheth, line 141: handle upgrade policy config for all contracts generically instead of just the gateway. An override for the owner for the upgrade cap can be passed in as well, along with the upgrade policy.

sui/deploy-contract.js
npty: Yeah, i've thought about it. I've set it manually because I'd like to avoid having additional checks in the program flows to ensure we've already handled that contract deployment, otherwise we throw error early. 

So i think it's cleaner to define supported packages at the beginning, outside of the program flow. Then, dynamically generated the sub-commands instead based on that values instead. The main idea of this pattern will also align with your suggestion to define a map to handle deployment script, instead of switch-case.

sui/deploy-utils.js
```diff
@@ -59,130 +70,49 @@ async function upgradePackage(client, keypair, packageName, packageConfig, build
...
-        await upgradePackage(client, keypair, packageName, packageConfig, builder, options);
-    } else {
-        await deployPackage(chain, client, keypair, packageName, packageConfig, builder, options);
-    }
-
-    if (offline) {
-        validateParameters({ isNonEmptyString: { txFilePath } });
-
-        const txB64Bytes = toB64(options.txBytes);
+const getDeployGatewayOptions = () => {
```
milapsheth, line 82: this should be moved to deploy-contract where gateway deployment is defined instead of utils

sui/deploy-contract.js
```diff
@@ -186,47 +250,51 @@ async function mainProcessor(args, options, processor) {
...
+
+    // Add 3rd level commands to 2nd level command `deploy`
+    deployContractCmds.forEach((cmd) => deployCmd.addCommand(cmd));
+
+    // 3rd level commands for `upgrade`
+    const upgradeContractCmds = supportedPackages.map((supportedPackage) => {
+        const { packageName } = supportedPackage;
+        return new Command(packageName)
+            .description(`Deploy ${packageName} contract`)
+            .command(`${packageName} <policy>`)
```
milapsheth, line 278: add policy as a flag with defined choices. it can be default to any upgrade

sui/deploy-utils.js
npty: hmm yeah, but need to move the `addDeployOptions` function (it was defined below this function) to deploy-contract as well because then i have to import `getDeployGatewayOptions` from `deploy-contract.js` in the utils file which is weird.

sui/utils.js
```diff
@@ -80,6 +83,28 @@ const deployPackage = async (packageName, client, keypair, options = {}) => {
...
 };
 
+const findPublishedObject = (published, packageDir, contractName) => {
+    const packageId = published.packageId;
+    return published.publishTxn.objectChanges.find((change) => change.objectType === `${packageId}::${packageDir}::${contractName}`);
+};
+
+const readMovePackageName = (moveDir) => {
+    try {
+        const moveToml = fs.readFileSync(`${__dirname}/../node_modules/@axelar-network/axelar-cgp-sui/move/${moveDir}/Move.toml`, 'utf8');
```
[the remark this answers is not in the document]
npty: yeah, we could add that

common/utils.js
```diff
@@ -15,7 +15,19 @@ const {
 const { normalizeBech32 } = require('@cosmjs/encoding');
 
 function loadConfig(env) {
-    return require(`${__dirname}/../axelar-chains-config/info/${env}.json`);
+    const config = require(`${__dirname}/../axelar-chains-config/info/${env}.json`);
+
+    if (!config.sui) {
```
milapsheth, line 20: we should do this generically for any chain, or remove this. I'd suggest just adding this example local.json to the readme instead of creating it dynamically

sui/deploy-contract.js
```diff
@@ -38,6 +37,30 @@ const {
...
+    Operators: postDeployOperators,
+    AxelarGateway: postDeployAxelarGateway,
+};
+
+/**
+ * Command Options Mapping
+ *
+ * This object maps each package name to a function that returns an array of command options.
+ */
+const CMD_OPTIONS = {
```
milapsheth, line 57: use a single multi level map instead of two separate ones

sui/deploy-contract.js
```diff
@@ -92,18 +117,20 @@ async function postDeployTest(published, config, chain, options) {
...
-      Operators: operatorsObjectId,
-      OwnerCap: ownerCapObjectId,
+        Operators: operatorsObjectId,
+        OwnerCap: ownerCapObjectId,
     };
 }
 
-async function postDeployAxelarGateway(published, keypair, client, config, chain, options) {
+async function postDeployAxelarGateway(published, args) {
+    const { keypair, client, config, chain, options } = args;
```
milapsheth, line 133: pass all the same args to each post deploy method instead of wrapping it inside args?

sui/deploy-contract.js
```diff
@@ -178,30 +205,19 @@ async function postDeployAxelarGateway(published, keypair, client, config, chain
...
-            break;
-        case 'Operators':
-            await postDeployOperators(published, chain);
-            break;
-        default:
-            throw new Error(`${packageName} is not supported.`);
-    }
+    // Execute post-deployment function
+    const executePostDeploymentFn = POST_DEPLOY_FUNCTIONS[packageName];
+    executePostDeploymentFn(published, { keypair, client, config, chain, options });
```
milapsheth, line 220:
```suggestion
    await executePostDeploymentFn(published, { keypair, client, config, chain, options });
```

milapsheth (approved):

sui/deploy-contract.js
npty: will do. i just don't want to declare a function with long positional args, but actually we did that everywhere across the codebase 😂

PR #332, opened 2024-08-07
feat(evm)!: support its v2 deployment

milapsheth:


cosmwasm/utils.js
```diff
@@ -450,6 +447,27 @@ const makeAxelarnetGatewayInstantiateMsg = (config, chainName) => {
...
+
+    if (!validateAddress(axelarnetGatewayAddress)) {
+        throw new Error('Missing or invalid AxelarnetGateway.address in axelar info');
+    }
+
+    return {
+        governance_address: governanceAddress,
+        admin_address: adminAddress,
+        axelarnet_gateway_address: axelarnetGatewayAddress,
+        its_addresses: {},
```
jcs47, line 467: shouldn't we allow for this field to be populated via the config file?

package.json
```diff
@@ -26,7 +26,7 @@
     "@axelar-network/axelar-cgp-solidity": "6.3.1",
     "@axelar-network/axelar-cgp-sui": "^0.4.1",
     "@axelar-network/axelar-gmp-sdk-solidity": "5.10.0",
-    "@axelar-network/interchain-token-service": "1.2.4",
+    "@axelar-network/interchain-token-service": "file:../interchain-token-service",
```
jcs47, line 29: is this temporary?

milapsheth: yep

milapsheth:
> >Configure remote chains to route via ITS hub (run this for each remote chain you want to connect to):
> node evm/its.js -n [chain] --action setTrustedAddress --trustedChain [remote chain, e.g. avalanche-fuji, xrpl] --trustedAddress hub
> 
> @milapsheth Should we make it `trustedChains` and allow multiple chains as an input?

Yeah, would be a nice improvement. We should add a specific new action that sets the trusted address to `hub` for one or more chains and avoids user error. This action can stay as-is since different chains can have different addresses if using the legacy approach.

evm/interchainTokenFactory.js
```diff
@@ -137,6 +137,9 @@ async function processCommand(config, chain, options) {
                 gasOptions,
             );
 
+            const tokenId = await interchainTokenFactory.interchainTokenId(wallet.address, deploymentSalt);
+            printInfo('tokenId', tokenId);
+
             await handleTx(tx, chain, interchainTokenService, options.action, 'TokenManagerDeployed', 'InterchainTokenDeploymentStarted');
```
eguajardo, line 143: shouldn't `handleTx` check for `InterchainTokenDeployed` instead of `InterchainTokenDeploymentStarted` when not deploying remotely? I don't see the `InterchainTokenDeploymentStarted` event in a `deployInterchainToken` transaction (https://testnet.snowtrace.io/tx/0x0125c0d4006cc2d8fda1ae7c0c3f27507b2fdcc3ea90685b0c94b4b9006b1c2a/eventlog?chainid=43113) and I don't get the warning either

evm/gateway.js
```diff
@@ -508,6 +511,9 @@ if (require.main === module) {
 
     program.addOption(new Option('--payload <payload>', 'gmp payload'));
     program.addOption(new Option('--commandID <commandID>', 'execute command ID'));
+    program.addOption(new Option('--messageId <messageId>', 'GMP call message ID'));
+    program.addOption(new Option('--sourceChain <sourceChain>', 'GMP source chain'));
+    program.addOption(new Option('--sourceAddress <sourceAddress>', 'GMP source address'));
     program.addOption(new Option('--destination <destination>', 'GMP destination address'));
     program.addOption(new Option('--destinationChain <destinationChain>', 'GMP destination chain'));
```
jcs47, line 518: why not merge all of this into a single flag that receives a json object?

milapsheth: this will be cleaned up to use nested CLI cmds (similar to sui/stellar scripts). Using JSON input is a bit worse UX wise when using manually

jcs47 (approved):

PR #333, opened 2024-08-07
chore: update Readme and remove duplicate entries

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-4615


evm/check-ownership-request.js
```diff
@@ -128,25 +198,39 @@ async function verifyChains(config, tokenAddress, destinationChains) {
...
         } catch {
-            // printWarn(`No Interchain token found for address ${tokenAddress} on chain ${chain}`);
             invalidDestinationChains.push(chain);
         }
     }
 
     return invalidDestinationChains;
 }
 
+function updateCSVFile(filePath, columnNames, data) {
```
milapsheth, line 208: An existing library should be used instead of writing your own csv writer. Ideally, it allows you to use structured objects instead of dealing with clunky array indices

milapsheth (approved):

PR #336, opened 2024-08-14
chore: iterate through all owner objects to find target 

sdaveas:
`getOwnedObjects` returns a limited number of objects per page (50 by default), and our target might not be found in the first one. That's why we should iterate through all pages of the result until no next page is found, or we track our target object.

### DIFF
--- a/sui/utils.js
+++ b/sui/utils.js
@@ -146,14 +146,23 @@ const getSigners = async (keypair, config, chain, options) => {
 };
 
 const findOwnedObjectId = async (client, ownerAddress, objectType) => {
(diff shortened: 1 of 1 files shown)


npty:
Thank you for the PR @sdaveas 🙏🏻 

I'll close this PR since I will utilize the `paginateAll` function in [this PR](https://github.com/axelarnetwork/axelar-contract-deployments/pull/255/files#diff-ec2e04c7d99cff29e4cb780360ce51c57cf8c041723c9ec49930367af5addafc) instead.

PR #338, opened 2024-08-19
chore: cleanup and refactor

npty:
# Description

This PR refactors the sui scripts by separating the `utils` files and the command program files.

# Changes

- All utils files are move into `sui/utils` folder and added the `index.js` file which exports everything there.
- All command program files now imports the utils files from the single location, `./utils`, instead of `./types-utils`, `cli-utils`, `utils`, etc.
- Removed unused file such as `deploy-test.js` because the `Test` contract deployment now becomes a part of `deploy-contract.js`.
- Renamed `deploy-utils.js` to `upgrade-utils.js` because the exported functions all involve with the contract upgrade.
# Test

All commands are tested manually to check if there's any error or unexpected outcome.


sui/utils/index.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
milapsheth, line 1: I prefer to not flatten it so it's more clear where the import comes from, but not a strong opinion if you think this is more standard

milapsheth (approved):

sui/utils/index.js
npty: It can be commonly found in many js projects, but it's not a strong pattern. We need to compare the trade-offs for each project. I agree that it's a bit less clear to find the source file for imported method. However, In VSCode, if you click at the imported method, it will redirect you to the original file, not the index file. That also works for some methods in GitHub as well.

But the benefits that I think it's worth to add is, when you're working on a script, and you want to import some utils functions, you can get all available options from autocomplete in the `require('./utils')` statement. This way, you don't have to remember where the function is located.

PR #341, opened 2024-08-19
feat(amplifier): refactor amplifier script options

eguajardo:
https://axelarnetwork.atlassian.net/browse/AXE-4668

### DIFF
--- a/cosmwasm/cli-utils.js
+++ b/cosmwasm/cli-utils.js
@@ -1,0 +1,41 @@
+'use strict';
+
+require('dotenv').config();
+
+const { addExtendedOptions } = require('../common');
+const { governanceAddress } = require('./utils');
+
+const { Option } = require('commander');
+
+const addCommonAmplifierOptions = (program, options = {}) => {
+    const ops = {
+        ignoreParallel: true,
+        ignoreSaveChainSeparately: true,
+        ignoreGasOptions: true,
+        ignoreChainNames: true,
+        ignorePrivateKey: true,
+        ignoreVerify: true,
+        contractName: true,
+        salt: true,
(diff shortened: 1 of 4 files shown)


cosmwasm/cli-utils.js
```diff
@@ -0,0 +1,41 @@
...
+        ignoreGasOptions: true,
+        ignoreChainNames: true,
+        ignorePrivateKey: true,
+        ignoreVerify: true,
+        contractName: true,
+        salt: true,
+        ...options,
+    };
+
+    addExtendedOptions(program, ops);
```
milapsheth, line 23: i'd say just use baseOptions, extendedOptions is more evm specific. Create `addDeployOptions` perhaps that adds extra params like salt, instantiate2, artifactPath etc.

cosmwasm/cli-utils.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+require('dotenv').config();
+
+const { addExtendedOptions } = require('../common');
+const { governanceAddress } = require('./utils');
+
+const { Option } = require('commander');
+
+const addCommonAmplifierOptions = (program, options = {}) => {
```
milapsheth, line 10:
```suggestion
const addAmplifierOptions = (program, options = {}) => {
```
common is implied

cosmwasm/cli-utils.js
milapsheth: Since we refactored the utils to add `addEnvOptions`, just define the other options directly instead of using addBaseOptions/addExtendedOptions

milapsheth (approved):

PR #344, opened 2024-08-22
chore: reuse addEnvOptions

npty:
# Description

This PR removes the duplicated env options

### DIFF
--- a/common/cli-utils.js
+++ b/common/cli-utils.js
@@ -11,10 +11,19 @@ const CHAIN_CONFIG_PATH = `${__dirname}/../axelar-chains-config/info`;
 // A list of available chain environments which are the names of the files in the CHAIN_CONFIG_PATH
 const CHAIN_ENVIRONMENTS = fs.readdirSync(CHAIN_CONFIG_PATH).map((chainName) => chainName.split('.')[0]);
 
-const addBaseOptions = (program, options = {}) => {
+const addEnvOption = (program, defaultValue) => {
     program.addOption(
-        new Option('-e, --env <env>', 'environment').choices(CHAIN_ENVIRONMENTS).default('testnet').makeOptionMandatory(true).env('ENV'),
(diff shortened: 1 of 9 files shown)


common/cli-utils.js
```diff
@@ -92,7 +104,14 @@ const addExtendedOptions = (program, options = {}) => {
     return program;
 };
 
+if (require.main === module) {
+    addBaseOptions(new Command());
```
milapsheth, line 108: I don't want to add it by default, and let the script decide since they might want to customize it a bit

cosmwasm/deploy-contract.js
```diff
@@ -114,6 +115,11 @@ const programHandler = () => {
 
     addCommonAmplifierOptions(program);
 
+    addEnvOption(program);
```
milapsheth, line 118: add amplifier options should have this internally

cosmwasm/submit-proposal.js
```diff
@@ -251,6 +252,24 @@ const programHandler = () => {
 
     addCommonAmplifierOptions(program, { predictOnly: true });
 
+    // TODO: combine deploy-contract and submit-proposal options to remove duplicates
+    addEnvOption(program, 'devnet-amplifier');
```
milapsheth, line 256: Same

common/cli-utils.js
```diff
@@ -92,7 +104,14 @@ const addExtendedOptions = (program, options = {}) => {
...
+}
+
 module.exports = {
+    CHAIN_CONFIG_PATH,
+    CHAIN_ENVIRONMENTS,
```
milapsheth, line 113: These constants shouldn't be exposed anymore

milapsheth (approved):

common/cli-utils.js
npty: this was mistakenly pushed. i added this for testing locally.

PR #348, opened 2024-08-27
feat(cosmwasm): support for deploying the axelarnet-gateway contract

jcs47:


cosmwasm/utils.js
```diff
@@ -420,6 +420,28 @@ const makeMultisigProverInstantiateMsg = (config, chainName) => {
     };
 };
 
+const makeAxelarnetGatewayInstantiateMsg = (config, chainName) => {
+    const {
+        axelar: { contracts },
+    } = config;
+    const chainConfig = getChainConfig(config, chainName);
+
+    const { axelarId: chainId } = chainConfig;
```
milapsheth, line 429: let's avoid introducing multiple terms for the same value (chainId means something different for external chains). Instead of destructuring all fields, use it as `chainConfig.axelarId`. It's more readable.

cosmwasm/utils.js
```diff
@@ -420,6 +420,28 @@ const makeMultisigProverInstantiateMsg = (config, chainName) => {
...
+        Router: { address: routerAddress },
+    } = contracts;
+
+    if (!validateAddress(routerAddress)) {
+        throw new Error('Missing or invalid Router.address in axelar info');
+    }
+
+    return {
+        router_address: routerAddress,
+        chain_name: chainId,
```
milapsheth, line 441: Lowercase it as well. we don't support upper case chain names. So it'll end up being `axelarnet`

milapsheth (approved):

common/utils.js
```diff
@@ -57,12 +57,16 @@ function printLog(log) {
...
+
 const isNonEmptyString = (arg) => {
-    return typeof arg === 'string' && arg !== '';
+    return isString(arg) && arg !== '';
 };
 
-const isString = (arg) => {
-    return typeof arg === 'string';
+const isStringLowercase = (arg) => {
+    return isString(arg) && arg === arg.toLowerCase();
```
milapsheth, line 69:
```suggestion
    return isNonEmptyString(arg) && arg === arg.toLowerCase();
```
should check for non-empty by default

PR #339, opened 2024-08-19
chore(sui): gh workflow for testing sui commands

npty:
# Description

[AXE-4664](https://axelarnetwork.atlassian.net/browse/AXE-4664)

This PR adds the test workflow for sui which follows the same pattern as evm test workflow.

## Todo

- [x] Fix error in rotate signers command
- [x] Fix error in execute incoming gmp contract call command
- [x] Add multisig tests

[AXE-4664]: https://axelarnetwork.atlassian.net/browse/AXE-4664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #349, opened 2024-08-28
feat(sui): add init multisig command

npty:
# Description

This PR added the `init` command to `multisig.js`, in order to make [the multisig flow](https://github.com/axelarnetwork/axelar-contract-deployments/pull/339#discussion_r1733924184) smoother.

The init command receives the following params from the cli options: 
- `base64PublicKeys`
- `threshold`
- `schemaTypes`
- `weights` 

Then, it constructs the `multisig` object save into the config file.

### DIFF
--- a/.github/workflows/test-sui.yaml
+++ b/.github/workflows/test-sui.yaml
@@ -139,7 +139,44 @@ jobs:
 
       ###### Command: Multisig ######
 
-      # TODO: Add multisig tests
+      - name: Init Multisig
+        run: |
+          sui client new-address secp256k1 multisig1
+          sui client new-address secp256k1 multisig2
+          KEY_1=$(sui keytool export --key-identity multisig-1 --json | jq .key.publicBase64Key | sed 's/"//g')
+          KEY_2=$(sui keytool export --key-identity multisig-2 --json | jq .key.publicBase64Key | sed 's/"//g')
+          ADDRESS_1=$(sui keytool export --key-identity multisig-1 --json | jq .key.suiAddress | sed 's/"//g')
+          ADDRESS_2=$(sui keytool export --key-identity multisig-2 --json | jq .key.suiAddress | sed 's/"//g')
+          MULTISIG_ADDRESS=$(sui keytool multi-sig-address --pks $KEY_1 $KEY_2 --weights 1 1 --threshold 1 --json | jq .multisigAddress | sed 's/"//g')
+          node sui/multisig.js --action init --threshold 1 --base64PublicKeys $KEY_1 $KEY_2 --schemeTypes secp256k1 secp256k1
+          node sui/faucet.js --recipient $ADDRESS_1
+          node sui/faucet.js --recipient $ADDRESS_2
+          node sui/faucet.js --recipient $MULTISIG_ADDRESS
+          echo "KEY_1=$KEY_1" >> $GITHUB_ENV
+          echo "KEY_2=$KEY_2" >> $GITHUB_ENV
+          echo "ADDRESS_1=$ADDRESS_1" >> $GITHUB_ENV
+          echo "ADDRESS_2=$ADDRESS_2" >> $GITHUB_ENV
+          echo "MULTISIG_ADDRESS=$MULTISIG_ADDRESS" >> $GITHUB_ENV
+
+      - name: Transfer Upgrade Cap to Multisig Address
+        run: |
+          upgrade_cap=$(cat axelar-chains-config/info/local.json | jq -r '.sui.contracts.AxelarGateway.objects.UpgradeCap')
+          node sui/transfer-object.js --objectId $upgrade_cap --recipient $MULTISIG_ADDRESS
+
+      - name: Generate Unsigned Tx File
+        run: |
+          node sui/deploy-contract.js upgrade AxelarGateway any_upgrade --offline --txFilePath ./tx-upgrade.json --sender $MULTISIG_ADDRESS
+
+      - name: Sign Tx File with Multisig Signer
+        run: |
+          pk_1=$(sui keytool export --key-identity multisig-1 --json | jq .exportedPrivateKey | sed 's/"//g')
+          pk_2=$(sui keytool export --key-identity multisig-2 --json | jq .exportedPrivateKey | sed 's/"//g')
+          node sui/multisig.js --action sign --txBlockPath ./tx-upgrade.json --signatureFilePath signature-1.json --offline --privateKey $pk_1
+          node sui/multisig.js --action sign --txBlockPath ./tx-upgrade.json --signatureFilePath signature-2.json --offline --privateKey $pk_2
+
+      - name: Submit Signed Tx File
+        run: |
+          node sui/multisig.js --txBlockPath ./tx-upgrade.json --signatureFilePath ./combined.json --action combine --signatures signature-1.json signature-2.json
 
       ###### Command: Transfer Object ######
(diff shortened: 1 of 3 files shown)


.github/workflows/test-sui.yaml
```diff
@@ -33,8 +33,8 @@ jobs:
       - name: Spin up Sui Network
         run: nohup sh -c "sui-test-validator" > nohup.out 2> nohup.err < /dev/null &
 
-      - name: Sleep for 30 seconds
-        run: sleep 30s
+      - name: Sleep for 40 seconds
```
milapsheth, line 36:
```suggestion
      - name: Wait for Sui network spinup
```
Is there a certain log that can we can wait for to check if the network is ready instead of just sleeping?

.github/workflows/test-sui.yaml
```diff
@@ -139,7 +139,55 @@ jobs:
...
+
+          # Get multisig address
+          MULTISIG_ADDRESS=$(sui keytool multi-sig-address --pks $KEY_1 $KEY_2 --weights 1 1 --threshold 1 --json | jq -r .multisigAddress)
+
+          # Initialize multisig
+          node sui/multisig.js --action init --threshold 1 --base64PublicKeys $KEY_1 $KEY_2 --schemeTypes secp256k1 secp256k1
+
+          # Faucet operations
+          node sui/faucet.js --recipient $ADDRESS_1
+          node sui/faucet.js --recipient $ADDRESS_2
```
milapsheth, line 162: Isn't the multisig address the one that needs funding and not the participants?

sui/multisig.js
```diff
@@ -153,13 +213,21 @@ if (require.main === module) {
...
     program.addOption(new Option('--offline', 'run in offline mode'));
     program.addOption(new Option('--combinedSignPath <file>', 'combined signature file path'));
     program.addOption(new Option('--signatureFilePath <file>', 'signed signature will be stored'));
 
+    // The following options are only used with the init action
+    program.addOption(
+        new Option('--base64PublicKeys [base64PublicKeys...]', 'An array of public keys to use for init the multisig address'),
+    );
+    program.addOption(new Option('--weights [weights...]', 'weight for multisig'));
+    program.addOption(new Option('--schemeTypes [schemeTypes...]', 'scheme types for each base64 public keys'));
```
milapsheth, line 228: Mention that it defaults to secp256k1 for each

sui/multisig.js
```diff
@@ -153,13 +213,21 @@ if (require.main === module) {
...
     program.addOption(new Option('--signatures [files...]', 'array of signed transaction files'));
     program.addOption(new Option('--offline', 'run in offline mode'));
     program.addOption(new Option('--combinedSignPath <file>', 'combined signature file path'));
     program.addOption(new Option('--signatureFilePath <file>', 'signed signature will be stored'));
 
+    // The following options are only used with the init action
+    program.addOption(
+        new Option('--base64PublicKeys [base64PublicKeys...]', 'An array of public keys to use for init the multisig address'),
+    );
+    program.addOption(new Option('--weights [weights...]', 'weight for multisig'));
```
milapsheth, line 227: Mention the default here. Warning isn't needed then

sui/multisig.js
```diff
@@ -1,9 +1,63 @@
...
+
+    const schemeTypes = options.schemeTypes || Array(base64PublicKeys.length).fill('secp256k1');
+    const weights = options.weights || Array(base64PublicKeys.length).fill(1);
+
+    if (!options.schemeTypes) {
+        printWarn('Scheme types not provided, defaulting to secp256k1');
+    }
+
+    if (!options.weights) {
+        printWarn('Weights not provided, defaulting to 1');
```
milapsheth, line 33: mention default so this is not needed

sui/multisig.js
```diff
@@ -153,13 +213,21 @@ if (require.main === module) {
     addBaseOptions(program);
 
     program.addOption(new Option('--txBlockPath <file>', 'path to unsigned tx block'));
-    program.addOption(new Option('--action <action>', 'action').choices(['sign', 'combine', 'execute']).makeOptionMandatory(true));
+    program.addOption(new Option('--action <action>', 'action').choices(['sign', 'combine', 'execute', 'init']).makeOptionMandatory(true));
```
milapsheth, line 216: @blockchainguyy Refactor this to use Commands like we do for other Sui scripts

.github/workflows/test-sui.yaml
```diff
@@ -139,7 +139,55 @@ jobs:
...
+          node sui/faucet.js --recipient $MULTISIG_ADDRESS
+
+          # Set environment variables
+          echo "KEY_1=$KEY_1" >> $GITHUB_ENV
+          echo "KEY_2=$KEY_2" >> $GITHUB_ENV
+          echo "ADDRESS_1=$ADDRESS_1" >> $GITHUB_ENV
+          echo "ADDRESS_2=$ADDRESS_2" >> $GITHUB_ENV
```
milapsheth, line 169: not needed in the env?

.github/workflows/test-sui.yaml
```diff
@@ -139,7 +139,55 @@ jobs:
...
+      - name: Sign Tx File with Multisig Signer
+        run: |
+          pk_1=$(sui keytool export --key-identity multisig1 --json | jq .exportedPrivateKey | sed 's/"//g')
+          pk_2=$(sui keytool export --key-identity multisig2 --json | jq .exportedPrivateKey | sed 's/"//g')
+          node sui/multisig.js --action sign --txBlockPath ./tx-upgrade.json --signatureFilePath signature-1.json --offline --privateKey $pk_1
+          node sui/multisig.js --action sign --txBlockPath ./tx-upgrade.json --signatureFilePath signature-2.json --offline --privateKey $pk_2
+
+      - name: Submit Signed Tx File
+        run: |
+          node sui/multisig.js --txBlockPath ./tx-upgrade.json --signatureFilePath ./combined.json --action combine --signatures signature-1.json signature-2.json
```
milapsheth, line 190: Test if gateway approve still works post upgrade

milapsheth: Is the contract address being updated to the new package id after upgrade?

.github/workflows/test-sui.yaml
npty: nice catch! thanks

.github/workflows/test-sui.yaml
npty: fixed by using https://github.com/nev7n/wait_for_response/tree/master which simply does http head call periodically to the given endpoint which is the faucet endpoint here.

sui/multisig.js
npty: The default value for `weights` needs to be an array of 1 which has the same length as `base64PublicKeys`, so i don't think we can set the default value here.

.github/workflows/test-sui.yaml
npty: @milapsheth Yes, it'll be upgraded to the new package id. Approval works with either the upgraded contract or the original contract. In this case, should we implement some mechanism to guard access to all prior versions of the package, because they're still exist on-chain?

milapsheth: Foivos is looking into the upgrade mechanism to create a design for our contracts. We want some methods to be accessible on the old contract potentially, and some to not be depending on upgrade versioning

milapsheth (approved):

PR #351, opened 2024-08-28
refactor(amplifier): update to async style

eguajardo:
https://axelarnetwork.atlassian.net/browse/AXE-4714

### DIFF
--- a/cosmwasm/deploy-contract.js
+++ b/cosmwasm/deploy-contract.js
@@ -3,11 +3,16 @@
 require('dotenv').config();
 const { isNil } = require('lodash');
 
+const { instantiate2Address } = require('@cosmjs/cosmwasm-stargate');
+
 const { isNumber, printInfo, loadConfig, saveConfig, prompt, getChainConfig } = require('../common');
 const {
     prepareWallet,
     prepareClient,
+    fromHex,
+    getSalt,
     getChains,
+    updateContractConfig,
     fetchCodeIdFromCodeHash,
     uploadContract,
     instantiateContract,
@@ -18,7 +23,7 @@ const { Command, Option } = require('commander');
 const { addAmplifierOptions } = require('./cli-utils');
 
 const upload = async (client, wallet, chainName, config, options) => {
-    const { reuseCodeId, contractName, fetchCodeId } = options;
+    const { reuseCodeId, contractName, fetchCodeId, instantiate2, salt, chainNames } = options;
     const {
         axelar: {
             contracts: { [contractName]: contractConfig },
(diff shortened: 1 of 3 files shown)


cosmwasm/deploy-contract.js
```diff
@@ -29,31 +29,28 @@ const upload = (client, wallet, chainName, config, options) => {
     if (!fetchCodeId && (!reuseCodeId || isNil(contractConfig.codeId))) {
```
milapsheth, line 29: codeId should be extracted here. And then `contractConfig` should be defined as either the top-level config or chain specific (and initialize the chain specific config to {} if it doesn't exist), so you don't have the `if (chainConfig)` checks everywhere

cosmwasm/submit-proposal.js
```diff
@@ -195,53 +190,58 @@ const main = async (options) => {
...
+        }
+
+        default:
+            throw new Error('Invalid proposal type');
+    }
+
+    saveConfig(config, env);
 };
 
 const programHandler = () => {
```
milapsheth, line 247: not in this PR, but you can refactor the CLI handler to use commands. So options can be defined only for the cmds that need them, similar to axelard CLI
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/sui/gateway.js#L343-L358

milapsheth (approved):

cosmwasm/submit-proposal.js
eguajardo: yeah, I have been wanting to do that, but haven't had the chance to dig deeper into how sub commands work, will take a look at the sui script

cosmwasm/deploy-contract.js
```diff
@@ -29,31 +29,28 @@ const upload = (client, wallet, chainName, config, options) => {
...
-                printInfo('Expected contract address', address);
-            })
-            .then(() => ({ wallet, client }));
-    }
+        const { address, codeId } = await uploadContract(client, wallet, config, options);
+
+        printInfo('Uploaded contract binary');
+        contractConfig.codeId = codeId;
+
+        if (!address) {
```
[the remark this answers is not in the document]
milapsheth: this shouldn't be handled by `uploadContract`. Instead once upload completes, if options.instantiate2 was set, then call a `predictAddress` helper with `options` that can return the predicted address.

cosmwasm/deploy-contract.js
eguajardo: refactoring contractConfig will be done in a different PR after evaluating the best way to also move codeId and other top level contract configs to the contract chain config

milapsheth (approved):

PR #356, opened 2024-08-29
feat: add flow config to stagenet

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-4644

### DIFF
--- a/axelar-chains-config/info/devnet-amplifier.json
+++ b/axelar-chains-config/info/devnet-amplifier.json
@@ -66,6 +66,70 @@
         "api": "https://api.routescan.io/v2/network/testnet/evm/43113/etherscan"
       }
     },
+    "avalanche-fuji": {
+      "name": "Avalanche Fuji",
+      "id": "avalanche-fuji",
+      "axelarId": "avalanche-fuji",
+      "chainId": 43113,
+      "rpc": "https://avalanche-fuji-c-chain-rpc.publicnode.com",
+      "tokenSymbol": "ETH",
+      "confirmations": 1,
+      "contracts": {
+        "AxelarGateway": {
+          "gasOptions": {
+            "gasLimit": 5000000
+          },
+          "deployer": "0xba76c6980428A0b10CFC5d8ccb61949677A61233",
+          "operator": "0xba76c6980428A0b10CFC5d8ccb61949677A61233",
+          "proxyDeploymentArgs": [
+            "0x3c19E701ed674350DE8cAEb0B85A388499246D2d",
+            "0xba76c6980428A0b10CFC5d8ccb61949677A61233",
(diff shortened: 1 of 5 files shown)


axelar-chains-config/info/stagenet.json
```diff
@@ -1691,6 +1691,78 @@
...
+          ],
+          "initialVerifierSetId": "c7483b242156a8149fe6c4efcf62184d9fbf6b1be5530872f80a472d33070cf6",
+          "address": "0x5A4fA5187BddD93097dE4F6062a2E96C82Ea2d61",
+          "implementation": "0xC5578b1c63CE803B9f7f185652CC1E14Af887b8f",
+          "implementationCodehash": "0x3bd2a46c04ccf59a0e454ef828232d66a71c5a9b8004e0ea516257a7ad0aa21c",
+          "deploymentMethod": "create",
+          "previousSignersRetention": 15,
+          "domainSeparator": "0x3dad2b901117c30d4f601c3b9e9ee89cec4371e79e9908122e8e1da1904347ee",
+          "minimumRotationDelay": 300
```
milapsheth, line 1718: For all amplifier chain configs, and in the gateway deploy script, can you add the above field?
```suggestion
          "minimumRotationDelay": 300,
          "chainType": "evm",
          "connectionType": "amplifier"
```
similarly, `"consensus"` for the existing core connections

axelar-chains-config/info/stagenet.json
```diff
@@ -1691,6 +1691,78 @@
...
+          "salt": "Create3Deployer"
+        },
+        "AxelarGasService": {
+          "collector": "0xc5C525B7Bb2a7Ce95C13Ee5aBdB7F8fd3cb77392",
+          "salt": "AxelarGasService",
+          "address": "0xeE1d04E4F9F9AdEB7be7d6996f4A726773460A87",
+          "implementation": "0x1F002cd258e2e818eC5ecCD46DD6cb938c5Cf7f8",
+          "deployer": "0xBeF25f4733b9d451072416360609e5A4c115293E"
+        },
+        "Operators": {
```
milapsheth, line 1745: has axelarons whitelisted their relayers on this contract? do we have a shared list of relayer addresses by network that we should be adding as operators during deployment? Will make the setup easier for axelarons

axelar-chains-config/info/stagenet.json
```diff
@@ -1691,6 +1691,78 @@
...
+          "deploymentMethod": "create2",
+          "codehash": "0xc561dc32ef670c929db9d7fbf6b5f6c074a62a30602481ba3b88912ca6d79feb",
+          "predeployCodehash": "0xc561dc32ef670c929db9d7fbf6b5f6c074a62a30602481ba3b88912ca6d79feb",
+          "salt": "Operators"
+        }
+      },
+      "explorer": {
+        "name": "Flowscan",
+        "url": "https://evm-testnet.flowscan.io/",
+        "api": "https://evm-testnet.flowscan.io/api"
```
milapsheth, line 1758: have we tested this? Add the explorer account for it and update the verification keys.json

axelar-chains-config/info/stagenet.json
```diff
@@ -1691,6 +1691,78 @@
...
+      },
+      "gasOptions": {
+        "gasLimit": 10000000
```
milapsheth, line 1761: by default, we set to 8M unless the chain needs something special

milapsheth (approved):

axelar-chains-config/info/stagenet.json
milapsheth: create another PR for this so you can do it for all configs. Just don't want to forget about it

axelar-chains-config/info/stagenet.json
sdaveas: We've been using our own gas service and operators contract for stagenet. If we should use the ones that are already deployed, we should update our configs.

To get the addresses of refunders per environment there's an axelarons command. Can also be found [here](https://github.com/axelarnetwork/axelarons-tools/blob/01076262318d60b88bba9a940f161a6ee20d7600/test-gmp/config/stagenet.json#L13) (it's the same set of addresses for all chains)

sdaveas: Cool that's even better. We can refresh our configs so that we use the deployed operators contracts. On a second look, I think I've diverged from doing that with Flow - all the other chains use the same operators contract, which means it's the one deployed by chainops. Surely that's the case for testnet/mainnet, we don't deploy anything of our own there.

And yes, operators addresses are always the same across all chains.

milapsheth (approved):

PR #361, opened 2024-09-02
chore: use structs from axelar-cgp-sui

npty:
# Description

Remove defined structs from types-utils.js and import from the `axelar-cgp-sui` package instead.


milapsheth (approved):

.github/workflows/test-sui.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
           SUI_VERSION: ${{ env.SUI_VERSION }}
 
       - name: Spin up Sui Network
-        run: nohup sh -c "sui-test-validator" > nohup.out 2> nohup.err < /dev/null &
+        run: nohup sh -c "RUST_LOG="off,sui_node=info" sui start --with-faucet --force-regenesis" > nohup.out 2> nohup.err < /dev/null &
```
milapsheth, line 40: maybe add a comment describing why sui is being used and with certain args such as `force-regenesis` for future clarity?

npty: i've used `sui start` command because when unzip the tarfile for version `mainnet-1.31.1`, there's no `sui-test-validator` binary, so i've used this command instead. but then I notice that there's `sui-test-validator` binary exist in version `testnet-1.32.0`. So, i assume that the `sui-test-validator` only exists in the testnet version. maybe we should revert back to `sui-test-validator`?

npty: will do. there's deprecation output from `sui-test-validator` command as follows:

```
sui-test-validator binary has been deprecated in favor of sui start, which is a more powerful command that allows you to start the local network with more options.

How to install/build the sui binary IF:
    A: you only need the basic functionality, so just faucet and no persistence (no indexer, no GraphQL service), build from source as usual (cargo build --bin sui) or download latest archive from release archives (starting from testnet v1.28.1 or devnet v1.29) and use sui binary.
    B: you need to also start an indexer (--with-indexer ), or a GraphQL service (--with-graphql), you either:
    - download latest archive from release archives (starting from testnet v1.28.1 or devnet v1.29) and use sui-pg binary.
  OR
    - build from source. This requires passing the indexer feature when building the sui binary, as well as having libpq/postgresql dependencies installed (just as when using sui-test-validator):
        - cargo build --bin sui --features indexer
        - cargo run --features indexer --bin sui -- start --with-faucet --force-regenesis --with-indexer --with-graphql

Running the local network:
 - (Preferred) In the simplest form, you can replace sui-test-validator with sui start --with-faucet --force-regenesis. This will create a network from a new genesis and start a faucet (127.0.0.1:9123). This will not persist state.
 - Use the drop-in replacement script: sui/scripts/sui-test-validator.sh and pass in all the flags/options as you used to.
 
 Use sui start --help to see all the flags and options, such as:
  * --with-indexer --> to start the indexer on the default host and port. Note that this requires a Postgres database to be running locally, or you need to set the different options to connect to a remote indexer database.
  * --with-graphql --> to start the GraphQL server on the default host and port
 ```

PR #362, opened 2024-09-03
chore: check relevant changes in job

npty:
# Description

This PR moves Sui file changes detection inside the job, allowing the workflow to be set as a required check.

### DIFF
--- a/.github/workflows/test-sui.yaml
+++ b/.github/workflows/test-sui.yaml
@@ -1,16 +1,36 @@
 name: Test Sui
 
-on:
-  pull_request:
-    paths:
-      - sui/**
-      - common/**
-      - .github/actions/setup-sui/**
-      - .github/workflows/test-sui.yaml
+on: pull_request
 
 jobs:
+  check-relevant-changes:
+    name: Check for Relevant Changes
+    runs-on: blacksmith-2vcpu-ubuntu-2204
+    outputs:
+      run_tests: ${{ steps.filter.outputs.sui == 'true' || steps.filter.outputs.common == 'true' || steps.filter.outputs.github == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-sui.yaml
```diff
@@ -1,16 +1,36 @@
...
-  pull_request:
-    paths:
-      - sui/**
-      - common/**
-      - .github/actions/setup-sui/**
-      - .github/workflows/test-sui.yaml
+on: pull_request
 
 jobs:
+  check-relevant-changes:
```
milapsheth, line 6: Keep the two as a single job and add this as the first step? Looks a bit cleaner on the checks

milapsheth (approved):

.github/workflows/test-sui.yaml
```diff
@@ -16,7 +10,27 @@ jobs:
...
+              - '.github/workflows/test-sui.yaml'
+
+      - name: Summarize Changes
+        run: |
+          echo "Changes in sui: ${{ steps.filter.outputs.sui }}"
+          echo "Changes in common: ${{ steps.filter.outputs.common }}"
+          echo "Changes in github: ${{ steps.filter.outputs.github }}"
+
+      - name: Setup Node.js
+        if: steps.filter.outputs.sui == 'true' || steps.filter.outputs.common == 'true' || steps.filter.outputs.github == 'true'
```
milapsheth, line 33: will this if only skip this step or all the rest as well?

npty: actually, I concern about this after commit as well 😅. updated in the commit below.

milapsheth (approved):

PR #367, opened 2024-09-05
chore(sui): reuse workflow from axelar-cgp-sui

npty:
# Description

[AXE-4702](https://axelarnetwork.atlassian.net/browse/AXE-4702)

[AXE-4702]: https://axelarnetwork.atlassian.net/browse/AXE-4702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/test-sui.yaml
```diff
@@ -35,25 +35,23 @@ jobs:
...
-      - name: Install
-        run: npm ci
-
-      - name: Read SUI_VERSION from JSON
-        id: read-json
+      - name: Prepare version.json file
         run: |
-          SUI_VERSION=$(jq -r '.SUI_VERSION' sui/version.json)
-          echo "SUI_VERSION=${SUI_VERSION}" >> $GITHUB_ENV
+          curl -sL https://raw.githubusercontent.com/axelarnetwork/axelar-cgp-sui/v0.6.0/version.json -o version.json
```
milapsheth, line 40: get the `version.json` from the installed node_modules/axelar-cgp-sui instead? otherwise we end up having to manage this url instead

.github/workflows/test-sui.yaml
```diff
@@ -35,25 +35,23 @@ jobs:
...
 
       - name: Setup Sui CLI and install dependencies
-        uses: ./.github/actions/setup-sui
-        with:
-          SUI_VERSION: ${{ env.SUI_VERSION }}
+        uses: axelarnetwork/axelar-cgp-sui/.github/actions/install@v0.6.0
```
milapsheth, line 43: Avoid hardcoding the version. Get it from package.json/node_modules

.github/actions/setup-sui/action.yaml
```diff
@@ -1,15 +1,16 @@
 name: Setup Sui CLI
```
milapsheth, line 1: this file will be deleted now?

milapsheth (approved):

.github/workflows/test-sui.yaml
npty: we don't include `version.json` in the published package. we cannot avoid edit this file completely because of the `uses` syntax.

.github/workflows/test-sui.yaml
npty: Dynamic values cannot be used in the `uses` syntax because it is resolved at compile time, not at runtime.

.github/actions/setup-sui/action.yaml
npty: yeah, should be deleted after it works. i keep it as a reference while i'm working on it.

.github/workflows/test-sui.yaml
milapsheth: hmm, use the latest main then so we don't have to keep bumping this. There's a small chance that the latest main breaks but seems unlikely

.github/workflows/test-sui.yaml
milapsheth: let's include the version.json in axelar-cgp-sui release?

.github/workflows/test-sui.yaml
npty: yeah, that would be better

.github/workflows/test-sui.yaml
```diff
@@ -35,12 +35,21 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
```
milapsheth, line 39:
```suggestion
        uses: actions/setup-node@v4
```
use the latest one?

milapsheth (approved):

PR #376, opened 2024-09-16
fix: contract verification for flow

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-4738

### DIFF
--- a/evm/verify-contract.js
+++ b/evm/verify-contract.js
@@ -152,8 +152,10 @@ async function processCommand(config, chain, options) {
         case 'AxelarGateway': {
             if (contractConfig.connectionType === 'amplifier') {
                 verifyAmplifierGateway(chain, contractConfig, env, wallet, verifyOptions, options);
-            } else {
+            } else if (contractConfig.connectionType === 'consensus') {
                 verifyConsensusGateway(config, chain, contractConfig, env, wallet, verifyOptions, options);
+            } else {
+                throw new Error(`Incompatible Gateway connection type`);
             }
 
             break;


evm/utils.js
```diff
@@ -1017,6 +1022,10 @@ async function getWeightedSigners(config, chain, options) {
     return { signers: [signers], verifierSetId };
 }
 
+const verifyContractByName = (env, chain, contract, args, options = {}, name) => {
```
milapsheth, line 1025:
```suggestion
const verifyContractByName = (env, chain, name, contract, args, options = {}) => {
```
options should always be at the end

milapsheth, line 1025: add a docstring about why this method exists

evm/verify-contract.js
```diff
@@ -29,10 +38,12 @@ async function processCommand(config, chain, options) {
...
     printInfo('Verifying contract', contractName);
-    printInfo('Contract address', options.address || chain.contracts[contractName]?.address);
 
-    const contractConfig = chain.contracts[contractName];
-    const contractAddress = options.address || contractConfig.address;
+    const contractAddress =
+        options.address || chain.contracts[contractName === 'AxelarAmplifierGateway' ? 'AxelarGateway' : contractName]?.address;
+    printInfo('Contract address', contractAddress);
+
+    const contractConfig = chain.contracts[contractName === 'AxelarAmplifierGateway' ? 'AxelarGateway' : contractName];
```
milapsheth, line 46: no changes are needed here if you can handle it generically as per below

evm/utils.js
```diff
@@ -1022,7 +1022,8 @@ async function getWeightedSigners(config, chain, options) {
     return { signers: [signers], verifierSetId };
 }
 
-const verifyContractByName = (env, chain, contract, args, options = {}, name) => {
+// To be used if multiple contracts are detected with matching bytecode
```
milapsheth, line 1025:
```suggestion
// Verify contract using it's source code path. The path is retrieved dynamically by the name.
```

evm/verify-contract.js
```diff
@@ -101,24 +150,11 @@ async function processCommand(config, chain, options) {
...
-            const setupParams = defaultAbiCoder.encode(
-                ['address', 'address', 'bytes'],
-                [contractConfig.deployer, contractConfig.deployer, '0x'],
-            );
-
-            await verifyContract(env, chain.name, auth, [authParams], verifyOptions);
-            await verifyContract(env, chain.name, tokenDeployer, [], verifyOptions);
-            await verifyContract(env, chain.name, implementation, [auth, tokenDeployer], verifyOptions);
-            await verifyContract(env, chain.name, contractAddress, [implementation, setupParams], verifyOptions);
+            if (contractConfig.connectionType === 'amplifier') {
```
milapsheth, line 153: If connectionType isn't present, throw an error that connectionType should be set to the expected values

milapsheth (approved):

PR #380, opened 2024-09-18
chore(cosmwasm): update axelarnet instantiation message

jcs47:


cosmwasm/utils.js
```diff
@@ -451,6 +452,7 @@ const makeAxelarnetGatewayInstantiateMsg = (config, chainName) => {
 
     return {
         router_address: routerAddress,
+        nexus_gateway: nexusAddress,
         chain_name: axelarId.toLowerCase(),
```
milapsheth, line 456: we decided to stick with `axelar` as the chain name for axelarnet-gateway. can you use that instead? you can update axelarId in the config. Although, the evm deploy-contract for the Governance contract should use a new `axelarConsensusId` field which is still `Axelarnet`.

milapsheth (approved):

PR #381, opened 2024-09-19
chore(sui): updateMoveToml when upgrade

npty:
# Description

https://axelarnetwork.atlassian.net/browse/AXE-4977


milapsheth (approved):

.github/workflows/test-sui.yaml
```diff
@@ -110,6 +110,9 @@ jobs:
       - name: Deploy Utils
         run: node sui/deploy-contract deploy Utils
 
+      - name: Deploy VersionControl
+        run: node sui/deploy-contract deploy VersionControl
+
```
npty, line 115: @jcs47 `Utils` and `VersionControl` contracts need to be deployed before Gateway contract like above

PR #383, opened 2024-09-19
feat(amplifier): add support for migration through governance proposal

eguajardo:
https://axelarnetwork.atlassian.net/browse/AXE-5157


common/utils.js
```diff
@@ -355,7 +355,7 @@ const getSaltFromKey = (key) => {
     return keccak256(defaultAbiCoder.encode(['string'], [key.toString()]));
 };
 
-const getContractConfig = async (config, chain) => {
+const getContractStorageConfig = async (config, chain) => {
```
milapsheth, line 358: I like the previous name better

cosmwasm/cli-utils.js
```diff
@@ -100,13 +104,17 @@ const addInstantiateProposalOptions = (program) => {
...
-    program.addOption(new Option('--msg <msg>', 'json encoded execute message'));
+    program.addOption(new Option('--msg <msg>', 'json encoded execute message').makeOptionMandatory(true));
 };
 
 const addParamChangeProposalOptions = (program) => {
     program.addOption(new Option('--changes <changes>', 'parameter changes'));
 };
 
+const addMigrateOptions = (program) => {
+    program.addOption(new Option('--msg <msg>', 'json encoded migration message').default('{}'));
```
milapsheth, line 115: Avoid the default, make the user submit it explicitly

cosmwasm/submit-proposal.js
```diff
@@ -135,12 +130,8 @@ const instantiate = async (client, wallet, config, options) => {
...
 
         const { contractName, instantiate2, predictOnly, fetchCodeId } = options;
-        const {
-            axelar: {
-                contracts: { [contractName]: contractConfig },
-            },
-            chains: { [chainName]: chainConfig },
-        } = config;
+        const contractConfig = getAmplifierContractConfig(config, contractName);
+        const chainConfig = getChainConfig(config, chainName);
```
milapsheth, line 134: Can we remove the multiple chain support next?

cosmwasm/utils.js
```diff
@@ -675,6 +686,20 @@ const getParameterChangeParams = ({ title, description, changes }) => ({
...
 
+const getMigrateContractParams = (config, options, chainName) => {
+    const { contractName, msg } = options;
+
+    const contractConfig = getAmplifierContractConfig(config, contractName);
+    const chainConfig = getChainConfig(config, chainName);
+
+    return {
+        ...getSubmitProposalParams(options),
+        contract: contractConfig[chainConfig?.axelarId]?.address || contractConfig.address,
```
milapsheth, line 697: We should clean this up soon, having to constantly handle chain specific vs generic config is pretty confusing

cosmwasm/utils.js
```diff
@@ -737,6 +762,15 @@ const encodeParameterChangeProposal = (options) => {
     };
 };
 
+const encodeMigrateContractProposal = (config, options, chainName) => {
+    const proposal = MigrateContractProposal.fromPartial(getMigrateContractParams(config, options, chainName));
+
+    return {
```
milapsheth, line 768: Can we use the cosmwasm JS SDK to encode this instead of hardcoding?

cosmwasm/utils.js
```diff
@@ -737,6 +762,15 @@ const encodeParameterChangeProposal = (options) => {
...
+        typeUrl: '/cosmwasm.wasm.v1.MigrateContractProposal',
+        value: Uint8Array.from(MigrateContractProposal.encode(proposal).finish()),
+    };
+};
+
 const encodeSubmitProposal = (content, config, options, proposer) => {
```
milapsheth, line 774: We should create a test-cosmwasm gh action similar to test-sui/test-evm that runs cmds against a local node. Can you create a ticket and take that on in the future?

milapsheth (approved):

common/utils.js
eguajardo: I don't know, I think it's confusing because we usually use the `get_Config` to retrieve something from the json file, but this method is getting the Config from the instantiated contract's storage

cosmwasm/cli-utils.js
eguajardo: thought about it, but I sometimes find myself forgetting about the right string which parses correctly to Rust `Empty`, so I thought making things easier by providing the default. The proposal would fail anyway if the message is not formatted correctly, but I don't have strong opinions about it.

cosmwasm/submit-proposal.js
eguajardo: was planning to prioritize it like this:
1. subcommands for non-governance scripts
2. migrate script through admin account
3. remove multiple chain support

let me know if you think I should re-prioritize

cosmwasm/utils.js
eguajardo: I'm already using cosmjs, or what do you mean?  If you mean using the `SigningCosmWasmClient` like we do with non-governance stuff, it doesn't support directly submitting a proposal, so I need to build the message and then broadcast it

common/utils.js
milapsheth: ah, that's fair, can you add a docstring to clarify this? Maybe we can all it `getAmplifierContractOnchainConfig`, a bit verbose but avoids confusion

cosmwasm/cli-utils.js
milapsheth: I'd prefer to be explicit. The user initiating the migration should ensure they know what the migration msg looks like.  You can mention it in the help text here, that if there is no migration msg, use `'{}'`.

cosmwasm/submit-proposal.js
milapsheth: I think removing multiple chain support, and standardizing chain specific vs generic config reading will make the others easier to implement and review

cosmwasm/utils.js
milapsheth: yeah, i meant for encoding the gov Msg below. Weird that it doesn't support it. Are we on the right version?

PR #385, opened 2024-09-23
feat(sui): request tokens from faucet only if recipient balance is below a specific amount

jcs47:


sui/deploy-contract.js
```diff
@@ -272,7 +272,7 @@ async function upgrade(keypair, client, supportedPackage, policy, config, chain,
 async function mainProcessor(args, options, processor) {
     const config = loadConfig(options.env);
     const [keypair, client] = getWallet(config.sui, options);
-    await printWalletInfo(keypair, client, config.sui, options);
+    await printWalletInfo(keypair.toSuiAddress(), client, config.sui, options);
```
milapsheth, line 275: do this within `printWalletInfo`

sui/faucet.js
```diff
@@ -31,6 +34,12 @@ if (require.main === module) {
     program
         .name('faucet')
         .addOption(new Option('--recipient <recipient>', 'recipient to request funds for'))
+        .addOption(
+            new Option(
+                '--minBalance <amount>',
+                'tokens will only be requested from the faucet if recipient balance is below the amount provided',
+            ).default('1'),
```
milapsheth, line 41: does this cause a failure from the faucet?

milapsheth (approved):

sui/faucet.js
jcs47: If used too many times it can get us rate limited. Furthermore, even if not rate limited, I am currently unable to used the faucet because it throws a 500 response. Hence why I opened this PR

sui/deploy-contract.js
milapsheth: ah, check if it has the sui keypair type internally and only then convert it (or check if it's a string and don't convert)

sui/faucet.js
```diff
@@ -9,14 +9,17 @@ async function processCommand(config, chain, options) {
...
 
-    await printWalletInfo(keypair, client, chain, options);
+    await printWalletInfo(recipient, client, chain, options);
 
-    await requestSuiFromFaucetV0({
-        host: getFaucetHost(chain.networkType),
-        recipient,
-    });
+    const balance = Number((await client.getBalance({ owner: recipient })).totalBalance) / 1e9;
+    if (balance < Number(options.minBalance)) {
```
milapsheth, line 15: if balance is >= min, print warning and return instead. otherwise query the faucet

sui/deploy-contract.js
jcs47: done

sui/faucet.js
jcs47: done

sui/faucet.js
```diff
@@ -1,15 +1,22 @@
...
     const recipient = options.recipient || keypair.toSuiAddress();
 
-    await printWalletInfo(keypair, client, chain, options);
+    await printWalletInfo(recipient, client, chain, options);
+
+    const balance = Number((await client.getBalance({ owner: recipient })).totalBalance) / 1e9;
+
+    if (balance >= Number(options.minBalance)) {
+        printWarn('Wallet balance above minimum, skipping faucet request');
+        process.exit(0);
```
milapsheth, line 18: should return from this method and not exit the whole process

sui/utils/sign-utils.js
```diff
@@ -69,10 +69,14 @@ function getWallet(chain, options) {
     return [keypair, client];
 }
 
-async function printWalletInfo(keypair, client, chain, options) {
-    printInfo('Wallet address', keypair.toSuiAddress());
-
-    const coins = await client.getBalance({ owner: keypair.toSuiAddress() });
+async function printWalletInfo(wallet, client, chain, options) {
+    const owner =
```
milapsheth, line 73: walletAddress perhaps instead of owner

sui/utils/sign-utils.js
```diff
@@ -69,10 +69,14 @@ function getWallet(chain, options) {
...
+        wallet instanceof Ed25519Keypair || wallet instanceof Secp256k1Keypair || wallet instanceof Secp256r1Keypair
```
milapsheth, line 74: is there a base KeyPair class?

PR #386, opened 2024-09-24
feat: add option to iterate over chain's type

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-5265

### DIFF
--- a/evm/utils.js
+++ b/evm/utils.js
@@ -648,6 +648,11 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
         chains = chains.filter((chain) => !config.chains[chain].chainType || config.chains[chain].chainType === 'evm');
     } else if (options.chainNames) {
         chains = options.chainNames.split(',');
+        chains.forEach((chain) => {
+            if (config.chains[chain].chainType && config.chains[chain].chainType !== 'evm') {
+                throw new Error(`Cannot run script for a non EVM chain: ${chain}`);
+            }
+        });
     }
 
     if (chains.length === 0) {


common/cli-utils.js
```diff
@@ -30,16 +30,15 @@ const addBaseOptions = (program, options = {}) => {
...
-        );
+        program.addOption(new Option('-n, --chainNames <chainNames>', 'chains to run the script over').env('CHAINS'));
         program.addOption(new Option('--skipChains <skipChains>', 'chains to skip over'));
         program.addOption(
             new Option(
                 '--startFromChain <startFromChain>',
                 'start from a specific chain onwards in the config, useful when a cmd fails for an intermediate chain',
             ),
         );
+        program.addOption(new Option('--chainType <chainType>', 'chain type to run script over'));
```
milapsheth, line 41: this doesn't need to be configurable. each chain specific script should use the appropriate value when loading configs

evm/utils.js
```diff
@@ -636,18 +636,27 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
...
     const config = loadConfig(options.env);
-    let chains = options.chainName ? [options.chainName] : options.chainNames.split(',');
     const chainsToSkip = (options.skipChains || '').split(',').map((str) => str.trim().toLowerCase());
 
+    let chains = [];
+
     if (options.chainNames === 'all') {
         chains = Object.keys(config.chains);
+        chains = chains.map((chain) => chain.trim().toLowerCase());
+    } else if (options.chainType) {
```
milapsheth, line 649: should always filter for `evm` chain type, or if chain type field isn't present in the config (to avoid breaking compatibility for now).

evm/utils.js
```diff
@@ -636,18 +636,27 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
...
     }
 
-    if (!options.chainName && !options.chainNames) {
-        throw new Error('Chain names were not provided');
-    }
-
     printInfo('Environment', options.env);
 
     const config = loadConfig(options.env);
-    let chains = options.chainName ? [options.chainName] : options.chainNames.split(',');
```
milapsheth, line 646: do any scripts still use `chainName`? switch to the generic chainNames option to simplify

evm/utils.js
```diff
@@ -636,18 +636,22 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
...
     const chainsToSkip = (options.skipChains || '').split(',').map((str) => str.trim().toLowerCase());
 
+    let chains = [];
+
     if (options.chainNames === 'all') {
         chains = Object.keys(config.chains);
+        chains = chains.filter((chain) => config.chains[chain].chainType === 'evm');
```
milapsheth, line 648:
```suggestion
        chains = chains.filter((chain) => !config.chains[chain].chainType || config.chains[chain].chainType === 'evm');
```
treat it as evm by default to avoid breaking any local configs being used. we can remove this after a while

evm/utils.js
```diff
@@ -636,18 +636,22 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
...
+    } else if (options.chainNames) {
+        chains = options.chainNames.split(',');
```
milapsheth, line 650: ensure that each user provided chain satisfies the above check as well

milapsheth (approved):

PR #391, opened 2024-09-25
ci: update the actions used in test-sui workflows

apolikamixitos:


.github/workflows/test-sui.yaml
```diff
@@ -36,18 +36,20 @@ jobs:
...
-        uses: actions/setup-node@v4
+        uses: axelarnetwork/axelar-cgp-sui/.github/actions/nodejs-ci@main
         with:
-          node-version: '18'
-          cache: 'npm'
+          nodejs-version: '18' # Hardcoded to ensure consistency.
 
-      - name: Install dependencies
-        run: npm ci
+      - name: Get SUI Version
```
milapsheth, line 43:
```suggestion
      - name: Get Sui Version
```

.github/workflows/test-sui.yaml
```diff
@@ -36,18 +36,20 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: axelarnetwork/axelar-cgp-sui/.github/actions/nodejs-ci@main
```
milapsheth, line 39: weird to use cgp sui's node setup action (which is meant for the repo itself atm). we should use the action from the generic repo once that's setup. So, I'd avoid changing this (adds unnecessary coupling)

.github/workflows/test-sui.yaml
```diff
@@ -36,18 +36,20 @@ jobs:
...
+        run: |
+          SUI_VERSION=$(jq -r '.SUI_VERSION' "node_modules/@axelar-network/axelar-cgp-sui/version.json")
+          echo "SUI_VERSION=$SUI_VERSION" >> $GITHUB_ENV
 
       - name: Setup Sui CLI
-        uses: axelarnetwork/axelar-cgp-sui/.github/actions/install-sui-cli@main
+        uses: axelarnetwork/axelar-cgp-sui/.github/actions/setup-sui@main
         with:
-          sui-version: $(jq -r '.SUI_VERSION' node_modules/@axelar-network/axelar-cgp-sui/version.json)
+          sui-version: ${{ env.SUI_VERSION }}
```
milapsheth, line 52: what's the reason to create a separate step instead of calling jq directly? style? not a strong opinion if you prefer to separate it

milapsheth (approved):

.github/workflows/test-sui.yaml
```diff
@@ -36,18 +36,20 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: axelarnetwork/axelar-cgp-sui/.github/actions/nodejs-ci@main
         with:
-          node-version: '18'
-          cache: 'npm'
+          nodejs-version: '18' # Hardcoded to ensure consistency.
```
milapsheth, line 41:
```suggestion
          node-version: '18' # Hardcoded to ensure consistency.
```

PR #412, opened 2024-10-21
chore(cosmwasm): remove nexus gateway support

jcs47:


eguajardo (approved):

eguajardo:
Should we remove the nexus gateway from the json files or should we leave it for reference?

milapsheth (approved):

PR #416, opened 2024-10-24
chore: update config.sui to config.chains.sui

npty:


sui/deploy-contract.js
```diff
@@ -127,7 +127,7 @@ async function postDeployExample(published, keypair, client, config, chain, opti
     const [gmpSingletonObjectId] = getObjectIdsByObjectTypes(published.publishTxn, [`${published.packageId}::gmp::Singleton`]);
 
     // ITS Example Params
-    const itsObjectId = config.sui.contracts.ITS?.objects?.ITS;
+    const itsObjectId = config.chains.sui.contracts.ITS?.objects?.ITS;
```
milapsheth, line 130: use `getChainConfig` instead of hardcoding the name. See what we do for stellar
https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/stellar/deploy-contract.js#L101

milapsheth: chain config should be retrieved in the mainProcessor and provided to this function so you only need to use the helper once

npty:
> Can you base your PR on this one: https://github.com/axelarnetwork/axelar-contract-deployments/pull/418/files
> 
> Want to make sure it is compatible with the devnet deployment.
> 
> This is how we solve naming issue for Stellar, just use the same util, should be easy to switch: https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/stellar/deploy-contract.js#L101
> 
> Also note `chainType` field and update usage so we don't run into issues in evm scripts/sui scripts if the value is incorrect. Lastly, we will need to add axelarId into sui otherwise e2e tests fail, please create a PR there if required, and update them to use this new config.

can you clarify the `chainType` field and the desired usage? what do you mean by "adding axelard" into sui? could you pls create tickets? i don't think that's related to this pr.

sui/deploy-contract.js
```diff
@@ -332,9 +332,10 @@ async function upgrade(keypair, client, supportedPackage, policy, config, chain,
 
 async function mainProcessor(args, options, processor) {
     const config = loadConfig(options.env);
-    const [keypair, client] = getWallet(config.sui, options);
-    await printWalletInfo(keypair, client, config.sui, options);
-    await processor(keypair, client, ...args, config, config.sui, options);
+    const sui = getChainConfig(config, 'sui');
```
milapsheth, line 335: define a chainName base option that defaults to `sui` but can be overriden similar to stellar. We'll deploy `sui-test` name on devnet initially until the final release is ready

milapsheth, line 335:
```suggestion
    const chain = getChainConfig(config, options.chainName);
```

npty:
> > can you clarify the chainType field (should i print it for every commands?) and the desired usage?
> 
> `chainType` should be specified for every chain within the chains object, evm chains specify `evm`, sui will specify `sui`. Sui scripts should check if the CHAINS provided in .env leads to a chain config which has `sui` as its `chainType`. A similar check should be enforced on `evm` chains, so should be built like a common util.
> 
> You will also have to handle cases like this:
> 
> https://github.com/axelarnetwork/axelar-contract-deployments/blob/222f36cfbe57ee2ff88154f72f6a72bbb2ba30f9/evm/utils.js#L646
> 
> Where `evm` script will now try to execute on sui since it has been moved to the chains object. You can use `chainType` to filter out non evm chains in such cases. Feel free to create new tickets and assign me as. a reporter if required.

i've added `chainType` in [this commit](https://github.com/axelarnetwork/axelar-contract-deployments/pull/416/commits/f7151bbdc0305769f2b348e0a2a4064cba1ecdf1). Not sure if there's any other places to add `chainType`. Could you pls check? 

I'll create a new ticket to support `chainNames` later.

npty:
> > what do you mean by "adding axelard" into sui?
> > it was `axelarId` not `axelard`.
> 
> https://github.com/axelarnetwork/axelar-e2e/blob/db8d3f368e74ed9b52cf0b48fa69d76be3fb5d50/test/sui.js#L37
> 
> You can see the e2e tests read axelardId from `sui` config. I was talking about, updating the e2e tests so they can run with the new config, ask help from Joao or me if required.

I think axelarId already existed, no? i'm not sure where `axelar-e2e` reads the chain config file from.

npty:
> @npty let's just make sure e2e tests for sui work after the config update. I can help you with how they work with devnet setup.
So this is a blocker to merge?

I'm currently not able to verify if it works for e2e tests right now since e2e tests don't work for me at the first place. Could you verify it?

PR #417, opened 2024-10-24
fix: check destination chain for gmp

npty:
# Description

https://axelarnetwork.atlassian.net/browse/AXE-6143

This PR checks if the destination chain exists in the config in the `send-call` command.


sui/gmp.js
```diff
@@ -18,8 +18,14 @@ const {
     utils: { arrayify },
 } = ethers;
 
-async function sendCommand(keypair, client, chain, args, options) {
+async function sendCommand(keypair, client, chains, args, options) {
     const [destinationChain, destinationAddress, feeAmount, payload] = args;
+
+    if (options.env !== 'local' && !chains[destinationChain]) {
+        throw new Error(`Chain ${destinationChain} not found in the config`);
```
milapsheth, line 25: we shouldn't fail. it might be a partner chain that's not listed in our config, just print a warning that it's not found in the local config so it might not be a registered chain

PR #427, opened 2024-11-07
feat(stellar): script for requesting testnet funds

jcs47:
https://axelarnetwork.atlassian.net/browse/AXE-6456

### DIFF
--- a/stellar/faucet.js
+++ b/stellar/faucet.js
@@ -1,6 +1,6 @@
 const { Command, Option } = require('commander');
 const { ASSET_TYPE_NATIVE, getWallet, addBaseOptions, getBalances } = require('./utils');
-const { saveConfig, loadConfig, printInfo, printWarn, getChainConfig } = require('../common');
+const { loadConfig, printInfo, printWarn, getChainConfig } = require('../common');
 const { Horizon } = require('@stellar/stellar-sdk');
 
 require('./cli-utils');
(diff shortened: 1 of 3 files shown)


stellar/faucet.js
```diff
@@ -0,0 +1,53 @@
...
+
+if (require.main === module) {
+    const program = new Command();
+
+    program
+        .name('faucet')
+        .addOption(new Option('--recipient <recipient>', 'recipient to request funds for'))
+        .addOption(
+            new Option(
+                '--minBalance <amount>',
```
ahramy, line 41: We want to move towards a consistent naming scheme. The CLI flags/naming should be kebab case than camel case. Please refer `gateway.js` or `deploy-contract-js` in Stellar Folder.

```suggestion
                '--min-balance <amount>',
```

stellar/faucet.js
```diff
@@ -0,0 +1,53 @@
+const { Command, Option } = require('commander');
```
ahramy, line 1: Could you also update `README.md` file regarding to how to use stellar faucet?

stellar/faucet.js
```diff
@@ -0,0 +1,55 @@
...
+    await horizonServer.friendbot(recipient).call();
+
+    printInfo('Funds requested', recipient);
+}
+
+async function mainProcessor(options, processor) {
+    const config = loadConfig(options.env);
+    const chain = getChainConfig(config, options.chainName);
+    await processor(chain, options);
+    saveConfig(config, options.env);
```
ahramy, line 32: Does this script saves balances and sequence to config? if not, we should remove `saveConfig`.

jcs47: good point, removed this

PR #442, opened 2024-11-21
feat: add support to generate unsigned txs in sui scripts

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-6405

### DIFF
--- a/sui/its.js
+++ b/sui/its.js
@@ -1,4 +1,4 @@
-const { Command, Option } = require('commander');
+const { Command } = require('commander');
 const { TxBuilder } = require('@axelar-network/axelar-cgp-sui');
 const { loadConfig, saveConfig, getChainConfig } = require('../common/utils');
 const { addBaseOptions, addOptionsToCommands, getWallet, printWalletInfo, broadcastFromTxBuilder, saveGeneratedTx } = require('./utils');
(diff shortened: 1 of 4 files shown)


sui/its.js
```diff
@@ -56,6 +63,9 @@ if (require.main === module) {
         .name('setup-trusted-address')
         .description('Setup trusted address')
         .command('setup-trusted-address <trusted-chain> <trusted-address>')
+        .addOption(new Option('--offline', 'store tx block for offline signing'))
+        .addOption(new Option('--sender <sender>', 'transaction sender'))
+        .addOption(new Option('--txFilePath <file>', 'unsigned transaction will be stored'))
```
npty, line 68: Perhaps create a utility function to add these options, similar to `addBaseOptions` ?

sui/utils/utils.js
```diff
@@ -270,6 +270,17 @@ const checkTrustedAddresses = (trustedAddresses, destinationChain) => {
...
 };
 
+const saveGeneratedTx = async (tx, message, client, options) => {
+    const { txFilePath } = options;
+    validateParameters({ isNonEmptyString: { txFilePath } });
+
+    const txBytes = await tx.build({ client });
+    const txB64Bytes = toB64(txBytes);
+
+    writeJSON({ message, status: 'PENDING', unsignedTx: txB64Bytes }, txFilePath);
```
npty, line 280: do we need `status: PENDING` here or we could simply delete the whole file after broadcasting?

npty (commented):
How do you determine which scripts have to support offline signings? I notice that there're some scripts that haven't been modified e.g. `faucet`, `gas-service`, `multisig`, etc.

PR #463, opened 2024-12-04
feat: add ledger support for Sui

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-6403

### DIFF
--- a/axelar-chains-config/info/devnet-amplifier.json
+++ b/axelar-chains-config/info/devnet-amplifier.json
@@ -630,11 +630,14 @@
       "horizonRpc": "https://horizon-testnet.stellar.org",
       "networkType": "testnet",
       "chainType": "stellar",
+      "decimals": 7,
+      "finality": "1",
+      "approxFinalityWaitTime": 1,
+      "tokenSymbol": "XLM",
       "explorer": {
         "name": "Soroban Nownodes",
         "url": "https://soroban.nownodes.io/testnet"
       },
-      "tokenSymbol": "XLM",
       "contracts": {
         "axelar_gateway": {
           "address": "CBECMRORSIPG4XG4CNZILCH233OXYMLCY4GL3GIO4SURSHTKHDAPEOVM",
--- a/axelar-chains-config/info/mainnet.json
+++ b/axelar-chains-config/info/mainnet.json
@@ -2501,11 +2501,18 @@
     "rpc": "https://rpc.ankr.com/stellar_soroban",
     "horizonRpc": "https://horizon.stellar.org",
     "networkType": "mainnet",
+    "chainType": "stellar",
+    "decimals": 7,
+    "finality": "1",
+    "tokenSymbol": "XLM",
+    "contracts": {
+      "AxelarGateway": {}
+    },
+    "approxFinalityWaitTime": 1,
     "explorer": {
       "name": "Soroban Nownodes",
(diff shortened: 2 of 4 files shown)


RiceAndMeet (approved):

sui/utils/sign-utils.js
```diff
@@ -75,10 +81,18 @@ function getSuiClient(chain, rpc) {
...
-        wallet instanceof Ed25519Keypair || wallet instanceof Secp256k1Keypair || wallet instanceof Secp256r1Keypair
-            ? wallet.toSuiAddress()
-            : wallet;
+    let owner;
+
+    if (options.privateKey !== 'ledger') {
+        owner =
+            wallet instanceof Ed25519Keypair || wallet instanceof Secp256k1Keypair || wallet instanceof Secp256r1Keypair
+                ? wallet.toSuiAddress()
+                : wallet;
```
milapsheth, line 90: When is it not an instance of a KeyPair? Supporting the latter case seems weird, even though this was being done before

sui/utils/sign-utils.js
```diff
@@ -40,6 +41,13 @@ function getWallet(chain, options) {
         }
     }
 
+    const client = getSuiClient(chain, options.rpc);
+
+    if (options.privateKey === 'ledger') {
```
milapsheth, line 46: Make use of the `privateKeyType` instead, that makes more sense instead of reusing the `privateKey`. We should change EVM scripts to also have include a privateKeyType (since we could have hex, mnemonic, or ledger types) as a better design

sui/utils/sign-utils.js
```diff
@@ -157,15 +171,10 @@ async function broadcastSignature(client, txBytes, signature, actionName) {
 
 async function signTransactionBlockBytes(keypair, client, txBytes, options) {
     const serializedSignature = (await keypair.signTransaction(txBytes)).signature;
-    let publicKey;
 
-    try {
-        publicKey = await verifyTransactionSignature(txBytes, serializedSignature);
-    } catch {
-        throw new Error(`Cannot verify tx signature`);
```
milapsheth, line 165: why remove this try/catch? Not needed?

sui/utils/LedgerSigner.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const { toSerializedSignature, messageWithIntent } = require('@mysten/sui/cryptography');
+const { Signer } = require('@mysten/sui/cryptography');
+const { Ed25519PublicKey } = require('@mysten/sui/keypairs/ed25519');
+const TransportNodeHid = require('@ledgerhq/hw-transport-node-hid').default;
+const Sui = require('@mysten/ledgerjs-hw-app-sui').default;
+
+class LedgerSigner extends Signer {
```
milapsheth, line 9: :+1:

sui/utils/LedgerSigner.js
```diff
@@ -0,0 +1,80 @@
...
+
+    async toSuiAddress() {
+        if (!this.sui) {
+            await this.init();
+        }
+
+        return `0x${(await this.sui.getPublicKey(this.path)).address.toString('hex')}`;
+    }
+
+    async signTransaction(bytes) {
```
milapsheth, line 55: Have you tested this out with a ledger?

PR #470, opened 2024-12-10
feat(stellar): add interchain token wasm hash for interchain token service

ahramy:
[AXE-6870](https://axelarnetwork.atlassian.net/browse/AXE-6870)

[AXE-6870]: https://axelarnetwork.atlassian.net/browse/AXE-6870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stellar/deploy-contract.js
+++ b/stellar/deploy-contract.js
@@ -76,6 +76,7 @@ async function getInitializeArgs(config, chain, contractName, wallet, options) {
             if (!chain?.contracts?.interchain_token?.wasmHash) {
                 throw new Error(`interchain_token contract's wasm hash does not exist.`);
             }
+
             const interchainTokenWasmHash = nativeToScVal(Buffer.from(chain?.contracts?.interchain_token?.wasmHash, 'hex'), {
                 type: 'bytes',
             });
@@ -107,7 +108,7 @@ async function deploy(options, config, chain, contractName) {
             deployer: wallet.publicKey(),
             wasmHash: serializeValue(wasmHash),
         };
-        printInfo(contractName + ' Wasm Hash', serializeValue(wasmHash));
+        printInfo(contractName, JSON.stringify(chain.contracts[contractName], null, 2));
         return;
     }
 
@@ -127,6 +128,7 @@ async function deploy(options, config, chain, contractName) {
     chain.contracts[contractName] = {
         address: contractAddress,
         deployer: wallet.publicKey(),
+        wasmHash: serializeValue(wasmHash),
         initializeArgs: serializedArgs,
     };
 }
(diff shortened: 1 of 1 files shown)


stellar/deploy-contract.js
```diff
@@ -91,6 +102,15 @@ async function deploy(options, config, chain, contractName) {
     );
     const wasmHash = await uploadWasm(wasmPath, wallet, chain);
 
+    if (contractName === 'interchain_token') {
+        chain.contracts[contractName] = {
+            deployer: wallet.publicKey(),
+            wasmHash: serializeValue(wasmHash),
```
milapsheth, line 108: store deployer and wasmHash for all contracts, doesn't hurt to have it

stellar/deploy-contract.js
```diff
@@ -91,6 +102,15 @@ async function deploy(options, config, chain, contractName) {
...
+        };
+        printInfo(contractName + ' Wasm Hash', serializeValue(wasmHash));
```
milapsheth, line 110: the full contract config can be printed as JSON at the end of deployment

stellar/deploy-contract.js
ahramy: agreed. updated

stellar/deploy-contract.js
ahramy: updated. it will look like:
```
interchain_token: {
  "deployer": "GDEXMXS37JAZ54R2OEX56FV7OLYAOPIL7JH5QK6JEAOI3D2QALZT2KEI",
  "wasmHash": "e6598bf68a24110a075314d7564f9c6a677a48863f0c235e0d780686b7ca1cb8"
}
```

stellar/deploy-contract.js
```diff
@@ -107,7 +108,7 @@ async function deploy(options, config, chain, contractName) {
             deployer: wallet.publicKey(),
             wasmHash: serializeValue(wasmHash),
         };
-        printInfo(contractName + ' Wasm Hash', serializeValue(wasmHash));
+        printInfo(contractName, JSON.stringify(chain.contracts[contractName], null, 2));
         return;
     }
```
milapsheth, line 113: remove the if block, this should be done for any deployment, and not just interchain token

milapsheth (approved):

stellar/deploy-contract.js
ahramy: We should include an `if ... return` block for the interchain_token contract because we're only uploading the wasm (installing) rather than deploying the contract and initializing it with constructor args. So the if block ie necessary here.

PR #472, opened 2024-12-12
fix(cosmwasm): remove requirement to use chainName option for AxelarnetGateway instantiation

eguajardo:
This fixes the issue of saving the address under `AxelarnetGateway.axelar.address` instead of just `AxelarnetGateway.address`

### DIFF
--- a/cosmwasm/utils.js
+++ b/cosmwasm/utils.js
@@ -22,7 +22,6 @@ const {
     printInfo,
     isString,
     isStringArray,
-    isStringLowercase,
     isKeccak256Hash,
     isNumber,
     toBigNumberString,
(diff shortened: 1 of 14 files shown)


talalashraf (approved):

cosmwasm/utils.js
```diff
@@ -36,6 +36,32 @@ const { normalizeBech32 } = require('@cosmjs/encoding');
 const DEFAULT_MAX_UINT_BITS_EVM = 256;
 const DEFAULT_MAX_DECIMALS_WHEN_TRUNCATING_EVM = 255;
 
+const CONTRACT_NAME_COORDINATOR = 'Coordinator';
```
milapsheth, line 39: Instead of all these globals, define an object that maps the contract name to an object containing the corresponding instantiate fn and the scope type. Standardize the arguments to all instantiate fn so they can be dispatched generically. We do something analogous for sui/deploy-contract.js for contract specific code

cosmwasm/utils.js
```diff
@@ -503,79 +529,43 @@ const makeInstantiateMsg = (contractName, chainName, config) => {
     const { [contractName]: contractConfig } = contracts;
 
     switch (contractName) {
-        case 'Coordinator': {
-            if (chainConfig) {
-                throw new Error('Coordinator does not support chainName option');
-            }
-
+        case CONTRACT_NAME_COORDINATOR: {
             return makeCoordinatorInstantiateMsg(contractConfig, contracts);
```
milapsheth, line 533: Pass in options to the fn, it can extract the chainConfig internally if it needs to (you can use getChainConfig and make it fail if it's not found), but by default `chainName` is optional

jcs47 (approved):

jcs47 (approved):
still LGTM

PR #487, opened 2024-12-24
feat: add stellar upgrader contract to the deploy-contract.js script

cgorenflo:
AXE-6746


stellar/deploy-contract.js
```diff
@@ -139,30 +143,71 @@ async function uploadWasm(filePath, wallet, chain) {
...
     });
 
+    const contractAddress = Address.fromString(uncheckedContractAddress);
+
     const newWasmHash = await uploadWasm(wasmPath, wallet, chain);
     printInfo('New Wasm hash', serializeValue(newWasmHash));
 
     const operation = Operation.invokeContractFunction({
-        contract: contractAddress,
+        contract: chain.contracts.upgrader.address,
```
milapsheth, line 163: Include this in the validation above as well in case it's not been deployed/present in the config

stellar/deploy-contract.js
```diff
@@ -139,30 +143,71 @@ async function uploadWasm(filePath, wallet, chain) {
...
         function: 'upgrade',
-        args: [nativeToScVal(newWasmHash)],
+        args: [
+            nativeToScVal(contractAddress),
+            nativeToScVal(options.newVersion),
+            nativeToScVal(newWasmHash),
+            nativeToScVal([options.migrationData]),
+        ],
```
milapsheth, line 170:
```suggestion
        ].map(nativeToScVal),
```
Can use a map instead

stellar/deploy-contract.js
```diff
@@ -139,30 +143,71 @@ async function uploadWasm(filePath, wallet, chain) {
...
+        auth: await createUpgradeAuths(contractAddress, newWasmHash, options.migrationData, chain, wallet),
     });
+
     await broadcast(operation, wallet, chain, 'Upgraded contract', options);
     chain.contracts[contractName].wasmHash = serializeValue(newWasmHash);
     printInfo('Contract upgraded successfully!', contractAddress);
 }
 
+async function createUpgradeAuths(contractAddress, newWasmHash, migrationData, chain, wallet) {
+    // 20 seems a reasonable number of ledgers to allow for the upgrade to take effect
```
milapsheth, line 180: Create a global constant for it. Could also create an option with a default of 20 though I don't see us configuring it

stellar/deploy-contract.js
```diff
@@ -139,30 +143,71 @@ async function uploadWasm(filePath, wallet, chain) {
...
+                    subInvocations: [],
+                }),
+                wallet.publicKey(),
+                getNetworkPassphrase(chain.networkType),
+            ),
+        ),
+    );
+}
+
+const createAuthorizedFunc = (contractAddress, functionName, args) =>
```
milapsheth, line 202: Move this to the stellar utils.js

milapsheth (approved):

stellar/deploy-contract.js
```diff
@@ -200,6 +245,8 @@ function main() {
         .description('Upgrade a Stellar contract')
         .argument('<contract-name>', 'contract name to deploy')
         .addOption(new Option('--wasm-path <wasmPath>', 'path to the WASM file'))
+        .addOption(new Option('--new-version <newVersion>', 'new version of the contract'))
+        .addOption(new Option('--migration-data <migrationData>', 'migration data'))
```
milapsheth, line 249: Add an upgrade usage example in the stellar/README

stellar/deploy-contract.js
cgorenflo: you mean for the 20? What would be the benefit of that? It's literally just used here once, and, as you point out, that doesn't seem like something we would ever parameterize

stellar/README.md
```diff
@@ -100,6 +99,22 @@ node stellar/deploy-contract.js deploy interchain_token_service --chain-name <CH
...
+node stellar/deploy-contract.js deploy upgrader --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-soroban/target/wasm32-unknown-unknown/release/upgrader.optimized.wasm
+```
+
+After the `upgrader` is deployed, any other instantiated contract can be upgraded by calling the `upgrade` function
+
+```bash
+node stellar/deploy-contract.js upgrade <CONTRACT_NAME> --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-soroban/target/wasm32-unknown-unknown/release/<CONTRACT_NAME>.optimized.wasm --new-version <NEW_VERSION> --migration-data <MIGRATION_DATA>
+```
+
+where `<CONTRACT_NAME>` is the name of the contract to be upgraded and `--wasm-path` points to the upgraded bytecode. As a sanity check, `<NEW_VERSION>` must match the version number defined by the provided bytecode, so upgrading to the wrong version can be prevented. `<MIGRATION_DATA>` is the json encoded data that will be passed to the contract's `migrate` function.
```
milapsheth, line 115: Can you mention how MIGRATION_DATA is encoded for the no migration case and a more complex migration case?

stellar/deploy-contract.js
milapsheth: yeah, parameterization seems overkill. Would still be nice to define a constant variable for it though for style

milapsheth (approved):

PR #491, opened 2025-01-09
chore: stellar test 2025 q1

RiceAndMeet:

PR #499, opened 2025-01-14
refactor: add config params to stellar network for finality

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-7116

### DIFF
--- a/axelar-chains-config/info/devnet-amplifier.json
+++ b/axelar-chains-config/info/devnet-amplifier.json
@@ -633,11 +633,11 @@
       "decimals": 7,
       "finality": "1",
       "approxFinalityWaitTime": 1,
+      "tokenSymbol": "XLM",
       "explorer": {
         "name": "Soroban Nownodes",
         "url": "https://soroban.nownodes.io/testnet"
       },
-      "tokenSymbol": "XLM",
       "contracts": {
         "axelar_gateway": {
           "address": "CBGNPCKNCKKKCI5AUOS2K6H4JDLGPFHHXNDNOEZBIA3VO2NBEZQXFAG5",
--- a/axelar-chains-config/info/devnet-amplifier.json
+++ b/axelar-chains-config/info/devnet-amplifier.json
@@ -630,6 +630,9 @@
       "horizonRpc": "https://horizon-testnet.stellar.org",
       "networkType": "testnet",
       "chainType": "stellar",
+      "decimals": 7,
+      "finality": "1",
+      "approxFinalityWaitTime": 1,
       "explorer": {
         "name": "Soroban Nownodes",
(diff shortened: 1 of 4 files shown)


ahramy:
@blockchainguyy `stellar-test2` is deprecated on devnet-amplifier and the new connection is now `stellar-test-2025-q1`. Please reference this PR: https://github.com/axelarnetwork/axelar-contract-deployments/pull/491/files

axelar-chains-config/info/devnet-amplifier.json
```diff
@@ -630,6 +630,9 @@
       "horizonRpc": "https://horizon-testnet.stellar.org",
       "networkType": "testnet",
       "chainType": "stellar",
+      "decimals": 7,
+      "finality": "1",
```
milapsheth, line 634: @blockchainguyy Can you share your research on the finality mechanism? Since it's a critical piece, we should validate

axelar-chains-config/info/stagenet.json
```diff
@@ -1880,6 +1880,25 @@
...
+      "finality": "1",
+      "approxFinalityWaitTime": 1,
+      "contracts": {
+        "AxelarGateway": {}
+      },
+      "explorer": {
+        "name": "Stellar Expert",
+        "url": "https://stellar.expert/explorer/testnet"
+      },
+      "tokenSymbol": "XLM"
```
milapsheth, line 1901: perhaps move these before `contracts` key, so they're easier to spot

axelar-chains-config/info/mainnet.json
```diff
@@ -2501,6 +2501,13 @@
     "rpc": "https://rpc.ankr.com/stellar_soroban",
     "horizonRpc": "https://horizon.stellar.org",
     "networkType": "mainnet",
+    "chainType": "stellar",
+    "decimals": 7,
+    "finality": "1",
+    "contracts": {
+      "AxelarGateway": {}
```
milapsheth, line 2508: I don't like adding an empty contract field. Makes it more confusing when testing if a contract config is present

axelar-chains-config/info/devnet-amplifier.json
milapsheth: @blockchainguyy Can you confirm whether "transactions are finalized in about 5 secs" means that as soon as they're included in a ledger, it's final, or is there some delay after a ledger is created (queryable in the RPC)?

axelar-chains-config/info/mainnet.json
milapsheth: Let's make the config tests more sensible? Empty contract config is pointless, we should allow no contracts to be specified or if it's specified it should be done properly

milapsheth (approved):

PR #500, opened 2025-01-15
feat(stellar): update stellar its deployment script

AttissNgo:
Adds `operator` to ITS constructor args


nbayindirli (approved):
Was there an associated ticket? If not perhaps add motivation to desc

milapsheth (approved):

PR #518, opened 2025-01-27
feat: add token manager deployment script

ahramy:
[AXE-7267](https://axelarnetwork.atlassian.net/browse/AXE-7267)

[AXE-7267]: https://axelarnetwork.atlassian.net/browse/AXE-7267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stellar/deploy-contract.js
+++ b/stellar/deploy-contract.js
@@ -62,6 +62,10 @@ async function getInitializeArgs(config, chain, contractName, wallet, options) {
                 type: 'bytes',
             });
 
+            const tokenManagerWasmHash = nativeToScVal(Buffer.from(chain?.contracts?.token_manager?.wasmHash, 'hex'), {
+                type: 'bytes',
+            });
+
             return {
                 owner,
                 operator,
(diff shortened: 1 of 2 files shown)


nbayindirli (approved):

stellar/deploy-contract.js
```diff
@@ -84,6 +89,10 @@ async function getInitializeArgs(config, chain, contractName, wallet, options) {
             return { owner, operator };
         }
 
+        case 'token_manager': {
+            return { owner };
```
milapsheth, line 93: similar to interchain token, token manager is not initialized directly, but deployed via ITS

milapsheth (approved):

stellar/deploy-contract.js
ahramy: make sense. it should not be initialized here. updated code.

PR #519, opened 2025-01-28
feat(sui)!: add script option to pause sui contracts

Foivos:


sui/gateway.js
```diff
@@ -510,6 +509,65 @@ async function testNewField(value, options) {
...
+        },
+    });
+    let allowedFunctionsArray = response.data.content.fields.value.fields.version_control.fields.allowed_functions;
+    allowedFunctionsArray = allowedFunctionsArray.map((allowedFunctions) => allowedFunctions.fields.contents);
+
+    const versionsArg = [];
+    const allowedFunctionsArg = [];
+
+    for (const version in allowedFunctionsArray) {
+        const allowedFunctions = allowedFunctionsArray[version];
```
milapsheth, line 527:
```suggestion
        const allowedFunctions = allowedFunctionsArray[version].filter((function_name) => function_name !== 'allow_function');
```
use filter instead?

sui/gateway.js
```diff
@@ -612,6 +670,20 @@ if (require.main === module) {
...
+            mainProcessor(pause, [], options);
+        });
+
+    program
+        .command('unpause')
+        .description('Unpause the gateway')
+        .action((options) => {
+            mainProcessor(unpause, [], options);
+        });
+
```
milapsheth, line 686: update `test-sui.yaml` to run pause/unpause for gateway and ITS. Want to ensure that this action always work, instead of finding issues during an emergency

sui/gateway.js
```diff
@@ -510,6 +509,65 @@ async function testNewField(value, options) {
...
+        allowedFunctionsArg.push(allowedFunctions.join());
+    }
+
+    // Write the
+    writeJSON(
+        {
+            versions: versionsArg,
+            disallowedFunctions: allowedFunctionsArg,
+        },
+        `${__dirname}/../axelar-chains-config/info/sui-gateway-allowed-functions-${options.env}.json`,
```
milapsheth, line 551: save to the existing `sui.contracts.AxelarGateway` config which is already included in git

milapsheth: Instead of dynamically querying this, we should look into creating a json file in the cgp sui that lists the version controlled methods, and including it in the npm release. We already do something like this for the golden test, so shouldn't be hard to add a JSON file containing contract name -> allowed function list mapping

sui/its.js
```diff
@@ -124,8 +235,37 @@ if (require.main === module) {
...
         });
 
+    const allowFunctionsProgram = new Command()
+        .name('allow-functions')
+        .description('Allow functions')
+        .command('allow-functions <versions> <functions>')
+        .action((versions, functions, options) => {
+            mainProcessor(allowFunctions, options, [versions, functions], processCommand);
+        });
+
```
milapsheth, line 245: create a separate `contract.js` script that allows generically allowing/disallowing function by version for any of our contract. It can do also support `pause --all`, `unpause --all`. And for the curated `pause` and `unpause` we can create an object mapping for contract name -> curated function list

sui/pause.js
milapsheth, line 1: Call this `contract.js` since it's meant for generic utils for all our contracts

sui/pause.js
```diff
@@ -0,0 +1,243 @@
...
+    if (packageName === 'AxelarGateway') {
+        const contractConfig = chain.contracts.AxelarGateway;
+        return {
+            packageId: contractConfig.address,
+            singletonId: contractConfig.objects.Gateway,
+            versionedId: contractConfig.objects.Gatewayv0,
+            ownerCapId: contractConfig.objects.OwnerCap,
+            moduleName: 'gateway',
+            defaultFunctions: {
+                versions: [0, 0],
```
milapsheth, line 33: the current package version should be read from the contract config

milapsheth: https://github.com/axelarnetwork/axelar-contract-deployments/blob/main/axelar-chains-config/info/testnet.json#L2354

sui/gateway.js
```diff
@@ -510,6 +508,65 @@ async function testNewField(value, options) {
...
 
+async function pause(keypair, client, config, chain, contracts, args, options) {
+    const response = await client.getObject({
+        id: contracts.objects.Gatewayv0,
+        options: {
+            showContent: true,
+            showBcs: true,
+        },
+    });
+    let allowedFunctionsArray = response.data.content.fields.value.fields.version_control.fields.allowed_functions;
```
milapsheth, line 519: can be removed now

.github/workflows/test-sui.yaml
```diff
@@ -191,6 +197,11 @@ jobs:
           node sui/gmp.js execute ethereum 0x32034b47cb29d162d9d803cc405356f4ac0ec07fe847ace431385fe8acf3e6e5-3 0x4F4495243837681061C4743b74B3eEdf548D56A5 0x1234
 
       ###### Command: ITS Example ######
+      - name: ITS Pause
+        run: node sui/contract.js pause InterchainTokenService -y
+
+      - name: ITS Unpause
+        run: node sui/contract.js unpause InterchainTokenService -y
 
       - name: Prepare ITS Example Parameters
```
milapsheth, line 206:
```suggestion
      ###### Command: ITS ######
      - name: ITS Pause
        run: node sui/contract.js pause InterchainTokenService -y

      - name: ITS Unpause
        run: node sui/contract.js unpause InterchainTokenService -y

      ###### Command: ITS Example ######
      - name: Prepare ITS Example Parameters
```

.github/workflows/test-sui.yaml
```diff
@@ -169,6 +169,12 @@ jobs:
         run: node sui/gas-service.js collectGas --amount 0.1
 
       ###### Command: Gateway ######
+      - name: Gateway Pause
+        run: node sui/contract.js pause AxelarGateway -y
+
+      - name: Gateway Unpause
+        run: node sui/contract.js unpause AxelarGateway -y
+
```
milapsheth, line 177: Add some more coverage by testing other pause variations (with functions, all versions, specific version). Helps catch breaking changes. All variations don't need to be performed for ITS though

sui/contract.js
```diff
@@ -0,0 +1,264 @@
...
+        versionsArg = defaultFunctions.versions;
+        allowedFunctionsArg = defaultFunctions.functionNames;
+    } else {
+        if (options.version === 'all') {
+            throw new Error('Need to specify a version if providing specific functions.');
+        }
+
+        allowedFunctionsArg = functions.split(',');
+        versionsArg = allowedFunctionsArg.map(() => Number(options.version));
+    }
```
milapsheth, line 143:
```suggestion
    } else if (options.version !== 'all') {
        allowedFunctionsArg = functions.split(',');
        versionsArg = allowedFunctionsArg.map(() => Number(options.version));
    } else {
        throw new Error('Need to specify a version if providing specific functions.');
    }
```
reduce nesting

sui/contract.js
```diff
@@ -0,0 +1,264 @@
...
+    let allowedFunctionsArg = [];
+
+    if (functions === SPECIAL_UNPAUSE_FUNCTION_TAGS.DISALLOWED) {
+        versionsArg = contract.disallowedFunctions.versions.slice();
+        allowedFunctionsArg = contract.disallowedFunctions.functionNames.slice();
+    } else if (functions === SPECIAL_UNPAUSE_FUNCTION_TAGS.DEFAULT) {
+        versionsArg = defaultFunctions.versions;
+        allowedFunctionsArg = defaultFunctions.functionNames;
+    } else {
+        if (options.version === 'all') {
```
milapsheth, line 183: same

milapsheth (approved):

sui/contract.js
```diff
@@ -0,0 +1,261 @@
...
+            printInfo(`Functions that will be disallowed for version ${version}`, allowedFunctions);
+
+            versionsArg = versionsArg.concat(new Array(allowedFunctions.length).fill(Number(version)));
+            allowedFunctionsArg = allowedFunctionsArg.concat(allowedFunctions);
+        }
+    } else if (functions === SPECIAL_PAUSE_FUNCTION_TAGS.DEFAULT) {
+        versionsArg = defaultFunctions.versions;
+        allowedFunctionsArg = defaultFunctions.functionNames;
+    } else if (options.version !== 'all') {
+
```
milapsheth, line 137:
```suggestion
```

milapsheth (approved):

PR #522, opened 2025-01-29
feat(stellar): add bytecode download and deploying contracts

ahramy:
[AXE-7282](https://axelarnetwork.atlassian.net/browse/AXE-7282)

[AXE-7282]: https://axelarnetwork.atlassian.net/browse/AXE-7282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stellar/README.md
+++ b/stellar/README.md
@@ -100,7 +100,7 @@ node stellar/deploy-contract.js deploy interchain_token_service --version v1.0.0
 Note that example contract should use `--wasm-path` option to deploy contract
 
 ```bash
-node stellar/deploy-contract.js deploy example --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_example.optimized.wasm
+node stellar/deploy-contract.js deploy example --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_example.optimized.wasm
 ```
 
 ### Contract upgrades
@@ -114,7 +114,7 @@ node stellar/deploy-contract.js deploy upgrader --version v1.0.0
 After the `upgrader` is deployed, any other instantiated contract can be upgraded by calling the `upgrade` function
 
 ```bash
-node stellar/deploy-contract.js upgrade <CONTRACT_NAME> --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/<CONTRACT_NAME>.optimized.wasm --new-version <NEW_VERSION> --migration-data <MIGRATION_DATA>
+node stellar/deploy-contract.js upgrade <CONTRACT_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/<CONTRACT_NAME>.optimized.wasm --new-version <NEW_VERSION> --migration-data <MIGRATION_DATA>
 ```
 
 where `<CONTRACT_NAME>` is the name of the contract to be upgraded and `--wasm-path` points to the upgraded bytecode. As a sanity check, `<NEW_VERSION>` must match the version number defined by the provided bytecode, so upgrading to the wrong version can be prevented. `<MIGRATION_DATA>` is the json encoded data that will be passed to the contract's `migrate` function. If the flag is not provided, the default value `()` will be used, meaning that the migration data is of type `void`. The easiest way to generate the json data for complex types is to instantiate the rust type the contract expects and then use `serde_json::to_string` to convert it to json.
--- a/stellar/README.md
+++ b/stellar/README.md
@@ -68,7 +68,7 @@ stellar contract build
 Deploy the gateway contract
 
 ```bash
-node stellar/deploy-contract.js deploy axelar_gateway --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_axelar_gateway.optimized.wasm
+node stellar/deploy-contract.js deploy axelar_gateway --version v1.0.0
 ```
 
 Provide `--estimate-cost` to show the gas costs for the initialize transaction instead of executing it.
@@ -76,13 +76,13 @@ Provide `--estimate-cost` to show the gas costs for the initialize transaction i
 ### Operators
 
 ```bash
-node stellar/deploy-contract.js deploy axelar_operators --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_axelar_operators.optimized.wasm
+node stellar/deploy-contract.js deploy axelar_operators --version v1.0.0
 ```
 
 ### Gas Service
 
 ```bash
-node stellar/deploy-contract.js deploy axelar_gas_service --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_axelar_gas_service.optimized.wasm
+node stellar/deploy-contract.js deploy axelar_gas_service --version v1.0.0
 ```
 
 ### Interchain Token Service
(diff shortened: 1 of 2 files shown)


stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
+        const response = await fetch(url);
+
+        if (!response.ok) {
+            throw new Error(`Failed to download WASM file: ${response.statusText}`);
+        }
+
+        const buffer = await response.buffer();
+        writeFileSync(outputPath, buffer);
+        printInfo('Successfully downloaded WASM file to', outputPath);
+        return outputPath;
```
nbayindirli, line 64: Any validation we can perform on the file once downloaded?

ahramy: We might add some logic to validate the file such as a checksum, but we don't store the checksum at this moment. I believe the filename and file path `path/contract_name-v1.0.0.wasm` include some sanity checks, and this script will be executed by the chain ops or someone trustworthy for contract deployment. So, I think we don't need to perform a validation check, and I'd prefer to keep the code simple.

nbayindirli (approved):

stellar/README.md
```diff
@@ -68,51 +68,53 @@ stellar contract build
...
-node stellar/deploy-contract.js deploy token_manager --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_token_manager.optimized.wasm
-node stellar/deploy-contract.js deploy interchain_token_service --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_interchain_token_service.optimized.wasm
+node stellar/deploy-contract.js deploy interchain_token --version v1.0.0
+node stellar/deploy-contract.js deploy token_manager --version v1.0.0
+node stellar/deploy-contract.js deploy interchain_token_service --version v1.0.0
 ```
 
 ### Example
 
+Note that example contract should use `--wasm-path` option to deploy contract
```
milapsheth, line 100: Why is this needed? the example wasm should be pushed to R2 as well, just not crates.io. Is that not the case?

stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
+    'axelar_gateway',
+    'axelar_operators',
+    'axelar_gas_service',
+    'interchain_token',
+    'token_manager',
+    'interchain_token_service',
+    'upgrader',
+]);
+
+function getWasmUrl(contractName, version) {
```
milapsheth, line 31: This should be moved to `common/utils` since it'll be relevant for other contracts such as cosmwasm as well

stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
-const { readFileSync } = require('fs');
+const { writeFileSync, readFileSync } = require('fs');
+const fetch = require('node-fetch');
+const path = require('path');
 const {
     utils: { arrayify, id },
 } = ethers;
 require('./cli-utils');
 
+const AXELAR_RELEASE_BASE_URL = 'https://static.axelar.network/releases/axelar-cgp-stellar';
```
milapsheth, line 19: Update the release action to use `releases/stellar` as the base path and move the existing folder on R2. The repo name shouldn't matter (and thus shouldn't cause it to break if we ever change it again)

stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
+
+const SUPPORTED_CONTRACTS = new Set([
+    'axelar_gateway',
```
milapsheth, line 22: Similar to Sui deploy-contract.js, I'd like a mapping between the preferred name in our configs and the actual contract name. The config and CLI input should use `AxelarGateway` etc. Basically, makes it very easy to loop over any chain and get the address for the relevant axelar contract

stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
+
+    return `${AXELAR_RELEASE_BASE_URL}/stellar-${pathName}/${version}/wasm/stellar_${contractName}.wasm`;
+}
+
+async function downloadWasmFile(contractName, version) {
+    const url = getWasmUrl(contractName, version);
+    const tempDir = path.join(process.cwd(), 'artifacts');
+
+    // Create temp directory if it doesn't exist
+    const fs = require('fs');
```
milapsheth, line 46: not a big fan of require's within the function. move to the top?

stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
+
+    try {
+        const response = await fetch(url);
+
+        if (!response.ok) {
+            throw new Error(`Failed to download WASM file: ${response.statusText}`);
+        }
+
+        const buffer = await response.buffer();
+        writeFileSync(outputPath, buffer);
```
milapsheth, line 62: will this overwrite if it exists?

stellar/README.md
```diff
@@ -68,51 +68,53 @@ stellar contract build
...
 
 Deploy Interchain Token and Token Manager wasm first.
 
 ```bash
-node stellar/deploy-contract.js deploy interchain_token --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_interchain_token.optimized.wasm
-node stellar/deploy-contract.js deploy token_manager --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_token_manager.optimized.wasm
-node stellar/deploy-contract.js deploy interchain_token_service --chain-name <CHAIN_NAME> --wasm-path ../axelar-cgp-stellar/target/wasm32-unknown-unknown/release/stellar_interchain_token_service.optimized.wasm
+node stellar/deploy-contract.js deploy interchain_token --version v1.0.0
+node stellar/deploy-contract.js deploy token_manager --version v1.0.0
+node stellar/deploy-contract.js deploy interchain_token_service --version v1.0.0
```
milapsheth, line 95: Also allow pre-releases, i.e if a commit hash is provided instead of vx.y.z, download from the pre-release url instead. Pre-releases will be created once Laith's gh action in cgp-stellar repo is merged. So, once you tackle that PR, you can add this feature

stellar/deploy-contract.js
```diff
@@ -231,7 +297,8 @@ function main() {
     const deployCmd = new Command('deploy')
         .description('Deploy a Stellar contract')
         .argument('<contract-name>', 'contract name to deploy')
-        .addOption(new Option('--wasm-path <wasmPath>', 'path to the WASM file').makeOptionMandatory(true))
+        .addOption(new Option('--wasm-path <wasmPath>', 'path to the WASM file'))
+        .addOption(new Option('--version <version>', 'version of the contract to deploy (e.g., v1.0.0)'))
```
milapsheth, line 301: does commander allow enforcing that only one of these are provided?

stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
+    }
+
+    const pathName = contractName.replace(/_/g, '-');
+
+    return `${AXELAR_RELEASE_BASE_URL}/stellar-${pathName}/${version}/wasm/stellar_${contractName}.wasm`;
+}
+
+async function downloadWasmFile(contractName, version) {
+    const url = getWasmUrl(contractName, version);
+    const tempDir = path.join(process.cwd(), 'artifacts');
```
milapsheth, line 43: `fs` has an ability to create tempfiles, take a look at that perhaps?
https://github.com/axelarnetwork/axelar-cgp-sui/blob/81888b829095522d3d728527398bf63a5dfc5d5b/src/node/node-utils.ts#L15

stellar/deploy-contract.js
```diff
@@ -8,12 +8,77 @@ const { getDomainSeparator, getChainConfig } = require('../common');
...
+    }
+
+    const outputPath = path.join(tempDir, `${contractName}-${version}.wasm`);
+
+    try {
+        const response = await fetch(url);
+
+        if (!response.ok) {
+            throw new Error(`Failed to download WASM file: ${response.statusText}`);
+        }
```
milapsheth, line 59: catch should be added after this to not catch errors from later calls

PR #528, opened 2025-02-03
refactor(sui): convert minimumRotationDelay to seconds

npty:
https://axelarnetwork.atlassian.net/browse/AXE-7248

### DIFF
--- a/sui/deploy-contract.js
+++ b/sui/deploy-contract.js
@@ -239,7 +239,7 @@ async function postDeployAxelarGateway(published, keypair, client, config, chain
         },
         domainSeparator,
         operator,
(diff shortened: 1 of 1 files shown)


sui/deploy-contract.js
```diff
@@ -239,7 +239,7 @@ async function postDeployAxelarGateway(published, keypair, client, config, chain
         },
         domainSeparator,
         operator,
-        minimumRotationDelay,
+        minimumRotationDelay: minimumRotationDelay / 1000, // convert from milliseconds to seconds
```
milapsheth, line 242: Change the value in the mainnet sui config?

PR #533, opened 2025-02-06
feat(stellar): make data field optional for interchain transfer

ahramy:


stellar/its.js
```diff
@@ -102,7 +102,7 @@ async function interchainTransfer(wallet, _, chain, contract, args, options) {
         nativeToScVal(destinationChain, { type: 'string' }),
         hexToScVal(destinationAddress),
         nativeToScVal(amount, { type: 'i128' }),
-        hexToScVal(data),
+        data === '' ? nativeToScVal(null, { type: 'null' }) : hexToScVal(data),
```
milapsheth, line 105:
```suggestion
        data === '' ? nativeToScVal(null, { type: 'null' }) : hexToScVal(data),
```
Data should be a keyword CLI arg, so if it's undefined then the default value gets used

milapsheth (approved):

stellar/its.js
ahramy: updated

PR #534, opened 2025-02-06
docs: improve its v1.1.0 upgrade docs 

RiceAndMeet:


releases/cosmwasm/2025-01-ITS-v1.1.0.md
```diff
@@ -43,6 +43,17 @@ node cosmwasm/submit-proposal.js migrate \
   --deposit 2000000000
 ```
 
+Verify ITS hub contract version
```
milapsheth, line 46: Move this under the checklist section?

releases/evm/2025-01-ITS-v2.1.0.md
```diff
@@ -87,6 +87,12 @@ Migration Steps
 
 We’d also like to migrate minter for all native ITS tokens if they weren’t auto migrated, so we can remove the migration logic in the future. We’ll create scripts to do that in the future since this is not a blocker for the upgrade.
 
+### Verify Upgraded ITS Contracts
+
+```bash
```
milapsheth, line 92: Link to the verification instructions since some other setup needs to be done but we don't need to duplicate cmds here

releases/evm/2025-01-ITS-v2.1.0.md
```diff
@@ -118,8 +124,15 @@ node evm/interchainTokenFactory.js --action linkToken --destinationChain chain2
 node evm/its.js --action transferMintership --tokenAddress [tokenAddress] --minter [tokenManager]
 ```
 
+> [!NOTE]  
+> `tokenManager` address is different per token, which can be found on tx output from `registerCustomToken` action on chain
```
milapsheth, line 128: This should be before the cmd that needs the token manager not after someone runs it incorrectly. Also, instead of a note, show how to use `evm/its.js --action tokenManagerAddress ...` to get the address on each chain as a step

releases/evm/2025-01-ITS-v2.1.0.md
```diff
@@ -118,8 +124,15 @@ node evm/interchainTokenFactory.js --action linkToken --destinationChain chain2
...
+
 - [ ]  Interchain Transfer (both ways)
 
 ```bash
 node evm/its.js --action interchainTransfer --destinationChain chain2 --tokenId [tokenId] --destinationAddress [recipient] --amount 1 --gasValue 0
 ```
+
+> [!NOTE]
```
milapsheth, line 136: Same, note shouldn't be at the end. Add a sentence as part of register custom token step to record the tokenId from the output

releases/evm/2025-01-ITS-v2.1.0.md
```diff
@@ -110,29 +108,35 @@ node evm/its.js --action registerTokenMetadata --tokenAddress [tokenAddress]
...
 
 # Link to remote token. Ensure GMP call is executed
 node evm/interchainTokenFactory.js --action linkToken --destinationChain chain2 --destinationTokenAddress [remote token address] --tokenManagerType 4 --linkParams "0x" --salt "salt6789"
 ```
 
+- [ ] Fetch tokenManager address for deployed token on both chains
+
+> [!NOTE] 
+> Use `tokenId` from `registerCustomToken` tx output
+
```
milapsheth, line 122:
```suggestion
```
Not needed since it was mentioned above

releases/evm/2025-01-ITS-v2.1.0.md
```diff
@@ -110,29 +108,35 @@ node evm/its.js --action registerTokenMetadata --tokenAddress [tokenAddress]
...
 ```
 
-> [!NOTE]  
-> `tokenManager` address is different per token, which can be found on tx output from `registerCustomToken` action on chain
-
 - [ ]  Interchain Transfer (both ways)
 
+> [!NOTE] 
+> Use `tokenId` from `registerCustomToken` tx output
+
```
milapsheth, line 139:
```suggestion
```
Same. We want to aim for the docs to be as concise as possible for readability while being precise

milapsheth (approved):

evm/utils.js
```diff
@@ -210,7 +210,7 @@ function isValidBytesAddress(input) {
 }
 
 function isValidBytesArray(input) {
-    const bytesRegex = /^0x[a-fA-F0-9]/;
+    const bytesRegex = /^0x([a-fA-F0-9]{40}|)$/;
```
milapsheth, line 213:
```suggestion
    const bytesRegex = /^0x([a-fA-F0-9])?$/;
```
Why add 40 chats? This is an arbitrary byte array

milapsheth, line 213: Also check that the length is even and fail early if not

milapsheth (approved):

PR #544, opened 2025-02-11
refactor: reorganize deploy-contract command structure

npty:


stellar/deploy-contract.js
```diff
@@ -296,39 +321,50 @@ function main() {
...
-            mainProcessor(options, upgrade, contractName);
-        });
+    const upgradeCmd = new Command('upgrade').description('Upgrade a Stellar contract');
 
     // Add base options to all 2nd level commands
     addBaseOptions(upgradeCmd, { address: true });
     addBaseOptions(deployCmd, { address: true });
 
+    // 3rd level commands for `deploy`
+    const deployContractCmds = Array.from(SUPPORTED_CONTRACTS).map((contractName) => {
```
ahramy, line 334: Does this work for contracts not listed in `SUPPORTED_CONTRACTS`? Just wanted to double check.

npty: No, it doesn't work because the commands are generated from `SUPPORTED_CONTRACTS`. If we need more contracts in the deploy commands, we need to add the package name into the array.

ahramy: I see. then could you add `example` to the `SUPPORTED_CONTRACTS`? The `example` contract could be deployed without calling `getWasmUrl` function, but with this change, it must be included in `SUPPORTED_CONTRACTS`.

ahramy (approved):

stellar/deploy-contract.js
npty: Done

PR #546, opened 2025-02-11
feat: rotation script

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-7233

### DIFF
--- a/axelar-chains-config/info/mainnet.json
+++ b/axelar-chains-config/info/mainnet.json
@@ -2017,7 +2017,8 @@
           "salt": "Multisig v5.5"
         },
         "InterchainTokenService": {
-          "skip": true
+          "skip": true,
+          "address": ""
         },
         "Operators": {
           "owner": "0x6f24A47Fc8AE5441Eb47EFfC3665e70e69Ac3F05",
@@ -2681,9 +2682,7 @@
       "finality": "1",
       "tokenSymbol": "XLM",
       "tokenAddress": "CAS3J7GYLGXMF6TDJBBYYSE3HQ6BBSMLNUQ34T6TZMYMW2EVH34XOWMA",
-      "contracts": {
-        "AxelarGateway": {}
-      },
+      "contracts": {},
       "approxFinalityWaitTime": 1,
       "explorer": {
         "name": "Stellar Expert",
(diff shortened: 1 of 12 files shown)


sui/rotate.sh
```diff
@@ -0,0 +1,172 @@
...
+
+# Functions
+set_namespace() {
+    print_section "Setting Namespace"
+    kubectl config set-context --current --namespace="$NAMESPACE"
+    print_success "Namespace set to $NAMESPACE"
+}
+
+query_current_verifier_set() {
+    kubectl exec -it "$BASTION" -c "$BASTION_CONTAINER" -- bash -c "axelard q wasm contract-state smart $1 '\"current_verifier_set\"'"
```
milapsheth, line 42: Internal infra details/cmds shouldn't be added to a public repo

sui/rotate.sh
```diff
@@ -0,0 +1,172 @@
...
+}
+
+query_next_verifier_set() {
```
milapsheth, line 45: Avoid creating bash scripts, they're not very readable and maintainable. Create a JS script instead. We already have JS examples on how to interact with Axelar contracts. This script will be useful for performing rotations for any connections, so convert to JS and make it more generic. Depending on `chainType`, it can call the `submitProof` helper from the relevant Sui/Stellar/EVM for simplicity

milapsheth: The script should only do a single rotation, the caller can loop over it if needed

sui/rotate.sh
```diff
@@ -0,0 +1,172 @@
...
+get_worker_address() {
+    kubectl exec -it "$WORKER" -n "$NAMESPACE2" -c ampd-verifier-address -- bash -c "cat verifier-address"
+}
+
+query_active_verifiers() {
+    kubectl exec -it "$BASTION" -c "$BASTION_CONTAINER" -- bash -c \
+        "axelard q wasm contract-state smart $SERVICE_REGISTRY '{\"active_verifiers\":{\"service_name\": \"$SERVICE_NAME\",\"chain_name\":\"sui\"}}'"
+}
+
+register_or_deregister() {
```
milapsheth, line 58: Shouldn't be part of this script. The caller can do this before/after calling the rotation script

PR #556, opened 2025-02-13
refactor: update its scripts

blockchainguyy:


.github/workflows/test-evm.yaml
```diff
@@ -119,10 +119,10 @@ jobs:
         run: node evm/deploy-its.js -m create2 -u -y
```
milapsheth, line 119: Update orchestration scripts and merge together with this PR to avoid breaking it

.github/workflows/test-evm.yaml
```diff
@@ -119,10 +119,10 @@ jobs:
...
 
       - name: InterchainTokenService set hub
-        run: node evm/its.js --action setTrustedAddress --trustedChain axelar --trustedAddress axelar1xyz -y
+        run: node evm/its.js setTrustedAddress --trustedChain axelar --trustedAddress axelar1xyz -y
```
milapsheth, line 122:
```suggestion
        run: node evm/its.js set-trusted-address axelar axelar1xyz -y
```
use kebab-case and convert args that are required to positional. See sui scripts for reference

evm/its.js
```diff
@@ -715,7 +689,8 @@ if (require.main === module) {
     program.addOption(new Option('--pauseStatus <pauseStatus>', 'pause status').choices(['true', 'false']));
     program.addOption(new Option('--rawSalt <rawSalt>', 'raw deployment salt').env('RAW_SALT'));
 
-    program.action((options) => {
+    program.action((command, options) => {
+        options.action = command;
```
milapsheth, line 693: Avoid quick and dirty workarounds. Refactor it properly (see Sui scripts for reference)

sui/its.js
```diff
@@ -5,13 +5,18 @@ const { addBaseOptions, addOptionsToCommands, getWallet, printWalletInfo, broadc
 const { bcs } = require('@mysten/sui/bcs');
 
 const SPECIAL_CHAINS_TAGS = {
-    ALL_EVM: 'all-evm', // All EVM chains that have InterchainTokenService deployed
+    ALL: 'all',
```
milapsheth, line 8: This change should be in a separate PR than the above refactor since they're unrelated and can be addressed independently without blocking each other

sui/its.js
```diff
@@ -5,13 +5,18 @@ const { addBaseOptions, addOptionsToCommands, getWallet, printWalletInfo, broadc
...
 };
 
+function getAllTrustedChains(config) {
```
milapsheth, line 11: This should be a common/util that gets used in EVM `deploy-its.js` as well
https://github.com/axelarnetwork/axelar-contract-deployments/blob/feat/ITS-hub-upgrade-v1.1.0/evm/deploy-its.js#L135

sui/its.js
```diff
@@ -5,13 +5,18 @@ const { addBaseOptions, addOptionsToCommands, getWallet, printWalletInfo, broadc
...
+    return allChains;
+}
+
 function parseTrustedChains(config, trustedChains) {
-    if (trustedChains === SPECIAL_CHAINS_TAGS.ALL_EVM) {
-        const evmChains = Object.keys(config.chains).filter((chain) => config.chains[chain].contracts?.InterchainTokenService?.address);
-        return evmChains;
+    if (trustedChains[0] === SPECIAL_CHAINS_TAGS.ALL) {
+        const allChains = getAllTrustedChains(config);
+        return allChains.filter((chain) => config.chains[chain].contracts?.InterchainTokenService?.address);
```
milapsheth, line 19: The filter should be done in the helper. It shouldn't be giving invalid ITS chains to begin with

milapsheth, line 19: EVM ITS uses `skip` identifier for filtering. Switch that to use the new helper as well and update configs appropriately (ITS contract config can be removed from those chains altogether)

sui/its.js
```diff
@@ -5,13 +5,18 @@ const { addBaseOptions, addOptionsToCommands, getWallet, printWalletInfo, broadc
...
     }
 
     return trustedChains;
```
milapsheth, line 22: If a special tag isn't used, should at least check that the chain names correspond to correct `axelarId` in the config and warn if not (it's possible a third party chain, e..g xrpl isn't in the config but want to catch user errors in other scenarios)

sui/its.js
```diff
@@ -5,13 +5,18 @@ const { addBaseOptions, addOptionsToCommands, getWallet, printWalletInfo, broadc
...
+function getAllTrustedChains(config) {
+    const allChains = Object.keys(config.chains);
+    return allChains;
+}
+
 function parseTrustedChains(config, trustedChains) {
-    if (trustedChains === SPECIAL_CHAINS_TAGS.ALL_EVM) {
-        const evmChains = Object.keys(config.chains).filter((chain) => config.chains[chain].contracts?.InterchainTokenService?.address);
-        return evmChains;
+    if (trustedChains[0] === SPECIAL_CHAINS_TAGS.ALL) {
```
milapsheth, line 17: check length as well

sui/its.js
```diff
@@ -5,13 +5,18 @@ const { addBaseOptions, addOptionsToCommands, getWallet, printWalletInfo, broadc
 const { bcs } = require('@mysten/sui/bcs');
 
 const SPECIAL_CHAINS_TAGS = {
```
milapsheth, line 7: not really needed now?

sui/its.js
milapsheth, line 11:
```suggestion
function getITSChains(config) {
```

milapsheth, line 11: This could take the `trustedChains` list and apply specialization internally, and ensure that the names match a valid `axelarId`

PR #577, opened 2025-02-25
feat(evm): add legacy ITS custom token linkToken support

milapsheth:


evm/its.js
```diff
@@ -696,6 +747,12 @@ if (require.main === module) {
     program.addOption(new Option('--destinationAddress <destinationAddress>', 'destination address'));
     program.addOption(new Option('--params <params>', 'params for TokenManager deployment'));
     program.addOption(new Option('--tokenAddress <tokenAddress>', 'token address to use for token manager deployment'));
+    program.addOption(
+        new Option(
+            '--destinationTokenAddress <destinationTokenAddress>',
+            'destination token address on the destination chain to the source token to',
```
isi8787, line 753:
```suggestion
            'token address on the destination chain for the source token',
```

milapsheth, line 753:
```suggestion
            'token address on the destination chain to link with the token on the source chain corresponding to the tokenId',
```

PR #585, opened 2025-03-04
chore: stellar stagenet rollout

RiceAndMeet:


axelar-chains-config/info/stagenet.json
```diff
@@ -2031,6 +2105,21 @@
           "domainSeparator": "0x254a1ae90417f8649798698e95a1140808a701c38c8f5723a930b9d0666e54b4",
           "address": "axelar1ktjfpfrlgmh8s76ltjhhv0xcj0st0d0xrjh5rvm3gr5u66x9xvpsm0juw8"
         },
+        "stellar": {
```
milapsheth, line 2108: wrong chain name, should have been `stellar-2024-q4`, please follow the release doc carefully. we can let this one through, but the next addition should use the correct naming

evm/gateway.js
```diff
@@ -318,7 +318,16 @@ async function processCommand(config, chain, options) {
...
-            const { commandId, destination, payloadHash, sourceChain, sourceAddress } = options;
+            const { destination, payloadHash, sourceChain, sourceAddress } = options;
+
+            let commandId;
+
+            if (options.messageId) {
+                // Derive commandId for Amplifier gateway
+                commandId = id(`${sourceChain}_${options.messageId}`);
+            } else {
+                commandId = options.commandID.startsWith('0x') ? options.commandID : id(parseInt(options.commandID).toString());
```
milapsheth, line 329: why is an integer commandId being passed in? only for testing? not very obvious, should add comments. Another option is to pad the 0x value to make it 32 bytes so the same format can be used

milapsheth (approved):

axelar-chains-config/info/stagenet.json
RiceAndMeet: Oh man ...

evm/gateway.js
RiceAndMeet: commandId failed to pass validation check when calling function here hence the conversion. This is only for testing.

milapsheth: can we revert the change then? There is also an `isMessageApproved` method that can be used for amplifier EVM chains

RiceAndMeet: Revert the change

PR #588, opened 2025-03-06
feat(stellar): add support to deploy-contract upgrade for versioned WASM file download

nbayindirli:
[AXE-7920](https://axelarnetwork.atlassian.net/browse/AXE-7920)

Tested successfully via `v1.1.0` Stellar release.

[AXE-7920]: https://axelarnetwork.atlassian.net/browse/AXE-7920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stellar/contract.js
+++ b/stellar/contract.js
@@ -11,22 +11,29 @@ require('./cli-utils');
 
 const MAX_INSTANCE_TTL_EXTENSION = 535679;
 
-async function submitOperation(wallet, chain, _contractName, contract, operation, args, options, showReturnValue = true) {
+async function submitOperation(wallet, chain, _contractName, contract, args, options, operation = '') {
+    if (!operation) {
+        operation = args.operation;
+    }
+
     const callOperation = Array.isArray(args) ? await contract.call(operation, ...args) : await contract.call(operation);
 
-    const returnValue = await broadcast(callOperation, wallet, chain, `${operation}`, options);
+    const returnValue = await broadcast(callOperation, wallet, chain, `${operation}`, options, args.simulate);
+    const result = args.simulate ? returnValue.result.retval._value : returnValue.value();
 
-    if (showReturnValue && returnValue.value()) {
-        printInfo(`${operation} returned`, serializeValue(returnValue.value()));
+    if (result !== undefined) {
+        printInfo(`${_contractName}:${operation} returned`, serializeValue(result));
+    } else {
+        printInfo(`${_contractName}:${operation} succeeded`);
     }
 }
 
 async function transferOwnership(wallet, chain, _contractName, contract, args, options) {
-    return await submitOperation(wallet, chain, _contractName, contract, 'transfer_ownership', [addressToScVal(args)], options, false);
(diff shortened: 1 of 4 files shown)


stellar/deploy-contract.js
```diff
@@ -289,7 +291,7 @@ async function createUpgradeAuths(contractAddress, newWasmHash, migrationData, c
     return Promise.all(
         [
             createAuthorizedFunc(contractAddress, 'upgrade', [nativeToScVal(newWasmHash)]),
-            createAuthorizedFunc(contractAddress, 'migrate', [nativeToScVal(migrationData)]),
+            migrationData ? createAuthorizedFunc(contractAddress, 'migrate', [nativeToScVal(migrationData)]) : null,
```
ahramy, line 294: Let's make sure this works by testing upgrade

stellar/deploy-contract.js
```diff
@@ -349,7 +351,7 @@ function main() {
         return new Command(contractName)
             .description(`Upgrade ${contractName} contract`)
             .addOption(new Option('--wasm-path <wasmPath>', 'path to the WASM file'))
-            .addOption(new Option('--new-version <newVersion>', 'new version of the contract'))
+            .addOption(new Option('--new-version <newVersion>', 'new version of the contract to upgrade to (e.g., v1.1.0)'))
```
ahramy, line 354: Could we just make this `--version` so that we don't need to pass `version` as argument? since `upgrade` implies it

stellar/deploy-contract.js
```diff
@@ -207,7 +208,7 @@ async function deploy(options, config, chain, contractName) {
         return;
     }
 
-    const wasmPath = await getWasmPath(options, contractName);
+    const wasmPath = await getWasmFile(options.wasmPath, options.version, contractName);
```
ahramy, line 211: I think if we use `--version` instead of  `--new-version` for the upgrade, we don't need to change it here

stellar/deploy-contract.js
ahramy: yeah, then let's use `--new-version`

milapsheth: Hmm, I'd prefer to keep it simpler by using `--version`. We're also using `--wasm-path` rather than `--new-wasm-path`. Instead, the script should check if the provided version is greater than the deployed version (which it can query from the contract) for better validation

stellar/deploy-contract.js
milapsheth, line 294: Isn't the `migrate` method always called, regardless of if the migration data is empty?

stellar/README.md
```diff
@@ -114,49 +114,49 @@ node stellar/deploy-contract.js deploy upgrader --version v1.0.0
 After the `upgrader` is deployed, any other instantiated contract can be upgraded by calling the `upgrade` function
```
milapsheth, line 114: PR title should be `feat(stellar): ...`

PR #592, opened 2025-03-07
fix: pass correct migration default argument (expected type VOID)

cgorenflo:


stellar/deploy-contract.js
```diff
@@ -375,6 +376,35 @@ function main() {
...
+
+    if (Array.isArray(parsed)) {
+        return parsed.map(sanitizeMigrationData);
+    }
+
+    if (parsed !== null && typeof parsed === 'object') {
+        return Object.fromEntries(Object.entries(parsed).map(([key, value]) => [key, sanitizeMigrationData(value)]));
+    }
+
+    return parsed;
```
ahramy, line 405: It would be great to include some migration data examples for each case in the README or perhaps in a comment

PR #600, opened 2025-03-10
refactor(stellar): rename repo to axelar amplifier stellar

ahramy:
https://axelarnetwork.atlassian.net/browse/AXE-7986

### DIFF
--- a/stellar/deploy-contract.js
+++ b/stellar/deploy-contract.js
@@ -25,7 +25,7 @@ const {
 } = ethers;
 require('./cli-utils');
 
-const AXELAR_RELEASE_BASE_URL = 'https://static.axelar.network/releases/axelar-cgp-stellar';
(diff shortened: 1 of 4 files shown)


stellar/deploy-contract.js
```diff
@@ -25,7 +25,7 @@ const {
 } = ethers;
 require('./cli-utils');
 
-const AXELAR_RELEASE_BASE_URL = 'https://static.axelar.network/releases/axelar-cgp-stellar';
+const AXELAR_RELEASE_BASE_URL = 'https://static.axelar.network/releases/axelar-amplifier-stellar';
```
milapsheth, line 28: before merging this, we also need to rename it on R2 at the same time to avoid breaking the deployment

milapsheth (approved):

PR #615, opened 2025-03-18
feat(interchain-token-service): encode recipient based on chain type

ahramy:
[AXE-7392](https://axelarnetwork.atlassian.net/browse/AXE-7392)

[AXE-7392]: https://axelarnetwork.atlassian.net/browse/AXE-7392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -541,7 +541,7 @@ function asciiToBytes(string) {
  *       - Additional encoding logic can be added for new chain types.
  */
 function encodeITSDestination(config, destinationChain, destinationAddress) {
-    const { chainType } = config.chains[destinationChain] || config[destinationChain];
+    const chainType = config.chains[destinationChain]?.chainType || config[destinationChain]?.chainType;
 
     switch (chainType) {
         case 'stellar':
--- a/common/utils.js
+++ b/common/utils.js
@@ -541,7 +541,7 @@ function asciiToBytes(string) {
  *       - Additional encoding logic can be added for new chain types.
  */
 function encodeITSDestination(config, destinationChain, destinationAddress) {
-    const { chainType } = getChainConfig(config, destinationChain);
+    const { chainType } = config.chains[destinationChain] || config[destinationChain];
 
     switch (chainType) {
         case 'stellar':
--- a/common/utils.js
+++ b/common/utils.js
@@ -529,6 +529,31 @@ const parseTrustedChains = (config, trustedChains) => {
     return trustedChains.length === 1 && trustedChains[0] === 'all' ? itsEdgeChains(config) : trustedChains;
 };
 
+function stellarAddressToBytes(address) {
+    return hexlify(Buffer.from(address, 'ascii'));
+}
+
+function encodeDestinationAddress(destinationChain, destinationAddress) {
+    if (!destinationChain || !destinationAddress) {
+        throw new Error('Both destinationChain and destinationAddress are required');
+    }
+
+    const chain = destinationChain.toLowerCase();
+
+    if (chain.includes('stellar')) {
+        validateParameters({
+            isValidStellarAddress: { destinationAddress },
+        });
+
+        return stellarAddressToBytes(destinationAddress);
+    } else if (chain.includes('sui')) {
+        // TODO: add sui specific address encoding function
+        return destinationAddress;
+    } else {
+        return destinationAddress;
+    }
+}
+
 module.exports = {
     loadConfig,
     saveConfig,
@@ -574,4 +599,6 @@ module.exports = {
     isValidStellarAddress,
     isValidStellarAccount,
     isValidStellarContract,
+    stellarAddressToBytes,
+    encodeDestinationAddress,
 };
--- a/common/utils.js
+++ b/common/utils.js
@@ -529,29 +529,30 @@ const parseTrustedChains = (config, trustedChains) => {
     return trustedChains.length === 1 && trustedChains[0] === 'all' ? itsEdgeChains(config) : trustedChains;
 };
 
-function stellarAddressToBytes(address) {
-    return hexlify(Buffer.from(address, 'ascii'));
+function asciiToBytes(string) {
+    return hexlify(Buffer.from(string, 'ascii'));
 }
 
-function encodeDestinationAddress(destinationChain, destinationAddress) {
-    if (!destinationChain || !destinationAddress) {
-        throw new Error('Both destinationChain and destinationAddress are required');
+/**
+ * Encodes the destination address for Interchain Token Service (ITS) transfers.
+ * This function ensures proper encoding of the destination address based on the destination chain type.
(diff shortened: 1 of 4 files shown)


common/utils.js
```diff
@@ -529,6 +529,31 @@ const parseTrustedChains = (config, trustedChains) => {
...
+
+    const chain = destinationChain.toLowerCase();
+
+    if (chain.includes('stellar')) {
+        validateParameters({
+            isValidStellarAddress: { destinationAddress },
+        });
+
+        return stellarAddressToBytes(destinationAddress);
+    } else if (chain.includes('sui')) {
```
milapsheth, line 549: what if chain name includes `sui` substring? Use a better approach. Pass in the actual `chainType` from the config for the chain. The helper can receive this and then have a switch case on it. The default behaviour can to encode the bytes as-is. This is the correct behaviour for EVM and Sui

common/utils.js
```diff
@@ -529,6 +529,31 @@ const parseTrustedChains = (config, trustedChains) => {
     return trustedChains.length === 1 && trustedChains[0] === 'all' ? itsEdgeChains(config) : trustedChains;
 };
 
+function stellarAddressToBytes(address) {
+    return hexlify(Buffer.from(address, 'ascii'));
+}
+
+function encodeDestinationAddress(destinationChain, destinationAddress) {
```
milapsheth, line 536:
```suggestion
function encodeITSDestination(destinationChain, destinationAddress) {
```
this is specifically for ITS, not clear from the name. Maybe explain via a docstring?

common/utils.js
```diff
@@ -529,6 +529,31 @@ const parseTrustedChains = (config, trustedChains) => {
...
+    if (!destinationChain || !destinationAddress) {
```
milapsheth, line 537: why are we doing extra validation here (and validateParameters should be used when doing so)? The caller should be sending non null values to begin with, otherwise every helper function needs to start doing this.

common/utils.js
```diff
@@ -529,6 +529,31 @@ const parseTrustedChains = (config, trustedChains) => {
     return trustedChains.length === 1 && trustedChains[0] === 'all' ? itsEdgeChains(config) : trustedChains;
 };
 
+function stellarAddressToBytes(address) {
```
milapsheth, line 532:
```suggestion
function asciiToBytes(string) {
```
make it more generic wrt what it's actually doing. Other chains might do the same too

common/utils.js
ahramy: make sense. removed extra validation

common/utils.js
ahramy: make sense, updated.

common/utils.js
ahramy: make sense, updated

common/utils.js
ahramy: I thought the same, but I just noticed that `chainType` exists from your comment. updated.

common/utils.js
```diff
@@ -529,6 +529,32 @@ const parseTrustedChains = (config, trustedChains) => {
...
+
+/**
+ * Encodes the destination address for Interchain Token Service (ITS) transfers.
+ * This function ensures proper encoding of the destination address based on the destination chain type.
+ * Note: - Stellar addresses are converted to ASCII byte arrays.
+ *       - EVM and Sui addresses are returned as-is (default behavior).
+ *       - Additional encoding logic can be added for new chain types.
+ */
+function encodeITSDestination(config, destinationChain, destinationAddress) {
+    const { chainType } = config.chains[destinationChain] || config[destinationChain];
```
ahramy, line 544: note that `getChainConfig(config, destinationChain)` causes an error during the evm-test : 
```
node evm/its.js --action interchainTransfer --destinationChain remote --tokenId 0x88a9d17b8f4e6e4aaceb3c8f53d54eedb144276f1dd2b9f2d17de784aa090be7 --destinationAddress 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --amount 1 --gasValue 0 -y

/home/runner/_work/axelar-contract-deployments/axelar-contract-deployments/common/utils.js:474
        throw new Error(`Chain ${chainName} not found in config`);
              ^
Error: Chain remote not found in config
```

milapsheth (approved):

common/utils.js
ahramy: that's true, updated with `skipCheck`

stellar/its.js
```diff
@@ -179,7 +178,7 @@ async function execute(wallet, _, chain, contract, args, options) {
 
 async function encodeRecipient(wallet, _, chain, contract, args, options) {
     const [recipient] = args;
-    printInfo('Encoded Recipient', stellarAddressToBytes(recipient));
+    printInfo('Encoded Recipient', asciiToBytes(recipient));
```
milapsheth, line 181: use `encodeITSDestination` instead of the low level util

evm/its.js
```diff
@@ -348,10 +349,18 @@ async function processCommand(config, chain, options) {
                 await token.approve(interchainTokenService.address, amount, gasOptions).then((tx) => tx.wait());
             }
 
+            const destinationAddressEncoded = encodeITSDestination(config, destinationChain, destinationAddress);
+
+            if (destinationAddressEncoded !== destinationAddress) {
+                printInfo(
+                    `The destination address "${destinationAddress}" was encoded as "${destinationAddressEncoded}" for "${destinationChain}".`,
+                );
+            }
```
milapsheth, line 358:
```
const itsDestinationAddress = encodeITSDestination(config, destinationChain, destinationAddress);
printInfo('Human-readable Destination address', destinationAddress);
printInfo('Encoded ITS destination address', itsDestinationAddress);
```
Maybe always print for consistency?

evm/its.js
```diff
@@ -348,10 +349,18 @@ async function processCommand(config, chain, options) {
                 await token.approve(interchainTokenService.address, amount, gasOptions).then((tx) => tx.wait());
             }
 
+            const destinationAddressEncoded = encodeITSDestination(config, destinationChain, destinationAddress);
```
milapsheth, line 352: The encoding needs to be applied for ITS minter address as well when deploying a token

stellar/its.js
ahramy: yeah, I think we can remove `encodeRecipient function in stellar, and use `encodeITSDestination` for ITS in stellar.

evm/its.js
ahramy: updated

evm/its.js
ahramy: that's true, but there is no `deployRemoteInterchainTokenWithMinter` call from the contract deployment script. Hence skip it for now, we can work on it when it's needed.

PR #617, opened 2025-03-19
feat: add contract state for cosmwasm contracts

ahramy:
[AXE-8190](https://axelarnetwork.atlassian.net/browse/AXE-8190)

### DIFF
--- a/cosmwasm/cli-utils.js
+++ b/cosmwasm/cli-utils.js
@@ -80,6 +80,13 @@ const addAmplifierOptions = (program, options) => {
     }
 };
 
+const addAmplifierQueryOptions = (program) => {
+    addEnvOption(program);
+
+    program.addOption(new Option('-m, --mnemonic <mnemonic>', 'mnemonic').makeOptionMandatory(true).env('MNEMONIC'));
+    program.addOption(new Option('-n, --chainName <chainName>', 'chain name').env('CHAIN').argParser((value) => value.toLowerCase()));
+};
+
 const addContractOptions = (program) => {
     program.addOption(new Option('-c, --contractName <contractName>', 'contract name').makeOptionMandatory(true));
     program.addOption(new Option('-n, --chainName <chainName>', 'chain name').env('CHAIN').argParser((value) => value.toLowerCase()));
@@ -169,4 +176,5 @@ const addProposalOptions = (program) => {
 
 module.exports = {
     addAmplifierOptions,
+    addAmplifierQueryOptions,
 };
--- a/cosmwasm/contract-state.js
+++ b/cosmwasm/contract-state.js
@@ -1,0 +1,71 @@
+'use strict';
+
+require('dotenv').config();
+
+const { prepareWallet, prepareClient, initContractConfig } = require('./utils');
+const { loadConfig, printInfo, printWarn } = require('../common');
+const { Command } = require('commander');
+const { addAmplifierQueryOptions } = require('./cli-utils');
+
+const CONTRACT_MAP = {
+    multisig: (config) => config.axelar.contracts.Multisig.address,
+    voting_verifier: (config, chainName) => config.axelar.contracts.VotingVerifier?.[chainName]?.address,
+};
+
+const rewards = async (client, config, options) => {
+    const getContractAddress = CONTRACT_MAP[options.contract];
+
+    if (!getContractAddress) {
+        printWarn(`Query to ${options.contract} is not supported.`);
+        return;
+    }
+
+    try {
+        const result = await client.queryContractSmart(config.axelar.contracts.Rewards.address, {
+            rewards_pool: {
+                pool_id: {
+                    chain_name: options.chainName,
+                    contract: getContractAddress(config, options.chainName),
+                },
+            },
+        });
+
+        printInfo(`Rewards pool for ${options.contract} on ${options.chainName}`, JSON.stringify(result, null, 2));
+    } catch (error) {
+        printWarn(`Failed to fetch rewards pool for ${options.contract} on ${options.chainName}`, `${error.message}`);
+    }
+};
+
+const mainProcessor = async (processor, options) => {
+    const { env } = options;
+    const config = loadConfig(env);
+
+    initContractConfig(config, options);
+
+    const wallet = await prepareWallet(options);
+    const client = await prepareClient(config, wallet);
+
+    await processor(client, config, options);
+};
+
+const programHandler = () => {
+    const program = new Command();
+
+    program.name('contract-state').description('Query contrct state');
+
+    const rewardCmd = program
+        .command('rewards')
+        .description('Query rewards contract state')
+        .argument('<contract>', 'Contract to query rewards')
+        .action((contract, options) => {
+            options.contract = contract;
+            mainProcessor(rewards, options);
+        });
+    addAmplifierQueryOptions(rewardCmd);
+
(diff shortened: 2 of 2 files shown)


cosmwasm/contract-state.js
milapsheth, line 1: call it `contract.js`?

cosmwasm/cli-utils.js
```diff
@@ -80,6 +80,13 @@ const addAmplifierOptions = (program, options) => {
     }
 };
 
+const addAmplifierQueryOptions = (program) => {
+    addEnvOption(program);
+
+    program.addOption(new Option('-m, --mnemonic <mnemonic>', 'mnemonic').makeOptionMandatory(true).env('MNEMONIC'));
```
milapsheth, line 86: why is a mnemonic needed for querying?

cosmwasm/contract-state.js
```diff
@@ -0,0 +1,71 @@
...
+
+    program.name('contract-state').description('Query contrct state');
+
+    const rewardCmd = program
+        .command('rewards')
+        .description('Query rewards contract state')
+        .argument('<contract>', 'Contract to query rewards')
+        .action((contract, options) => {
+            options.contract = contract;
+            mainProcessor(rewards, options);
```
milapsheth, line 63:
```suggestion
            mainProcessor(rewards, contract, options);
```
weird pattern of converting positional arg into keyword arg

cosmwasm/contract-state.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+};
+
+const mainProcessor = async (processor, options) => {
+    const { env } = options;
+    const config = loadConfig(env);
+
+    initContractConfig(config, options);
+
+    const wallet = await prepareWallet(options);
```
milapsheth, line 46: why is a wallet needed? try to avoid blindly copy pasting

cosmwasm/contract-state.js
```diff
@@ -0,0 +1,72 @@
...
+
+const programHandler = () => {
+    const program = new Command();
+
+    program.name('contract-state').description('Query contrct state');
+
+    const rewardCmd = program
+        .command('rewards')
+        .description('Query rewards contract state')
+        .argument('<contract>', 'Contract to query rewards')
```
milapsheth, line 60:
```suggestion
        .argument('<chain-name>', 'Query chain reward pools')
```
chain name should be an arg. It might not be an input for every cmd

milapsheth: Furthermore, just print both reward pools for the chain instead of making the user call twice? The goal of our scripts isn't to be a wrapper over CLI cmds, so I recommend thinking a step further on how to make the UX conceptually simpler

cosmwasm/contract-state.js
```diff
@@ -0,0 +1,72 @@
...
+    const wallet = await prepareWallet(options);
+    const client = await prepareClient(config, wallet);
+
+    await processor(client, config, options);
+};
+
+const programHandler = () => {
+    const program = new Command();
+
+    program.name('contract-state').description('Query contrct state');
```
milapsheth, line 55:
```suggestion
    program.name('query').description('Query contract state');
```

cosmwasm/contract-state.js
milapsheth, line 1: Call it `query.js` or `contract.js` maybe?

cosmwasm/contract-state.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+require('dotenv').config();
```
milapsheth, line 3: this is done in `common/cli-utils.js` already
```suggestion
```

cosmwasm/contract-state.js
```diff
@@ -0,0 +1,72 @@
...
+
+const { prepareWallet, prepareClient, initContractConfig } = require('./utils');
+const { loadConfig, printInfo, printWarn } = require('../common');
+const { Command } = require('commander');
+const { addAmplifierQueryOptions } = require('./cli-utils');
+
+const CONTRACT_MAP = {
```
milapsheth, line 10: avoid hardcoded maps. take contract name in pascal case and do a simple look up. If chainName key exists, then query the nested address instead

cosmwasm/contract-state.js
```diff
@@ -0,0 +1,72 @@
...
+    multisig: (config) => config.axelar.contracts.Multisig.address,
+    voting_verifier: (config, chainName) => config.axelar.contracts.VotingVerifier?.[chainName]?.address,
+};
+
+const rewards = async (client, config, options) => {
```
milapsheth, line 15: we use `async function` for cmd actions in other scripts. We should pick a consistent pattern

cosmwasm/contract-state.js
milapsheth: yeah, we should create a schema for our configs so it can be read in a typed manner

cosmwasm/cli-utils.js
```diff
@@ -80,6 +80,13 @@ const addAmplifierOptions = (program, options) => {
...
+    program.addOption(new Option('-n, --chainName <chainName>', 'chain name').env('CHAIN').argParser((value) => value.toLowerCase()));
```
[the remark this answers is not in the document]
milapsheth: i'd rather update `--chainName` to `--chain`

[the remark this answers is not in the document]
ahramy: I'd like to update `--chainName` to `--chain` since it's reading the `CHAIN` environment variable, but I won't include it in this PR as it would require changes in many places across the documentation. I've created a ticket in the backlog: https://axelarnetwork.atlassian.net/browse/AXE-8403

cosmwasm/contract-state.js
ahramy: updated to `query.js`

cosmwasm/contract-state.js
ahramy: updated to use a dummy wallet

cosmwasm/contract-state.js
ahramy: updated

cosmwasm/contract-state.js
ahramy: yeah, it would be great if the pattern is enforced. updated to use `async function`

cosmwasm/contract-state.js
ahramy: updated to print both reward pools. and no need to take chain name, it's being passed as `-n` paramter

cosmwasm/contract-state.js
ahramy: updated

cosmwasm/contract-state.js
ahramy: yeah, after I update the code, no need to define `CONTRACT_MAP`, but we definitely need to do this with typescript.

cosmwasm/query.js
```diff
@@ -0,0 +1,64 @@
...
+};
+
+const programHandler = () => {
+    const program = new Command();
+
+    program.name('query').description('Query contract state');
+
+    const rewardCmd = program
+        .command('rewards')
+        .description('Query rewards state')
```
isi8787, line 53: If keeping both pools at the same time: "Query rewards pool state for multisig and voting_verifier contracts"

ahramy: make sense, updated.

PR #562, opened 2025-02-17
feat: add new cli option to retrieve tagged contract version or a pre-release commit hash from R2

isi8787:
[AXE-7610](https://axelarnetwork.atlassian.net/browse/AXE-7610)

[AXE-7610]: https://axelarnetwork.atlassian.net/browse/AXE-7610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #638, opened 2025-03-31
fix(stellar): upgrade lacking symmetry with deploy script

nbayindirli:
fixes bugs introduced in #562

### DIFF
--- a/releases/stellar/2025-03-GMP-v1.1.0.md
+++ b/releases/stellar/2025-03-GMP-v1.1.0.md
@@ -112,10 +112,10 @@ node stellar/deploy-contract.js upgrade AxelarOperators --version 1.1.0 --migrat
 node stellar/deploy-contract.js upgrade AxelarGasService --version 1.1.0
 ```
 
-9. Deploy the `AxelarExample` contract with dummy ITS address
+9. Deploy the `AxelarExample` contract
 - Note that the AxelarExample contract version is v1.0.0
 ```bash
-node stellar/deploy-contract.js deploy AxelarExample --version 1.0.0 --use-dummy-its-address
+node stellar/deploy-contract.js deploy AxelarExample --version 1.0.0
 ```
 
 ## Checklist
--- a/releases/stellar/2025-03-ITS-v1.1.0.md
+++ b/releases/stellar/2025-03-ITS-v1.1.0.md
@@ -82,12 +82,6 @@ NOTE: InterchainToken + TokenManager contracts will be migrated automatically
 node stellar/deploy-contract.js upgrade Upgrader --version 1.1.0
 ```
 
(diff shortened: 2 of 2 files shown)


releases/stellar/2025-03-GMP-v1.1.0.md
```diff
@@ -109,13 +109,13 @@ node stellar/deploy-contract.js upgrade AxelarOperators --version v1.1.0 --migra
...
 ```bash
-node stellar/deploy-contract.js upgrade AxelarGasService --version v1.1.0
+node stellar/deploy-contract.js upgrade AxelarGasService --version 1.1.0
 ```
 
 9. Deploy the `AxelarExample` contract with dummy ITS address
 - Note that the AxelarExample contract version is v1.0.0
 ```bash
-node stellar/deploy-contract.js deploy AxelarExample --version v1.0.0 --use-dummy-its-address
+node stellar/deploy-contract.js deploy AxelarExample --version 1.0.0 --use-dummy-its-address
```
milapsheth, line 118: dummy address isn't needed anymore now that ITS is already deployed

releases/stellar/2025-03-ITS-v1.1.0.md
```diff
@@ -79,13 +79,13 @@ NOTE: InterchainToken + TokenManager contracts will be migrated automatically
 6. Upgrade the `Upgrader` contract
 
 ```bash
-node stellar/deploy-contract.js upgrade Upgrader --version v1.1.0
+node stellar/deploy-contract.js upgrade Upgrader --version 1.1.0
 ```
 
 7. Re-Deploy the `AxelarExample` contract
```
milapsheth, line 85: doesn't need to be redeployed

milapsheth (approved):

common/cli-utils.js
```diff
@@ -65,18 +65,17 @@ const addOptionsToCommands = (program, optionMethod, options) => {
```
milapsheth, line 65: use a more descriptive PR title?

releases/stellar/2025-03-GMP-v1.1.0.md
ahramy: @milapsheth That’s true, but I’d like to keep it here with `--use-dummy-its-address` to make it explicit that we need to re-deploy the AxelarExample contract in step 7.

releases/stellar/2025-03-ITS-v1.1.0.md
ahramy: @milapsheth If we don’t re-deploy the AxelarExample contract, it will call the previous ITS, which we want to avoid. So I think it needs to be re-deployed. Do we want to keep AxelarExample calling the previous ITS contract?

PR #639, opened 2025-04-01
feat: add script for rotating verifiers

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-7631

### DIFF
--- a/releases/ampd/2025-04-03-ampd-v1.6.0.md
+++ b/releases/ampd/2025-04-03-ampd-v1.6.0.md
@@ -1,0 +1,123 @@
+
+# Ampd v1.6.0
+
+|  | **Owner** |
+|-----------|------------|
+| **Created By** | @cjcobb23 <cj@interoplabs.io> |
+| **Deployment** | TBD
+
+| **Network** | **Deployment Status** | **Date** |
+|-------------|----------------------|----------|
+| **Devnet Amplifier** | Deployed | 2025-04-03 |
+| **Stagenet** | Deployed | 2025-04-03 |
+| **Testnet** | In Progress | 2025-04-03 |
+| **Mainnet** | - | TBD |
+
+
+[Release](https://github.com/axelarnetwork/axelar-amplifier/releases/tag/ampd-v1.6.0)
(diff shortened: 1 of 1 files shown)


cosmwasm/README.md
```diff
@@ -308,3 +308,32 @@ node cosmwasm/submit-proposal.js migrate \
...
 ```
+
+### Rotating verifier set
+1. Create a .env for rotation, containing mnemonic for multisig-prover-admin and environment:
+```
+MNEMONIC="<mnemonic for multisig prover admin>"
+ENV="<environment>"
+```
+
+2. Register/Deregister verifiers for the chain in scope:
```
isi8787, line 319: these commands are never really ran from the local terminal, right? usually it requires access to the k8s cluster so can we provide some more information as to where the action should take place

cosmwasm/rotate-signers.js
```diff
@@ -0,0 +1,128 @@
...
+};
+
+const programHandler = () => {
+    const program = new Command();
+
+    program.name('rotate-signers').description('Rotate signers');
+
+    const updateVerifiersCmd = program
+        .command('update-verifier-set <chain>')
+        .description('Update Verifier set')
```
isi8787, line 109:
```suggestion
        .description('Update verifier set')
```

cosmwasm/rotate-signers.js
```diff
@@ -0,0 +1,128 @@
...
+    return tx;
+};
+
+const getNextVerifierSet = async (config, chain) => {
+    const client = await CosmWasmClient.connect(config.axelar.rpc);
+    return await client.queryContractSmart(config.axelar.contracts.MultisigProver[chain].address, 'next_verifier_set');
+};
+
+const getVerifierSetStatus = async (config, chain, verifierStatus) => {
+    const client = await CosmWasmClient.connect(config.axelar.rpc);
```
isi8787, line 24: are we able to have global client connection rather than establishing new ones each time status check happen

isi8787 (approved):

PR #651, opened 2025-04-08
feat: add amplifier tag in non-evm

blockchainguyy:
This PR adds amplifier tag in non evm chain. 
It also creates decimals & connectionType checks to avoid empty fields.

JIRA: https://axelarnetwork.atlassian.net/browse/AXE-8537

### DIFF
--- a/axelar-chains-config/info/devnet-amplifier.json
+++ b/axelar-chains-config/info/devnet-amplifier.json
@@ -537,6 +537,7 @@
       "chainId": 545,
       "rpc": "https://testnet.evm.nodes.onflow.org",
       "tokenSymbol": "FLOW",
+      "decimals": 18,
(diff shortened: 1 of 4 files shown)


axelar-chains-config/info/devnet-amplifier.json
```diff
@@ -621,7 +621,8 @@
       },
       "confirmations": 1,
       "finality": "finalized",
-      "approxFinalityWaitTime": 1
+      "approxFinalityWaitTime": 1,
+      "decimals": 18
```
milapsheth, line 625: decimals should be added right after tokenSymbol for clarity. In a separate task, we should probably create a `"token": { .. }` object with relevant details to encapsulate it better

milapsheth (approved):

PR #655, opened 2025-04-09
feat: add decimals and connectionType param to chains config

blockchainguyy:
This PR adds amplifier tag in xrpl in devnet-amplifier environment.
It adds missing decimals values in chains.
It also improves tests logic to add missing chainNames.

### DIFF
--- a/releases/ampd/2025-04-03-ampd-v1.6.0.md
+++ b/releases/ampd/2025-04-03-ampd-v1.6.0.md
@@ -38,13 +38,21 @@ cosmwasm_contract = 'axelar1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq
 type = 'MultisigSigner'
 
 [[handlers]]
-chain_name = 'xrpl'
+chain_name = 'xrpl-evm'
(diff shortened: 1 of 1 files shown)


axelar-chains-config/tests/schema/index.js
```diff
@@ -25,24 +25,28 @@ export const axelarGatewaySchema = {
...
 };
 
 export const contractSchema = {
     id: '/info.chains.contracts',
     type: 'object',
     patternProperties: {
-        // PascalName e.g. 'AxelarGasService', 'AxelarGateway' etc.
-        '(^[a-z]|[A-Z])[a-z]*': {
+        // PascalName e.g. 'AxelarGasService' etc.
+        '^(?!AxelarGateway)(^[a-z]|[A-Z])[a-zA-Z]*$': {
```
milapsheth, line 42: it's fine to only have AxelarGateway as the available contract. That is the only required protocol contract technically so no need to exclude matching for it?

milapsheth (approved):

axelar-chains-config/tests/schema/index.js
milapsheth, line 42:
```suggestion
        '^([a-z]|[A-Z])[a-zA-Z]*$': {
```

PR #658, opened 2025-04-11
chore(stellar): upgrade stellar-2025-q1 stagenet v1.1.1

nbayindirli:
Summary Log: https://github.com/axelarnetwork/axelar-contract-deployments/pull/658/files#diff-3312e15f1891e3678b0b01ca9034fcafeeaac8d9aac5a056d49fd36e8a45e8af


releases/stellar/logs/v1.1.1/2025-q1/stagenet.md
```diff
@@ -0,0 +1 @@
+TK
```
milapsheth, line 1: add logs as you execute them instead of adding at the end

milapsheth (approved):

releases/stellar/logs/v1.1.1/2025-q1/stagenet.md
ahramy: yeah, but agreed that it should be added as we execute them

PR #667, opened 2025-04-18
chore(stellar): update readme and remove generate bindings

ahramy:
https://axelarnetwork.atlassian.net/browse/AXE-9087


stellar/generate-bindings.js
```diff
@@ -1,56 +0,0 @@
-'use strict';
-
```
milapsheth, line 2: Why remove this instead of fix it? We should look into generating these TS bindings for our contracts

milapsheth (approved):

stellar/generate-bindings.js
ahramy: I’m not sure why this generate-bindings script is needed. I looked into other chains - EVM and Sui - but couldn’t find similar usage. Also, I believe the Stellar CLI supports all contract calls and invocations, so this might not be necessary. Could you share where these generated TypeScript bindings would actually be used?

And if we want to add the generate-bindings script, I think it should be part of the TypeScript conversion. cc: @nbayindirli Right now, the existing script is very outdated and completely non-functional. It throws a lot of errors and isn’t usable, so I’d rather delete it and add it back from scratch if needed.

PR #672, opened 2025-04-23
chore(xrpl): upgrade xrpl dependency

k4m4:
- A backdoor was implanted into [xrpl.js](https://www.npmjs.com/package/xrpl) versions v4.2.1-4.2.4: https://github.com/XRPLF/xrpl.js/security/advisories/GHSA-33qr-m49q-rxfx.
- Even though `package.json` uses `^4.2.0`, this repo should've been unaffected since the `package-lock.json` file fixes v4.2.0.
- This PR bumps the `xrpl` dependency to v2.5.0, which contains the patch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "csv-parser": "^3.0.0",
     "path": "^0.12.7",
     "toml": "^3.0.0",
-    "xrpl": "^4.2.0"
+    "xrpl": "^4.2.5"
   },
   "devDependencies": {
     "@ledgerhq/hw-transport-node-hid": "^6.27.21",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "csv-parser": "^3.0.0",
     "path": "^0.12.7",
     "toml": "^3.0.0",
-    "xrpl": "^4.2.0"
+    "xrpl": "^4.2.5"
```
milapsheth, line 40: should remove `^` and pin the version so future compromised patches don't get included unintentionally

PR #685, opened 2025-04-29
chore: run TS prettier

nbayindirli:
AXE-8948

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+// TODO: Delete this file
+**/*
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -12,38 +12,38 @@ const { networks, etherscan } = importNetworks(chains, keys);
 networks.hardhat.hardfork = process.env.EVM_VERSION || 'merge';
 
 const config = {
-    solidity: {
-        version: '0.8.21',
-        settings: {
-            evmVersion: process.env.EVM_VERSION || 'london',
-            optimizer: {
-                enabled: true,
-                runs: 1000000,
-                details: {
-                    peephole: process.env.COVERAGE === undefined,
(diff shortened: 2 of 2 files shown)


.github/workflows/lint.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
         uses: useblacksmith/setup-node@v5
         with:
           node-version: 18.x
-          cache: 'npm'
+          cache: "npm"
```
milapsheth, line 17: will be inconsistent with gh actions in other repos. can we use single quotes in gh actions? double quotes allow interpolation too so we can use only when necessary?

common/cli-utils.js
```diff
@@ -1,96 +1,119 @@
-'use strict';
+"use strict";
```
milapsheth, line 1: As per chat gpt
>Single quotes '...' are generally preferred in most modern JavaScript/TypeScript style guides

common/cli-utils.js
```diff
@@ -1,96 +1,119 @@
...
 
 const addEnvOption = (program, defaultValue) => {
-    program.addOption(
-        new Option('-e, --env <env>', 'environment')
-            .choices(CHAIN_ENVIRONMENTS)
-            .default(defaultValue || 'testnet')
-            .makeOptionMandatory(true)
-            .env('ENV'),
-    );
+  program.addOption(
```
milapsheth, line 18: can we change indentation later? it's going to create a conflict for all the open PRs. we should merge some PRs first and then migrate if we want

.prettierrc.cjs
```diff
@@ -0,0 +1,3 @@
+// .prettierrc.cjs
```
milapsheth, line 1: unnecessary?

milapsheth (approved):

PR #702, opened 2025-05-08
refactor(stellar): move utils to TS

nbayindirli:
AXE-8848


stellar/utils.ts
```diff
@@ -193,14 +204,13 @@ async function broadcast(operation, wallet, chain, action, options = {}, simulat
 
     if (simulateTransaction) {
```
milapsheth, line 205: This feels out of place (the broadcast method looks like a bunch of independent logic being run depending on the action type, another example is the native payment). Create a separate simulate/query function?

stellar/utils.ts
```diff
@@ -193,14 +204,13 @@ async function broadcast(operation, wallet, chain, action, options = {}, simulat
...
         const tx = await buildTransaction(operation, server, wallet, chain.networkType, options);
-        const response = await server.simulateTransaction(tx);
-
-        if (response.error) {
-            throw new Error(response.error);
+        try {
+            const response = await server.simulateTransaction(tx);
+            printInfo('successfully simulated tx', `action: ${action}, networkType: ${chain.networkType}, chainName: ${chain.name}`);
```
milapsheth, line 209: Let the caller print if they wish instead? Verbose logging from a helper by default creates noise

stellar/utils.ts
```diff
@@ -39,6 +38,18 @@ const SUPPORTED_CONTRACTS = new Set([
     'Multicall',
 ]);
 
+type NetworkType = 'local' | 'futurenet' | 'testnet' | 'mainnet';
+
+interface Options {
```
milapsheth, line 43: Use a different name maybe? These options are different from the CLI provided options so a bit confusing

milapsheth (approved):

stellar/utils.ts
```diff
@@ -1,22 +1,21 @@
 'use strict';
 
-const {
+import {
```
milapsheth, line 3: Create a separate Jira ticket instead of reusing a different one?

PR #703, opened 2025-05-08
feat: add default values for wasm proposal

blockchainguyy:


axelar-chains-config/info/devnet-amplifier.json
```diff
@@ -1494,6 +1494,8 @@
     "grpc": "devnet-amplifier.axelar.dev:9090",
     "tokenSymbol": "AMPLIFIER",
     "gasPrice": "0.00005uamplifier",
-    "gasLimit": "auto"
+    "gasLimit": "auto",
```
milapsheth, line 1497: Add a nativeDenom field as well perhaps

axelar-chains-config/info/devnet-amplifier.json
```diff
@@ -1494,6 +1494,8 @@
...
+    "depositValue": "100000000",
```
milapsheth, line 1498:
```suggestion
    "govProposalDepositAmount": "100000000",
```
Use a more meaningful name

milapsheth: Update submit proposal script to read this value and make --deposit optional

axelar-chains-config/info/stagenet.json
```diff
@@ -2410,6 +2410,8 @@
     "tokenSymbol": "AXL",
     "axelarscanApi": "https://stagenet.api.axelarscan.io",
     "gasPrice": "0.007uaxl",
-    "gasLimit": "auto"
+    "gasLimit": "auto",
+    "depositValue": "100000000",
```
milapsheth, line 2414: Is it not the same as testnet? We can make it the same since stagenet should match testnet

axelar-chains-config/info/devnet-amplifier.json
```diff
@@ -1494,6 +1494,8 @@
...
+    "rewardAmount": "1000000uamplifier"
```
milapsheth, line 1499: Is this EVM specific? Don't think we should use a static value since it depends on the chain and AXL price?

milapsheth: But rewards can be configured differently for different chains, it's just that we aren't doing that rn. In fact as part of the smaller verifier set plan, we're looking at setting different rewards rates on chains now

milapsheth: Also, govProposalRewardAmount isn't a great name. Deposit amount is for gov proposals. Reward amount is for chain verifiers, so maybe verifierRewardPoolPerChain

PR #705, opened 2025-05-08
feat: get all registered tokens and add scripts to register them with the ITS HUB

Foivos:


cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
...
+        return this.token()[this.chainName()];
+    }
+
+    chainName() {
+        return this.chainNames[this.chainIndex];
+    }
+
+    rpc() {
+        const chainName = this.chainName();
+        return env === 'mainnet' || env === 'testnet' ? RPCs.axelar_bridge_evm.find((chain) => chain.name.toLowerCase() === chainName).rpc_addr : info.chains[chainName].rpc;
```
milapsheth, line 52: use a consistent rpc json structure. we have multiple environments as well, and this would allow us to reuse an rpc override mechanism in other evm scripts

cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
+require('dotenv').config();
+const env = process.env.ENV;
+const mnemonic = process.env.MNEMONIC;
```
milapsheth, line 3: use our cli pattern instead of reading env vars directly

cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
...
+            this.chainIndex++;
+            if (this.chainIndex >= this.chainNames.length) {
+                if(!this.incrementTokenIndex()) return false;
+            }
+            const chainName = this.chainName()
+            if(chainName === this.token().originChain.origin_chain) continue;
+            const current = this.get();
+            if (!current.registered) {
+                if (current.track) try {
+                    const provider = getDefaultProvider(this.rpc());
```
milapsheth, line 69: not very readable, use more newlines, return early, avoid deep nested code, add some comments

evm/get-origin-chains.js
```diff
@@ -0,0 +1,74 @@
...
+
+        // if squid is tracking this token, use this value.
+        /*if (squidTokens.tokens[tokenId]) {
+            token.originChain = squidTokens.tokens[tokenId].originAxelarChainId;
+            continue;
+        }*/
+
+        // if only a single chain is untacked, use that chain
+        const untracked = [];
+        for(const chainName of Object.keys(token)) {
```
milapsheth, line 59: where are these chain names coming from? we need to use the exact case sensitive `axelarId`s for the chains

cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
...
+    const client = await prepareClient(config, wallet);
+    const supply = tokenIterator.get().supply;
+    const supplyParam = supply ? {"tracked": String(supply)} : "untracked";
+    const msg = { "register_p2p_token_instance": {
+        "chain": tokenIterator.chainName(),
+        "token_id": tokenIterator.tokenId().slice(2),
+        "origin_chain": tokenIterator.token().originChain,
+        "decimals": tokenIterator.get().decimals,
+        "supply": supplyParam,
+    } };
```
milapsheth, line 97: add a dry run mode to make it easy to verify what the cmd will do before running it for sanity

cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
...
+    
+    await client.execute(account.address, info.axelar.contracts.InterchainTokenService.address, msg, 'auto');
+    // If registration is successfull skip this token in the future without needing to query.
+    tokenIterator.get().registered = true;
+
+    saveConfig(config, env);
+}
+
+if (require.main === module) {
+    (async () => {
```
milapsheth, line 121: the script should be able to register one or more or all provided tokenIds for a provided chain based on the usual cli input

releases/evm/2025-01-ITS-v2.1.0.md
```diff
@@ -81,11 +81,26 @@ node evm/deploy-its.js --upgrade
...
+For `blocknumber` go to the explorer and find the block when the its was deployed on that chain (There is sadly no way to query ethers for this information to my knowledge). Then do the following and wait for the data to be collected.
+
+```bash
+# Get all token managers on chains.
+node evm/get-token-manager-deployments.js
+```
 
-We’d also like to migrate minter for all native ITS tokens if they weren’t auto migrated, so we can remove the migration logic in the future. We’ll create scripts to do that in the future since this is not a blocker for the upgrade.
+Once you have the data collected for the chains you wish to migrate, simply run, to migrate all the tokens.
+```bash
```
milapsheth, line 100: How do we migrate in batches? E.g. migrate all known tokens first, and then collect and migrate again right before the upgrade? does it amount to running these cmds again? Can we provide a time range or so so it doesn't spend time going over all historical data again?

cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
...
+
+    const ensureChainRegistered = async (chain) => {
+        if(registeredChains[chain]) return;
+        registeredChains[chain] = true;
+        if(await client.queryContractSmart(info.axelar.contracts.InterchainTokenService.address, {"its_chain": {"chain": chain}})) return;
+        const msg = {"register_chains":{"chains":[{"chain":chain,"its_edge_contract":"source-its-contract","truncation":{"max_uint_bits":256,"max_decimals_when_truncating":255}}]}};
+        await client.execute(account.address, info.axelar.contracts.InterchainTokenService.address, msg, 'auto');
+    }
+    await ensureChainRegistered(tokenIterator.chainName());
+    await ensureChainRegistered(tokenIterator.token().originChain);
```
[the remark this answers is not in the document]
Foivos: I was testing on a new deployment on devnet where chains were not registered so that needed to be added.

cosmwasm/register-its-token.js
Foivos: they are just provided as a command argument now, or they can be added to the `.env`

cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
...
+    
+    await client.execute(account.address, info.axelar.contracts.InterchainTokenService.address, msg, 'auto');
+    // If registration is successfull skip this token in the future without needing to query.
+    tokenIterator.get().registered = true;
+
+    saveConfig(config, env);
+}
+
+if (require.main === module) {
```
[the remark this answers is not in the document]
Foivos: done

cosmwasm/register-its-token.js
```diff
@@ -0,0 +1,128 @@
...
+
+    rpc() {
+        const chainName = this.chainName();
+        return env === 'mainnet' || env === 'testnet' ? RPCs.axelar_bridge_evm.find((chain) => chain.name.toLowerCase() === chainName).rpc_addr : info.chains[chainName].rpc;
+    }
+
+    async getNext() {
+        const previous = this.get();
+        if (previous && previous.supply) delete previous.supply;
+        fs.writeFileSync(`./axelar-chains-config/info/tokens-${env}.json`, JSON.stringify(tokenInfo, null, 2));
```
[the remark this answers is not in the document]
Foivos: since each token id has multiple tokens (because of multiple chains) users can provide a list of token ids, and a list of chains, and their intersections will be registered. If nothing is provided then everything is migrated.

cosmwasm/register-its-token.js
Foivos: done, check it out and lmk if you find it ok.

evm/get-origin-chains.js
Foivos: these are the chain keys in the info json, converting to axelarId for the cosmwasm msg.

cosmwasm/register-its-token.js
Foivos: done

PR #707, opened 2025-05-08
feat: add default values for wasm proposal

Foivos:
**This PR adds depositValue & rewardAmount in all environments for wasm proposals**

JIRA: https://axelarnetwork.atlassian.net/browse/AXE-9304


cosmwasm/cli-utils.js
```diff
@@ -171,9 +170,9 @@ const addMigrateOptions = (program) => {
 };
 
 const addProposalOptions = (program) => {
-    program.addOption(new Option('-t, --title <title>', 'title of proposal').makeOptionMandatory(true));
```
isi8787, line 174: why are we removing the mandatory tag, seems like it can cause issue for mainnet if its left off

isi8787:
Need to update README with the changes

isi8787 (approved):

PR #735, opened 2025-05-10
refactor: use ts-node instead of node

nbayindirli:
AXE-9315

### DIFF
--- a/.github/workflows/test-evm.yaml
+++ b/.github/workflows/test-evm.yaml
@@ -22,12 +22,6 @@ jobs:
       - name: Install
         run: npm ci
 
-      - name: Debug node_modules
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test-evm.yaml
```diff
@@ -16,9 +16,18 @@ jobs:
...
+        run: echo "$(pwd)/node_modules/.bin" >> $GITHUB_PATH
+
       - name: Install
         run: npm ci
 
+      - name: Debug node_modules
+        run: |
+          ls -la node_modules/.bin/
+          which ts-node || echo "ts-node not found in PATH"
+          npm list ts-node
```
milapsheth, line 29: debug step isn't needed

milapsheth (approved):

PR #746, opened 2025-05-13
feat: add pre-commit for private RPC & EOL check

blockchainguyy:
**This PR add pre-commit to prevent private RPC from being exposed.**

JIRA: https://axelarnetwork.atlassian.net/browse/AXE-9370


.husky/scripts/detect-private-rpc.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+SENSITIVE_RPC_PATTERNS=(
```
isi8787, line 3: also check for infura , and we might be adding thirdweb also

PR #752, opened 2025-05-14
feat(stellar): always simulate read-only transactions

nbayindirli:
AXE-9360


stellar/utils.ts
```diff
@@ -160,7 +166,7 @@ async function sendTransaction(tx, server, action, options: Options = {}) {
         }
 
         if (getResponse.status !== 'SUCCESS') {
-            throw Error(`Transaction failed: ${getResponse.resultXdr}`);
+            throw Error(`Transaction failed: ${getResponse.txHash}`);
```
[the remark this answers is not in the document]
milapsheth: i think the goal was to display the error from the failed tx? Can we show that along with the tx hash?

stellar/utils.ts
```diff
@@ -176,7 +182,7 @@ async function sendTransaction(tx, server, action, options: Options = {}) {
         const returnValue = transactionMeta.v3().sorobanMeta().returnValue();
 
         if (options && options.verbose) {
-            printInfo('Transaction result', returnValue.value());
+            printInfo('Transaction result', JSON.stringify(returnValue.value()));
```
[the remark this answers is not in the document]
milapsheth: IIRC `printInfo` stringifies it if the argument is an object?

stellar/utils.ts
```diff
@@ -186,34 +192,56 @@ async function sendTransaction(tx, server, action, options: Options = {}) {
...
+    options: Options,
+    simulateTransaction?: boolean,
+): Promise<xdr.ScVal | rpc.Api.SimulateTransactionSuccessResponse> {
+    /* Allows user to require or disallow simulation. Will always attempt if not specified. */
+    if (simulateTransaction === undefined) simulateTransaction = true;
+
     const server = new rpc.Server(chain.rpc, { allowHttp: chain.networkType === 'local' });
+    const tx = await buildTransaction(operation, server, wallet, chain.networkType, options);
+
+    if (simulateTransaction) {
```
milapsheth, line 209: Create a separate method `simulate` instead of making the `broadcast` function more complicated

stellar/utils.ts
```diff
@@ -186,34 +192,56 @@ async function sendTransaction(tx, server, action, options: Options = {}) {
...
+        const response = await server.simulateTransaction(tx);
+        if (isReadOnly(response)) {
```
milapsheth, line 211: why the check? Can't we simulate write txs too if we wanted? maybe we're interested in knowing the return value or gas costs etc.?

stellar/utils.ts
```diff
@@ -186,34 +192,56 @@ async function sendTransaction(tx, server, action, options: Options = {}) {
...
 
     if (options && options.nativePayment) {
-        const tx = await buildTransaction(operation, server, wallet, chain.networkType, options);
         tx.sign(wallet);
         return sendTransaction(tx, server, action, options);
     }
+
     if (options && options.estimateCost) {
-        const tx = await buildTransaction(operation, server, wallet, chain.networkType, options);
         const resourceCost = await estimateCost(tx, server);
```
milapsheth, line 228: doesn't estimateCost also simulate the tx? merge the implementations to reduce duplication?

milapsheth: gh action is failing

stellar/utils.ts
```diff
@@ -186,34 +192,54 @@ async function sendTransaction(tx, server, action, options: Options = {}) {
...
-    }
+    const preparedTx = await prepareTransaction(operation, server, wallet, chain.networkType, options);
+    return sendTransaction(preparedTx, server, action, options);
+}
+
+function isReadOnly(response: rpc.Api.SimulateTransactionResponse): boolean {
+    if (!rpc.Api.isSimulationSuccess(response)) return false;
+    if (response.transactionData.getReadWrite().length > 0) return false;
+    if (response.result && response.result.auth.length > 0) return false;
+    if (response.stateChanges && response.stateChanges.length > 0) return false;
```
ahramy, line 239: May be try to optimize it something like this?

```suggestion
    if (
        !rpc.Api.isSimulationSuccess(response) ||
        response.transactionData.getReadWrite().length > 0 ||
        response.result?.auth?.length > 0 ||
        response.stateChanges?.length > 0
    ) {
        return false;
    }
```

stellar/utils.ts
ahramy: yeah, the gh action is failing because it was unable to set trusted chain `remote`

PR #757, opened 2025-05-16
fix(cosmwasm): add router and multisig to coordinator instantiation script

SDavidson1177:


cosmwasm/utils.js
```diff
@@ -215,7 +217,7 @@ const makeCoordinatorInstantiateMsg = (config, _options, contractConfig) => {
         throw new Error('Missing or invalid ServiceRegistry.address in axelar info');
     }
 
-    return { governance_address: governanceAddress, service_registry: registryAddress };
+    return { governance_address: governanceAddress, service_registry: registryAddress, router_address: routerAddress, multisig_address: multisigAddress };
```
ahramy, line 220: Nit: Since it's returning in json format, the order of the returned values shouldn't matter, but could you just double-check that it aligns with the order when initiating the coordinator for code readability?

ahramy (approved):

cosmwasm/utils.js
SDavidson1177: Yes. I just double checked, and the order is the same

PR #812, opened 2025-06-10
refactor: remove hardcoded artifact path

andrew-interop:
[AXE-8490](https://axelarnetwork.atlassian.net/browse/AXE-8490)

- Removed hardcoded TM/IT logic. This was an issue when deploy ITS was called using --artifact-path, due to ITS calling IT and TM. Now, getContractCodePath will create an artifact path using the contractName regardless of the contract.

Tested as such: followed stellar/README.md until the deployment step, then ran `ts-node stellar/deploy-contract.js deploy {contractName} --artifact-path ../axelar-amplifier-stellar/target/wasm32-unknown-unknown/release/stellar_{artifactName}.optimized.wasm` for:
- contractName: AxelarGateway, artifactName: axelar_gateway
- contractName: InterchainTokenService, artifactName: interchain_token_services
- contractName: AxelarOperators, artifactName: axelar_operators
- contractName: AxelarGasService, artifactName: axelar_gas_services
- contractName: Upgrader, artifactName: upgrader
- contractName: AxelarExample, artifactName: axelar_example

No tests were ran for upgrade or upload, as these functions are ran on individual contract names, and should not run into this issue.

`axelar-contract-deployments` and `axelar-amplifier-stellar` were in the same directory.

[AXE-8490]: https://axelarnetwork.atlassian.net/browse/AXE-8490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stellar/cli-utils.js
```diff
@@ -1,3 +1,33 @@
 'use strict';
 
 require('../common/cli-utils');
+
+const { Option } = require('commander');
+
+const addStoreOptions = (program) => {
```
isi8787, line 7: We should not be moving away from the common model. It is better to make sure this stays in the common/cli-utils and make the changes interchangeable with the rest of the codebase to unify rather than make more custom for each chain.

andrew-interop: Would it make more sense to:
1. Allow `--artifact-path` and `--artifact-dir`, and fail when CosmWasm calls `--artifact-dir` or fail when Stellar calls `--artifact-path`, or 
2. Integrate CosmWasm to work with `--artifact-dir` and eliminate `--artifact-path` completely?
Alternatively, I could do 1. and let 2. be a future Jira task?

isi8787: We can do 2, as long as this is an immediate follow to this PR. If i remember it sohuld be an easier task since that is the approach already taken in cosmowasm contracts . Strive to unify and keep codebase consistent

cosmwasm/utils.js
```diff
@@ -1101,17 +1101,23 @@ const getContractR2Url = (contractName, contractVersion) => {
     throw new Error(`Invalid contractVersion format: ${contractVersion}. Must be a semantic version (including prefix v) or a commit hash`);
 };
 
+function getContractArtifactPath(artifactDir, contractName) {
```
isi8787, line 1104: can we continue to use arrow notation

stellar/utils.ts
```diff
@@ -451,38 +451,23 @@ const getContractR2Url = (contractName, version) => {
...
 
-function getContractArtifactPath(artifactPath, contractName) {
-    const basePath = artifactPath.slice(0, artifactPath.lastIndexOf('/') + 1);
+function getContractArtifactPath(artifactDir, contractName) {
+    const basePath = artifactDir.endsWith('/') ? artifactDir : artifactDir + '/';
     const fileName = `stellar_${pascalToKebab(contractName).replace(/-/g, '_')}.optimized.wasm`;
     return basePath + fileName;
 }
 
 const getContractCodePath = async (options, contractName) => {
```
isi8787, line 460: i think with your changes this function is also now the same between stellar and evm so can be moved to common utils

isi8787 (approved):

PR #823, opened 2025-06-16
docs(release): stellar-2025-q2 release doc

ahramy:


releases/stellar/2025-06-GMP-v1.1.2.md
```diff
@@ -0,0 +1,185 @@
...
+| **Testnet**          | `GDOOQUMPCFOSHEYTEXVSZLDIBL4GOAT6RQMEHHYSRXE2RYML67RA4NPF` |
+
+```bash
+ts-node stellar/deploy-contract.js deploy AxelarOperators --version 1.1.2
+```
+
+5. Deploy Gas Service
+
+```bash
+ts-node stellar/deploy-contract.js deploy AxelarGasService --version 1.1.2
```
milapsheth, line 100: why use v1.1.2 for all contracts? The deployment should replicate what exists on mainnet, so the next release follows the same upgrade path. See the last release to see what versions of contract we deployed

milapsheth: https://github.com/axelarnetwork/axelar-contract-deployments/blob/feat/add-plume-mainnet/releases/stellar/2025-03-GMP-v1.1.1.md#L97

releases/cosmwasm/2025-06-Stellar-GMP-v1.1.2.md
```diff
@@ -0,0 +1,325 @@
+# Stellar GMP Amplifier
+
```
milapsheth, line 2: is this any different than the last one? if not, we could just point to the older one from the Stellar specific docs to avoid duplication

releases/stellar/2025-06-GMP-v1.1.2.md
ahramy: As discussed in Slack, we will keep Stagenet as is since it's only a patch bump. It will be reset next quarter anyway

ahramy: As discussed in Slack, we will keep Stagenet as is since it's only a patch bump. It will be reset next quarter anyway

releases/cosmwasm/2025-06-Stellar-GMP-v1.1.2.md
ahramy: It has not been changed. totally make sense, updated.

PR #847, opened 2025-07-01
feat(evm): add hyperliquid core util

isi8787:


evm/README.md
```diff
@@ -76,6 +76,27 @@ ts-node evm/deploy-its -e testnet -n ethereum -s '[salt]' --proxySalt 'v1.0.0' -
...
+For Hyperliquid chains, you can manually switch between BIG and SMALL blocks using the `hyperliquid-block-helper.js` script. This is particularly useful for deploying large contracts like the Gateway contract. For the ITS deployment the module is importer directly into `deploy-its.js` and block changes are handled internally.
+
+To switch block sizes the CHAINS and PRIVATE_KEY should be set in the .env:
+
+```bash
+# Switch to BIG blocks
+node evm/hyperliquid-block-helper.js --blockSize big
+
+# Switch to SMALL blocks  
+node evm/hyperliquid-block-helper.js --blockSize small
```
milapsheth, line 90: Rename this to `hyperliquid.js` since it can be a generic hyperliquid util
```suggestion
node evm/hyperliquid.js block-size big

node evm/hyperliquid.js block-size small
```

milapsheth: updating deployer on ITS token can be added to this file as well

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+    const curlCommand = `curl -s -X POST "${endpoint}" \
+        -H "Content-Type: application/json" \
+        -H "User-Agent: Mozilla/5.0 (compatible; Hyperliquid-Block-Helper/1.0)" \
+        -d '${JSON.stringify(payload)}' \
+        --connect-timeout 15 \
+        --max-time 30`;
+
+    const { stdout, stderr } = await execAsync(curlCommand);
+
+    if (stderr && !stderr.includes('curl')) {
```
milapsheth, line 85: avoid calling cli methods as that's more brittle wrt the environment. Use the `httpPost` helper util we have

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+        if (error.message.includes('does not exist')) {
+            printInfo('Result', 'Account not found, continuing without block size switch');
+            return { status: 'ok', message: 'Account not found, continuing without block size switch' };
+        }
+        if (error.message.includes('curl request failed')) {
+            printInfo('Result', 'Block size switch failed, continuing with deployment');
+            return { status: 'ok', message: 'Block size switch failed, continuing with deployment' };
+        }
+        throw error;
+    }
```
milapsheth, line 159: parse the API response in the callee and post info there vs parsing thrown exceptions

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+    printInfo('Block size', blockSize.toUpperCase());
+    printInfo('Network', network);
+
+    try {
+        const result = await switchBlockSize(privateKey, useBig, network);
+        printInfo('Result', result);
+        return result;
+    } catch (error) {
+        if (error.message.includes('does not exist')) {
+            printInfo('Result', 'Account not found, continuing without block size switch');
```
milapsheth, line 151: we have a `printError` helper

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+const { printInfo, printError, validateParameters, mainProcessor } = require('./utils');
+const { addEvmOptions } = require('./cli-utils');
+const execAsync = promisify(exec);
+
+function addressToBytes(address) {
+    return Buffer.from(address.replace('0x', ''), 'hex');
+}
+
+function actionHash(action, vaultAddress, nonce) {
+    const msgpack = require('msgpack-lite');
```
milapsheth, line 16: move imports to the top of the file

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+    const { privateKey, blockSize } = options;
+
+    validateParameters({
+        isNonEmptyString: { privateKey },
+        isNonEmptyString: { blockSize },
+    });
+
+    if (blockSize !== 'big' && blockSize !== 'small') {
+        throw new Error('Block size must be "big" or "small"');
+    }
```
milapsheth, line 129: should be handled by CLI parsing definition. commander allows `.choices` to for options. see existing usage

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+    if (!isHyperliquidChain) {
+        throw new Error(`Chain "${chain.name}" is not supported. This script only works on Hyperliquid chains.`);
+    }
+
+    const isMainnet = options.env === 'mainnet';
+    const network = isMainnet ? 'mainnet' : 'testnet';
+    const useBig = blockSize === 'big';
+
+    printInfo('Block size', blockSize.toUpperCase());
+    printInfo('Network', network);
```
milapsheth, line 143: env is already printed

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+
+    // Check if this is a Hyperliquid chain
+    const isHyperliquidChain = chain.name.toLowerCase().includes('hyperliquid') || chain.axelarId.toLowerCase().includes('hyperliquid');
+
+    if (!isHyperliquidChain) {
+        throw new Error(`Chain "${chain.name}" is not supported. This script only works on Hyperliquid chains.`);
+    }
+
+    const isMainnet = options.env === 'mainnet';
+    const network = isMainnet ? 'mainnet' : 'testnet';
```
milapsheth, line 139: use `networkType` config field instead of doing this

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+        isNonEmptyString: { privateKey },
+        isNonEmptyString: { blockSize },
+    });
+
+    if (blockSize !== 'big' && blockSize !== 'small') {
+        throw new Error('Block size must be "big" or "small"');
+    }
+
+    // Check if this is a Hyperliquid chain
+    const isHyperliquidChain = chain.name.toLowerCase().includes('hyperliquid') || chain.axelarId.toLowerCase().includes('hyperliquid');
```
milapsheth, line 132:
```suggestion
    const isHyperliquidChain = chain.axelarId.toLowerCase().includes('hyperliquid');
```
only check axelarId. Ideally, we use `chainType` for this, but we also rely on that being set to `evm` for other scripts

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+
+    const isMainnet = options.env === 'mainnet';
+    const network = isMainnet ? 'mainnet' : 'testnet';
+    const useBig = blockSize === 'big';
+
+    printInfo('Block size', blockSize.toUpperCase());
+    printInfo('Network', network);
+
+    try {
+        const result = await switchBlockSize(privateKey, useBig, network);
```
milapsheth, line 146:
```suggestion
        const result = await updateBlockSize(privateKey, useBig, network);
```
doesn't need to be a switch, might be setting the same value again

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+    validateParameters({ isNonEmptyString: { privateKey } });
+
+    if (!privateKey.startsWith('0x')) {
+        privateKey = '0x' + privateKey;
+    }
+
+    if (privateKey.length !== 66) {
+        throw new Error(`Invalid private key length: ${privateKey.length}`);
+    }
+
```
milapsheth, line 103: unnecessary validation since it was already done before
```suggestion
```

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+    if (stderr && !stderr.includes('curl')) {
+        throw new Error(`curl stderr: ${stderr}`);
+    }
+
+    const result = JSON.parse(stdout);
+    return result;
+}
+
+async function switchBlockSize(privateKey, useBig, network = 'mainnet') {
+    validateParameters({ isNonEmptyString: { privateKey } });
```
milapsheth, line 94:
```suggestion
    validateParameters({ isEVMPrivateKey: { privateKey } });
```
add a new validator instead?

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+
+    if (!privateKey.startsWith('0x')) {
+        privateKey = '0x' + privateKey;
```
milapsheth, line 97: this shouldn't be needed. Other scripts don't run into this issue

evm/hyperliquid-block-helper.js
milapsheth: or just print the api response error

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+        connectionId: hash,
+    };
+}
+
+async function signL1Action(wallet, action, activePool, nonce, isMainnet) {
+    const hash = actionHash(action, activePool, nonce);
+    const phantomAgent = constructPhantomAgent(hash, isMainnet);
+
+    const domain = {
+        chainId: 1337,
```
milapsheth, line 46: read the chain id from the config? Or create a global constant if it doesn't match the EVM chain id

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+    let vaultBuffer;
+    if (vaultAddress === null || vaultAddress === undefined) {
+        vaultBuffer = Buffer.from([0x00]);
+    } else {
+        const addressBytes = addressToBytes(vaultAddress);
+        vaultBuffer = Buffer.concat([Buffer.from([0x01]), addressBytes]);
+    }
+
+    const data = Buffer.concat([actionData, nonceBuffer, vaultBuffer]);
+    const { keccak256 } = require('ethers/lib/utils');
```
milapsheth, line 30: same, imports should be global

evm/hyperliquid-block-helper.js
```diff
@@ -0,0 +1,182 @@
...
+const { Command, Option } = require('commander');
+const { printInfo, printError, validateParameters, mainProcessor } = require('./utils');
+const { addEvmOptions } = require('./cli-utils');
+const execAsync = promisify(exec);
+
+function addressToBytes(address) {
+    return Buffer.from(address.replace('0x', ''), 'hex');
+}
+
+function actionHash(action, vaultAddress, nonce) {
```
milapsheth, line 15:
```suggestion
function actionHash(action, activePool, nonce) {
```
make the names match across the functions?

evm/interchainTokenFactory.js
```diff
@@ -292,6 +292,85 @@ async function processCommand(config, chain, options) {
             break;
         }
 
+        case 'getCurrentDeployer': {
```
milapsheth, line 295:
```suggestion
        case 'deployer': {
```
avoid get prefix

evm/interchainTokenFactory.js
```diff
@@ -292,6 +292,85 @@ async function processCommand(config, chain, options) {
...
+
+            try {
+                const HyperliquidInterchainToken = getContractJSON('HyperliquidInterchainToken');
+                const hyperliquidToken = new Contract(tokenAddress, HyperliquidInterchainToken.abi, wallet);
+
+                const currentDeployer = await hyperliquidToken.deployer();
+                printInfo('Current deployer', currentDeployer);
+            } catch (error) {
+                if (error.message.includes('deployer is not a function') || error.message.includes('execution reverted')) {
+                    const factoryDeployer = await interchainTokenFactory.getTokenDeployer(tokenId);
```
milapsheth, line 311: why would this be called from ITS factory? Call `deployer` on the token directly? I think we can move query and deployer update cli cmds to `hyperliquid.js`

evm/interchainTokenFactory.js
```diff
@@ -292,6 +292,85 @@ async function processCommand(config, chain, options) {
...
+            printInfo('New deployer', deployer);
+
+            const serviceOwner = await hyperliquidService.owner();
+            const isOperator = await hyperliquidService.isOperator(wallet.address);
+
+            if (wallet.address.toLowerCase() !== serviceOwner.toLowerCase() && !isOperator) {
+                throw new Error('Wallet does not have permission to update deployers. Must be service owner or operator.');
+            }
+
+            const tx = await hyperliquidService.updateTokenDeployer(tokenId, deployer, gasOptions);
```
milapsheth, line 365: ITS cmds should be in `its.js` not factory script

milapsheth: But in this case, I think we can move it to `hyperliquid.js`

axelar-chains-config/info/devnet-amplifier.json
```diff
@@ -1258,6 +1258,98 @@
           "address": "0x9186233C327329408a9034F22407244421cADbEF"
```
milapsheth, line 1258: PR titles should be more specific and meaningful, say
`feat(evm): add ability to update hyperliquid block sizes`

evm/hyperliquid-block-helper.js
isi8787: have a problem with the testnet URL throwing SSL/TLS incompatibility. still trying to figure out

evm/interchainTokenFactory.js
isi8787: moved to hyperliquid.js

evm/interchainTokenFactory.js
isi8787: moved deployer related actions to hyperliquid.js

evm/hyperliquid-block-helper.js
isi8787: moving these into global. Im not sure i will actually use somewhere else besides the signing function. The chain information is actually for hypercore so it does not match the config file

evm/hyperliquid-block-helper.js
milapsheth: Yeah, throwing the error is preferable, let the caller decide what to do with it

evm/hyperliquid-block-helper.js
milapsheth: In that case, add a field in the Hyperliquid config for Hypercore chain id. I imagine it'll be different between testnet/mainnet so we source it from the configs

evm/hyperliquid.js
```diff
@@ -61,23 +38,29 @@ function constructPhantomAgent(hash, isMainnet) {
     };
 }
 
-async function signL1Action(wallet, action, activePool, nonce, isMainnet) {
+async function signL1Action(wallet, action, activePool, nonce, isMainnet, chain) {
     const hash = actionHash(action, activePool, nonce);
     const phantomAgent = constructPhantomAgent(hash, isMainnet);
 
-    const signature = await wallet._signTypedData(HYPERLIQUID_CONFIG.domain, { Agent: HYPERLIQUID_CONFIG.types.Agent }, phantomAgent);
+    // Use chain.hypercore.domain if available, otherwise fall back to default
```
milapsheth, line 45: there's no default fallback? Let's require it to be available in the config

evm/hyperliquid.js
```diff
@@ -61,23 +38,29 @@ function constructPhantomAgent(hash, isMainnet) {
...
     const payload = { action, signature, nonce };
-    const endpoint = isMainnet ? HYPERLIQUID_CONFIG.endpoints.mainnet : HYPERLIQUID_CONFIG.endpoints.testnet;
+    const endpoint = `${chain.hypercore.url}/exchange`;
 
     const curlCommand = `curl -s -X POST "${endpoint}" \
         -H "Content-Type: application/json" \
-        -H "User-Agent: ${HYPERLIQUID_CONFIG.userAgent}" \
         -d '${JSON.stringify(payload)}' \
         --connect-timeout 15 \
         --max-time 30`;
```
milapsheth, line 66: use `execSync` instead of `await execAsync` maybe?

milapsheth: throw an error if stderr is non empty for any reason and not just for curl (or more reliable, check the exit code?)

evm/hyperliquid.js
```diff
@@ -92,28 +75,22 @@ async function sendRequest(action, signature, nonce, isMainnet) {
...
-            return { success: false, error: result.response || result };
-        } else {
-            printError('API Response Error', result.response || result);
-            throw new Error(`Block size switch failed: ${result.response || result}`);
-        }
-    }
+    const signature = await signL1Action(wallet, action, null, nonce, isMainnet, chain);
+    const result = await sendRequest(action, signature, nonce, chain);
+
+    return result.status === 'ok'
```
milapsheth, line 89: avoid use of ternary operator to perform side effects

evm/hyperliquid.js
```diff
@@ -92,28 +75,22 @@ async function sendRequest(action, signature, nonce, isMainnet) {
...
+        ? { success: true, data: result }
+        : (() => {
+              throw new Error(result.response || result);
+          })();
```
milapsheth, line 93:
```suggestion
    if (result.status !== 'ok') {
        throw new Error(result.response || result);
    }

    return result;
```
use the fail-early/return early on negative cases pattern so the happy path is always easy to read

milapsheth: and no need to return `success` in an object. failure would cause an exception now

evm/hyperliquid.js
```diff
@@ -147,7 +126,7 @@ async function processCommand(config, chain, options) {
             printInfo('Network', network);
 
             try {
-                const result = await updateBlockSize(privateKey, useBig, network);
+                const result = await updateBlockSize(privateKey, useBig, network, chain);
                 if (result.success) {
                     printInfo('Result', result.data);
                     return result.data;
```
milapsheth, line 132: clean up calling code since the helper has changed

evm/hyperliquid.js
```diff
@@ -334,22 +305,43 @@ async function main(options) {
 if (require.main === module) {
     const program = new Command();
 
-    program.name('HyperliquidBlockHelper').description('Script to manage Hyperliquid specific actions');
+    program.name('hyperliquid').description('Hyperliquid chain management commands');
+
+    // Update block size command
+    const updateBlockSizeCmd = program
+        .command('update-block-size')
```
milapsheth, line 312:
```suggestion
        .command('update-block-size <block-size>')
```
see the newer `stellar/its.js` for CLI parsing pattern

evm/hyperliquid.js
```diff
@@ -334,22 +305,43 @@ async function main(options) {
...
+        .description('Update Hyperliquid block size')
+        .addOption(new Option('--block-size <blockSize>', 'block size to switch to').choices(['big', 'small']).makeOptionMandatory(true));
+
+    addEvmOptions(updateBlockSizeCmd, { privateKey: true });
+
+    updateBlockSizeCmd.action((options) => {
+        options.action = 'updateBlockSize';
```
milapsheth, line 319: see `stellar/its.js` for how to do this cleanly without needing such workarounds

evm/hyperliquid.js
```diff
@@ -334,22 +305,43 @@ async function main(options) {
...
-            .makeOptionMandatory(true),
-    );
-    program.addOption(
-        new Option('--blockSize <blockSize>', 'block size to switch to (required for updateBlockSize action)').choices(['big', 'small']),
-    );
-    program.addOption(new Option('--tokenId <tokenId>', 'ID of the token'));
-    program.addOption(new Option('--deployer <deployer>', 'deployer address'));
+    addEvmOptions(updateTokenDeployerCmd, { privateKey: true });
+    updateTokenDeployerCmd.addOption(new Option('--tokenId <tokenId>', 'ID of the token').makeOptionMandatory(true));
+    updateTokenDeployerCmd.addOption(new Option('--deployer <deployer>', 'new deployer address').makeOptionMandatory(true));
```
milapsheth, line 341: these are required args, so make them positional as well

axelar-chains-config/info/stagenet.json
```diff
@@ -104,117 +104,6 @@
...
         }
       },
-      "explorer": {
-        "url": "https://testnet.snowtrace.io",
-        "api": "https://api-testnet.snowtrace.io/api"
-      },
-      "finality": "finalized",
-      "approxFinalityWaitTime": 1
-    },
-    "fantom": {
```
milapsheth, line 114: why remove all these other chains?

evm/hyperliquid.js
```diff
@@ -0,0 +1,351 @@
...
+    const sig = ethers.utils.splitSignature(signature);
+
+    return { r: sig.r, s: sig.s, v: sig.v };
+}
+
+async function sendRequest(action, signature, nonce, chain) {
+    const payload = { action, signature, nonce };
+    const endpoint = `${chain.hypercore.url}/exchange`;
+
+    const curlCommand = `curl -s -X POST "${endpoint}" \
```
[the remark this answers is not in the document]
isi8787: for some reason i have been stuck on this. THe hyperliquid testnet rpc throws SSL error that i have not been able to figure out using standard libraries

evm/deploy-its.js
```diff
@@ -202,81 +202,39 @@ async function deployAll(config, wallet, chain, options) {
             },
```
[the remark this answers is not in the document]
isi8787: PR 828 was suppose to be the original to just test the Hyperliquid ITS Contract changes, this was a follow up PR to have dedicated hyperliquid utility and update to release docs

evm/hyperliquid.js
```diff
@@ -241,23 +130,20 @@ async function getTokenDeployer(config, chain, options, wallet) {
             if (factoryDeployer !== AddressZero) {
```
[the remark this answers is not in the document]
isi8787: removing most of this since it was just to help debug during development

evm/hyperliquid.js
```diff
@@ -241,23 +130,20 @@ async function getTokenDeployer(config, chain, options, wallet) {
...
                 printInfo('Factory deployer', factoryDeployer);
             } else {
-                throw new Error('Token does not support deployer retrieval and no factory record found');
+                printError('Token does not support deployer retrieval and no factory record found');
```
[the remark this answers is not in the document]
isi8787: deleted the whole section it was used mainly as a debugger during development

evm/utils.js
```diff
@@ -1052,6 +1056,15 @@ const verifyContractByName = (env, chain, name, contract, args, options = {}) =>
 
 const isConsensusChain = (chain) => chain.contracts.AxelarGateway?.connectionType !== 'amplifier';
 
+/**
+ * Determines if a chain is a Hyperliquid chain based on its name or axelarId
```
milapsheth, line 1060:
```suggestion
 * Determines if a chain is a Hyperliquid chain based on its axelarId
```
incorrect

evm/utils.js
```diff
@@ -1052,6 +1056,15 @@ const verifyContractByName = (env, chain, name, contract, args, options = {}) =>
...
+ * @param {Object} chain - The chain object containing name and axelarId properties
+ * @returns {boolean} - Returns true if the chain is a Hyperliquid chain, false otherwise
+ */
```
milapsheth, line 1063: fluff comments, also outdated. let's remove it? The function name makes it pretty clear and easy to maintain

evm/utils.js
```diff
@@ -60,6 +60,10 @@ const IDeployer = require('@axelar-network/axelar-gmp-sdk-solidity/interfaces/ID
 const { exec } = require('child_process');
 const { verifyContract } = require(`${__dirname}/../axelar-chains-config`);
 
+const getITSHubAddress = (config) => {
```
milapsheth, line 63: avoid get prefix for getters. Also, add this to `common/utils.js` since it's not EVM specific, and we already have other generic ITS helpers there

evm/README.md
```diff
@@ -76,6 +76,20 @@ ts-node evm/deploy-its -e testnet -n ethereum -s '[salt]' --proxySalt 'v1.0.0' -
 Change the `-s SALT` to derive a new address. Production deployments use the release version, e.g. `v1.2.1`.
 `proxySalt` is used to derive the same address as a deployment on an existing chain.
 
+## Hyperliquid
+
+The Hyperliquid chain uses a dual architecture block model with fast blocks (2 seconds, 2M gas limit) and slow blocks (1 minute, 30M gas limit). The `hyperliquid.js` script provides utilities to set an account to used a specific block size, to query the deployer address of an interchain token, and to update the deployer address of an interchain token. The supported commands are:
+
+```bash
+# Update block size
+ts-node evm/hyperliquid.js update-block-size --block-size <small|big>
```
milapsheth, line 85:
```suggestion
ts-node evm/hyperliquid.js update-block-size <small|big>
```
outdated

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+        });
+
+    program
+        .command('update-token-deployer <tokenId> <deployer>')
+        .description('Update deployer address for a Hyperliquid interchain token')
+        .action((tokenId, deployer, options) => {
+            main(updateTokenDeployer, [tokenId, deployer], options);
+        });
+
+    addOptionsToCommands(program, addEvmOptions, { privateKey: true });
```
milapsheth, line 220:
```suggestion
    addOptionsToCommands(program, addEvmOptions);
```
why is this needed? It's already added by default, and not a valid option in the helper

evm/README.md
```diff
@@ -76,6 +76,20 @@ ts-node evm/deploy-its -e testnet -n ethereum -s '[salt]' --proxySalt 'v1.0.0' -
...
+
+# Get token deployer
+ts-node evm/hyperliquid.js deployer --token-id <id>
```
milapsheth, line 88: same

evm/deploy-its.js
```diff
@@ -44,12 +47,17 @@ async function deployAll(config, wallet, chain, options) {
...
 
     const provider = getDefaultProvider(chain.rpc);
-    const InterchainTokenService = getContractJSON('InterchainTokenService', artifactPath);
 
     const contractName = 'InterchainTokenService';
     const itsFactoryContractName = 'InterchainTokenFactory';
     const contracts = chain.contracts;
 
+    const interchainTokenContractName = isHyperliquidChain(chain) ? 'HyperliquidInterchainToken' : 'InterchainToken';
+    const ServiceContract = getContractJSON(
```
milapsheth, line 56:
```suggestion
    const InterchainTokenService = getContractJSON(
```
this name is fine I think to still get the core idea across

evm/deploy-its.js
```diff
@@ -168,14 +176,15 @@ async function deployAll(config, wallet, chain, options) {
                 );
             },
         },
+        // Deploy only the appropriate token implementation based on chain type
         interchainToken: {
             name: 'Interchain Token',
-            contractName: 'InterchainToken',
+            contractName: interchainTokenContractName,
```
milapsheth, line 182:
```suggestion
            contractName: 'InterchainToken',
```
should still keep the name the same, so the config doesn't use chain specific names. Otherwise, all scripts reading it need to be specialized

evm/deploy-its.js
```diff
@@ -168,14 +176,15 @@ async function deployAll(config, wallet, chain, options) {
                 );
             },
         },
+        // Deploy only the appropriate token implementation based on chain type
```
milapsheth, line 179:
```suggestion
```
add this comment before the variable that stores the contract name?

evm/deploy-its.js
```diff
@@ -337,6 +346,10 @@ async function deployAll(config, wallet, chain, options) {
             continue;
         }
 
+        if (isHyperliquidChain(chain) && shouldUseBigBlocks(key)) {
```
milapsheth, line 349: hmm, i don't like this separate configuration for when to use big blocks. Add an optional `useHyperliquidBigBlocks` field in the contract specific config above so you can just test if that is defined?

evm/deploy-its.js
```diff
@@ -349,6 +362,10 @@ async function deployAll(config, wallet, chain, options) {
             contractConfig[key] = contract.address;
         }
 
+        if (isHyperliquidChain(chain) && shouldUseBigBlocks(key)) {
+            await switchHyperliquidBlockSize(options, false, chain);
+        }
```
milapsheth, line 367: not robust to depend on a non-atomic revert after deployment. E.g. the deployment of the implementation fails in the middle. When you trigger the deployment again, the same account is still set in big block mode. Instead, use try/finally to always revert even if there's an exception on deploy?
```
try {
   const contract = await deployment.deploy();
} finally {
   switch back
}
```

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+    return {
+        source: source,
+        connectionId: hash,
+    };
+}
+
+async function signL1Action(wallet, action, activePool, nonce, chain) {
+    const hash = actionHash(action, activePool, nonce);
+    const phantomAgent = constructPhantomAgent(hash, chain.hypercore?.source);
+    const domain = chain.hypercore?.domain;
```
milapsheth, line 43: can you add all these hypercore configs to the corresponding hyperliquid testnet/mainnet PRs?

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+
+    try {
+        const TokenContract = getContractJSON('HyperliquidInterchainToken');
+        const token = new Contract(tokenAddress, TokenContract.abi, wallet);
+
+        const currentDeployer = await token.deployer();
+        printInfo('Current deployer', currentDeployer);
+    } catch (error) {
+        throw error;
+    }
```
milapsheth, line 112: wut lol. no need for a try/catch?

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+}
+
+async function deployer(wallet, chain, options) {
+    const [tokenId] = options.args;
+    validateParameters({
+        isNonEmptyString: { tokenId },
+    });
+
+    const contracts = chain.contracts;
+    const interchainTokenFactoryAddress = contracts.InterchainTokenFactory?.address;
```
milapsheth, line 90: this isn't being used?

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+    const interchainTokenServiceAddress = contracts.InterchainTokenService?.address;
+
+    validateParameters({
+        isValidAddress: { interchainTokenFactoryAddress, interchainTokenServiceAddress },
+    });
+
+    const IInterchainTokenService = getContractJSON('IInterchainTokenService');
+
+    const interchainTokenService = new Contract(interchainTokenServiceAddress, IInterchainTokenService.abi, wallet);
```
milapsheth, line 99:
```suggestion
    const interchainTokenService = new Contract(chain.contracts.InterchainTokenService?.address, IInterchainTokenService.abi, wallet);
```
remove single-use intermediate variables for brevity?

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+        if (!isHyperliquidChain(chain)) {
+            throw new Error(`Chain "${chain.name}" is not supported. This script only works on Hyperliquid chains.`);
+        }
+
+        const rpc = chain.rpc;
+        const provider = getDefaultProvider(rpc);
+        const wallet = new Wallet(options.privateKey, provider);
+
+        return processor(wallet, chain, options);
+    });
```
milapsheth, line 173:
```suggestion
    return mainProcessor(options, (config, chain, options) => {
        if (!isHyperliquidChain(chain)) {
            throw new Error(`Chain "${chain.name}" is not supported. This script only works on Hyperliquid chains.`);
        }

        const rpc = chain.rpc;
        const provider = getDefaultProvider(rpc);
        const wallet = new Wallet(options.privateKey, provider);

        return processor(wallet, chain, args, options);
    });
```
just pass `args` directly since you're using a closure anyways

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+    const tx = await service.updateTokenDeployer(tokenId, deployer, gasOptions);
+    await handleTx(tx, chain, service, 'updateTokenDeployer');
+
+    const updatedDeployer = await token.deployer();
+    printInfo('Updated deployer', updatedDeployer);
+}
+
+async function main(processor, args, options) {
+    options.args = args;
+    return mainProcessor(options, (config, chain, options) => {
```
milapsheth, line 163: nice, i like the use of the closure

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+
+    const serviceOwner = await service.owner();
+    const isOperator = await service.isOperator(wallet.address);
+
+    if (wallet.address.toLowerCase() !== serviceOwner.toLowerCase() && !isOperator) {
+        throw new Error('Wallet does not have permission to update deployers. Must be service owner or operator.');
+    }
+
+    const gasOptions = await getGasOptions(chain, options, 'HyperliquidInterchainTokenService');
+    const tx = await service.updateTokenDeployer(tokenId, deployer, gasOptions);
```
milapsheth, line 154:
```suggestion

    const tx = await service.updateTokenDeployer(tokenId, deployer, gasOptions);
```

evm/hyperliquid.js
```diff
@@ -0,0 +1,225 @@
...
+    const IInterchainTokenService = getContractJSON('IInterchainTokenService');
+    const interchainTokenService = new Contract(interchainTokenServiceAddress, IInterchainTokenService.abi, wallet);
+
+    const tokenAddress = await interchainTokenService.registeredTokenAddress(tokenId);
+    printInfo('Token address', tokenAddress);
+
+    const ServiceContract = getContractJSON('HyperliquidInterchainTokenService');
+    const service = new Contract(interchainTokenServiceAddress, ServiceContract.abi, wallet);
+
+    const TokenContract = getContractJSON('HyperliquidInterchainToken');
```
milapsheth, line 139:
```suggestion
    const HyperliquidInterchainToken = getContractJSON('HyperliquidInterchainToken');
```
the pattern is to use the contract name as-is

evm/hyperliquid.js
isi8787: removed

evm/README.md
```diff
@@ -82,13 +82,13 @@ The Hyperliquid chain uses a dual architecture block model with fast blocks (2 s
 
 ```bash
 # Update block size
-ts-node evm/hyperliquid.js update-block-size --block-size <small|big>
+ts-node evm/hyperliquid.js update-block-size <small|big>
 
 # Get token deployer
-ts-node evm/hyperliquid.js deployer --token-id <id>
+ts-node evm/hyperliquid.js deployer <id>
```
milapsheth, line 88:
```suggestion
ts-node evm/hyperliquid.js deployer <token-id>
```

evm/README.md
```diff
@@ -82,13 +82,13 @@ The Hyperliquid chain uses a dual architecture block model with fast blocks (2 s
...
 
 # Update token deployer
-ts-node evm/hyperliquid.js update-token-deployer --token-id <id> --deployer <address>
+ts-node evm/hyperliquid.js update-token-deployer <id> <address>
```
milapsheth, line 91:
```suggestion
ts-node evm/hyperliquid.js update-token-deployer <token-id> <address>
```

evm/deploy-its.js
```diff
@@ -244,6 +244,7 @@ async function deployAll(config, wallet, chain, options) {
         implementation: {
             name: 'Interchain Token Service Implementation',
             contractName: 'InterchainTokenService',
+            useHyperliquidBigBlocks: isHyperliquidChain(chain) ? true : false,
```
milapsheth, line 247:
```suggestion
            useHyperliquidBigBlocks: isHyperliquidChain(chain),
```
😄

evm/deploy-its.js
```diff
@@ -301,6 +302,7 @@ async function deployAll(config, wallet, chain, options) {
         interchainTokenFactoryImplementation: {
             name: 'Interchain Token Factory Implementation',
             contractName: 'InterchainTokenFactory',
+            useHyperliquidBigBlocks: isHyperliquidChain(chain) ? true : false,
```
milapsheth, line 305:
```suggestion
            useHyperliquidBigBlocks: isHyperliquidChain(chain),
```

evm/deploy-its.js
```diff
@@ -346,36 +348,43 @@ async function deployAll(config, wallet, chain, options) {
...
+            }
+
+            if (!(await isContract(contract.address, provider))) {
+                throw new Error(`Contract ${deployment.name} at ${contract.address} was not deployed on ${chain.name}`);
+            }
+        } catch (error) {
+            if (deployment.useHyperliquidBigBlocks) {
+                await switchHyperliquidBlockSize(options, false, chain);
+            }
+            throw error;
```
milapsheth, line 383: this is not needed. the `finally` block is always run regardless of whether there was an exception or not

evm/deploy-its.js
```diff
@@ -346,36 +348,43 @@ async function deployAll(config, wallet, chain, options) {
...
-        saveConfig(config, options.env);
-
-        if (chain.chainId !== 31337) {
-            await sleep(5000);
-        }
-
-        if (!(await isContract(contract.address, provider))) {
-            throw new Error(`Contract ${deployment.name} at ${contract.address} was not deployed on ${chain.name}`);
+        try {
+            const contract = await deployment.deploy();
```
milapsheth, line 358: only this statement needs to be in the try block as it can fail, not the post deploy cmds

evm/hyperliquid.js
```diff
@@ -112,19 +104,18 @@ async function deployer(wallet, chain, options) {
...
+async function updateTokenDeployer(wallet, chain, args, options) {
+    const [tokenId, deployer] = args;
     validateParameters({
         isNonEmptyString: { tokenId },
         isValidAddress: { deployer },
     });
 
     const contracts = chain.contracts;
-    const interchainTokenFactoryAddress = contracts.InterchainTokenFactory?.address;
     const interchainTokenServiceAddress = contracts.InterchainTokenService?.address;
```
milapsheth, line 115: remove intermediate variables, like it was done for other methods

evm/hyperliquid.js
```diff
@@ -169,26 +160,16 @@ async function main(processor, args, options) {
...
-        throw error;
-    }
-}
-
-function shouldUseBigBlocks(key) {
-    return key === 'implementation' || key === 'interchainTokenFactoryImplementation';
+    await updateBlockSize(wallet, chain, [blockType]);
 }
 
 if (require.main === module) {
```
milapsheth, line 175: Make the block size arg more restrictive by using `new Argument(...).choices(['big', 'small'])`
example [here](https://github.com/axelarnetwork/axelar-contract-deployments/pull/853/files#diff-834ca0a0151ef05d774cfd5ea8a7ecdec5e7302ee572c3a474d233242d60551dR463)

evm/utils.js
```diff
@@ -1057,7 +1057,7 @@ const verifyContractByName = (env, chain, name, contract, args, options = {}) =>
 const isConsensusChain = (chain) => chain.contracts.AxelarGateway?.connectionType !== 'amplifier';
 
 /**
- * Determines if a chain is a Hyperliquid chain based on its name or axelarId
+ * Determines if a chain is a Hyperliquid chain based on its axelarId
  * @param {Object} chain - The chain object containing name and axelarId properties
```
milapsheth, line 1061:
```suggestion
 * @param {Object} chain - The chain object containing axelarId property
```
Outdated. I'd rather delete the whole docstring. The function name is obvious. Comments add maintenance overhead, so we should only add them if we can't make the code self-descriptive

evm/hyperliquid.js
```diff
@@ -80,30 +80,22 @@ async function updateBlockSize(wallet, chain, options) {
     return result;
 }
 
-async function deployer(wallet, chain, options) {
-    const [tokenId] = options.args;
+async function deployer(wallet, chain, args) {
```
milapsheth, line 83:
```suggestion
async function deployer(wallet, chain, args, _options) {
```
since the caller is still passing in `options` even if we don't use it. same everywhere

milapsheth (approved):

evm/hyperliquid.js
```diff
@@ -0,0 +1,206 @@
...
+    await updateBlockSize(wallet, chain, [blockType]);
+}
+
+if (require.main === module) {
+    const program = new Command();
+
+    program.name('hyperliquid').description('Hyperliquid chain management commands');
+
+    program
+        .command('update-block-size <blockSize>')
```
milapsheth, line 181:
```suggestion
        .command('update-block-size <block-size>')
```
we use kebab case for the new cli syntax

evm/hyperliquid.js
```diff
@@ -0,0 +1,206 @@
...
+        .description('Update Hyperliquid block size')
+        .action((blockSize, options) => {
+            main(updateBlockSize, [blockSize], options);
+        });
+
+    program
+        .command('deployer <tokenId>')
```
milapsheth, line 188:
```suggestion
        .command('deployer <token-id>')
```

evm/hyperliquid.js
```diff
@@ -0,0 +1,206 @@
...
+        .description('Get deployer address for a Hyperliquid interchain token')
+        .action((tokenId, options) => {
+            main(deployer, [tokenId], options);
+        });
+
+    program
+        .command('update-token-deployer <tokenId> <deployer>')
```
milapsheth, line 195:
```suggestion
        .command('update-token-deployer <token-id> <deployer>')
```

PR #850, opened 2025-07-02
refactor: update the generate-config workflow to make it reusable

apolikamixitos:
# What:

- Refactor the workflow to make it reusable.

image


.github/workflows/reusable-update-config.yaml
```diff
@@ -0,0 +1,182 @@
...
+      - name: Create Pull Request
+        id: create_pr
+        run: |
+          cd config-repo
+          PR_TITLE="chore: update configs for ${{ env.NETWORKS }} from axelar-contract-deployments"
+          PR_BODY="Auto-generated PR with updated configuration files"
+          PR_HEAD="${{ steps.branch-name.outputs.branch }}"
+          PR_BASE="main"
+          # Capture both response and HTTP status code
+          RESPONSE=$(curl -s -w "%{http_code}" -o response.json -X POST \
```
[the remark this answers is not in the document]
apolikamixitos: @milapsheth, is there any reason why it's not great that we're using the API directly, and what's the advantage of using `gh` CLI instead? from the perspective of using the API, it gives us more flexibility to play with the body payload and build more customizable input in the processes we are having.

I would understand using `gh` for more basic manipulation on your machine for day-to-day operations, but I'm not sure of the added value of switching to `gh` in this case.

.github/workflows/generate-config.yaml
```diff
@@ -1,14 +1,21 @@
 name: Update config
-
 on:
   pull_request_target:
     types: [closed]
-
```
milapsheth, line 6: keep these newlines between sections for readability?

.github/workflows/reusable-update-config.yaml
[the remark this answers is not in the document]
milapsheth: the not very readable construction of the request, and parsing of the response we're doing isn't great. It also had bugs as we found before, precisely because it's trickier to deal with the API directly. Workflows usually use shell cmds so this is quite inconsistent

> from the perspective of using the API, it gives us more flexibility to play with the body payload and build more customizable input in the processes we are having

why can't we do this via the `gh` CLI? Let's try it out and see the difference

.github/actions/determine-networks/action.yaml
```diff
@@ -0,0 +1,48 @@
...
+            fi
+          done
+        done < <(echo '${{ inputs.paths_map }}' | jq -r 'to_entries[] | "\(.key)|\(.value | join(","))"')
+
+        if [ ${#NETWORKS[@]} -eq 0 ]; then
+          echo "No relevant changes detected."
+          echo "networks=none" >> $GITHUB_OUTPUT
+        else
+          echo "Detected networks: ${NETWORKS[*]}"
+          echo "networks=$(IFS=,; echo "${NETWORKS[*]}")" >> $GITHUB_OUTPUT
```
milapsheth, line 47: unnecessary complexity, let this action deal with a single file/network? The caller workflow can define a matrix to parallelly check multiple files/networks to keep it cleaner

.github/actions/update-config/action.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+        git config --local user.email "devops@interoplabs.io"
+        git config --local user.name "Interop Labs CI"
+        git checkout -b ${{ steps.branch-name.outputs.branch }}
+
+        IFS=',' read -ra NETWORK_LIST <<< "${{ inputs.networks }}"
+        for NETWORK in "${NETWORK_LIST[@]}"; do
+          echo "Generating config for $NETWORK"
+          poetry run python generate.py --network "$NETWORK" gen-configs all-configs
+        done
```
milapsheth, line 55: only handle a single network, let the caller define a matrix/handle the looping

milapsheth: also cleaner to have independent PRs for independent networks

.github/actions/update-config/action.yaml
```diff
@@ -0,0 +1,105 @@
...
+      run: |
+        cd config-repo
+        PR_NUMBER=${{ env.PR_NUMBER }}
+        PR_TITLE="chore: update configs for ${{ inputs.networks }} from ${{ inputs.source_repo }}"
+        # Attempt to merge the PR and capture both response and HTTP status code
+        RESPONSE=$(curl -s -w "%{http_code}" -o merge_response.json -X PUT \
+          -H "Authorization: token ${{ inputs.cicd_token }}" \
+          -H "Accept: application/vnd.github.v3+json" \
+          https://api.github.com/repos/${{ inputs.config_repo }}/pulls/$PR_NUMBER/merge \
+          -d "{\"commit_title\":\"Merge PR #$PR_NUMBER: $PR_TITLE\",\"merge_method\":\"squash\"}")
```
milapsheth, line 97:
```suggestion
          -d "{\"commit_title\":\"$PR_TITLE\",\"merge_method\":\"squash\"}")
```
Make the merge commits match the conventional commit format instead of adding this `Merge PR ...` prefix
https://github.com/axelarnetwork/chains/commits/main/

.github/workflows/reusable-update-config.yaml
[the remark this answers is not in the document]
apolikamixitos: That's quite a valid point, but I've thought mostly about the limitation of the `gh` CLI options, since we will always be tied to the available options provided on the CLI not on the API which is broader (the flexibility part I mentioned earlier), so far the `gh` CLI does offer basic PRs manipulation, I can move these options to use it instead.

.github/actions/determine-networks/action.yaml
apolikamixitos: @milapsheth coming back to your point, while working on the migration to `gh` CLI, I've came across a limitation about the `gh pr create` that doesn't return a `JSON` response to be able to extract the `PR_NUMBER` for example, so I had to make some changes to be able to get the PR number (with of course avoiding the "hackish" regex parsing of the URL struct and use `gh pr view` instead that returns the PR details in `json` format for example).

.github/actions/update-config/action.yaml
```diff
@@ -48,58 +48,56 @@ runs:
...
-        HTTP_STATUS=$(echo -n "$RESPONSE" | tail -c 3)
-        cat merge_response.json
-        if [[ "$HTTP_STATUS" != "200" ]]; then
-          echo "Failed to merge PR. HTTP status: $HTTP_STATUS"
-          cat merge_response.json
+        PR_TITLE="chore: update configs for ${{ inputs.network }} from ${{ inputs.source_repo }}"
+        MERGE_PR_OUTPUT=$(gh pr merge "$PR_NUMBER" \
+            --repo "${{ inputs.config_repo }}" \
+            --squash \
+            --subject "chore: merge PR #$PR_NUMBER: $PR_TITLE")
```
milapsheth, line 97:
```suggestion
        gh pr merge "$PR_NUMBER" \
            --repo "${{ inputs.config_repo }}" \
            --squash \
            --subject "chore: merge PR #$PR_NUMBER: $PR_TITLE"
```
we don't need the output of this cmd, so just let the cmd run and fail instead of checking separately?

.github/actions/determine-networks/action.yaml
milapsheth: Unresolved the conversation since the actions are still not refactored to handle a single network at a time

apolikamixitos: The process was indeed "per-network" with the `update-config` action, after the offline sync, I've implemented the "per-network" basis on the whole workflow instead (checking the network's changed files in a specific step instead of a specific action `determine-networks` that was discarded as suggested)

.github/actions/update-config/action.yaml
```diff
@@ -48,58 +48,56 @@ runs:
...
+          --body "$PR_BODY" \
+          --head "$PR_HEAD" \
+          --base "$PR_BASE")
+        CREATE_PR_EXIT_CODE=$?
+
+        if [ "$CREATE_PR_EXIT_CODE" -ne 0 ]; then
+            echo "Failed to create PR."
+            exit 1
         fi
+        PR_NUMBER=$(gh pr view "$CREATE_PR_URL" --repo "${{ inputs.config_repo }}" --json number -q .number)
```
milapsheth, line 83: Was `jq` supposed to be used here somehow?
```suggestion
        PR_NUMBER=$(gh pr view "$CREATE_PR_URL" --repo "${{ inputs.config_repo }}" --json | jq -q .number)
```

.github/actions/update-config/action.yaml
```diff
@@ -48,58 +48,56 @@ runs:
...
+        PR_TITLE="chore: update configs for ${{ inputs.network }} from ${{ inputs.source_repo }}"
+        gh pr merge "$PR_NUMBER" \
+            --repo "${{ inputs.config_repo }}" \
+            --squash \
+            --subject "chore: merge PR #$PR_NUMBER: $PR_TITLE"
+        MERGE_PR_EXIT_CODE=$?
+        if [ "$MERGE_PR_EXIT_CODE" -ne 0 ]; then
+          echo "Failed to merge PR #$PR_NUMBER."
           exit 1
         fi
```
milapsheth, line 102:
```suggestion
```
why is this needed? If the cmd fails, won't the step automatically fail?

.github/actions/update-config/action.yaml
```diff
@@ -48,58 +48,56 @@ runs:
...
-          -H "Accept: application/vnd.github.v3+json" \
-          https://api.github.com/repos/${{ inputs.config_repo }}/pulls/$PR_NUMBER/merge \
-          -d "{\"commit_title\":\"Merge PR #$PR_NUMBER: $PR_TITLE\",\"merge_method\":\"squash\"}")
-        HTTP_STATUS=$(echo -n "$RESPONSE" | tail -c 3)
-        cat merge_response.json
-        if [[ "$HTTP_STATUS" != "200" ]]; then
-          echo "Failed to merge PR. HTTP status: $HTTP_STATUS"
-          cat merge_response.json
+        PR_TITLE="chore: update configs for ${{ inputs.network }} from ${{ inputs.source_repo }}"
+        gh pr merge "$PR_NUMBER" \
```
milapsheth, line 94:
```suggestion

        gh pr merge "$PR_NUMBER" \
```
use newlines between variable declarations and cmds for readability

.github/workflows/reusable-update-config.yaml
```diff
@@ -21,31 +23,21 @@ permissions:
...
-          paths_map: ${{ inputs.paths_map }}
   update-config:
     env:
       GITHUB_TOKEN: ${{ secrets.SOURCE_REPO_GITHUB_TOKEN }}
-    needs: determine-networks
-    if: needs.determine-networks.outputs.networks != 'none'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Get all changed ${{ inputs.network }} changed files
```
milapsheth, line 32: use newlines between steps for readability

.github/workflows/reusable-update-config.yaml
```diff
@@ -3,10 +3,12 @@ name: Reusable Update Config
 on:
   workflow_call:
     inputs:
-      paths_map:
+      network:
+        required: true
+        type: string
+      files:
```
milapsheth, line 9: if multiple files are accepted, document the syntax so usage is clear. Otherwise, change variable name to `file`

.github/actions/update-config/action.yaml
apolikamixitos: @milapsheth no, the command line has a built-in jq expression filtering

.github/actions/update-config/action-original.yaml
```diff
@@ -0,0 +1,105 @@
+name: Update Config
```
milapsheth, line 1: Checked in by mistake?

PR #855, opened 2025-07-09
chore(releases): EVM ITS v2.1.1 release doc

milapsheth:


isi8787 (commented):
why does the interchainTokenDeployer address vary in a few instances (verified in testnet, stagenet, mainnet only 1)  ? All the other ones are pretty consistent for all deployments?

PR #858, opened 2025-07-09
fix: enable immediate PR merge

apolikamixitos:
# What:

- add a flag to enable immediate PR merge


.github/actions/update-config/action.yaml
```diff
@@ -95,5 +95,6 @@ runs:
         gh pr merge "$PR_NUMBER" \
```
milapsheth, line 95: revert the config change as well to trigger this action

milapsheth (approved):

.github/actions/update-config/action.yaml
apolikamixitos: Done

PR #862, opened 2025-07-11
fix: update transceiver script

blockchainguyy:


evm/utils.js
```diff
@@ -867,27 +878,6 @@ function getQualifiedContractName(contractName) {
...
- * @returns {Object} An object containing gas options for the transaction.
- *
- * @throws {Error} Throws an error if gas options are invalid and/or fetching the gas price fails when gasPriceAdjustment is present.
- *
- * Note:
- * - If 'options.gasOptions' is present, cli gas options override any config values.
- * - If 'options.offline' is true, static gas options from the contract or chain config are used.
- * - If 'gasPriceAdjustment' is set in gas options and 'gasPrice' is not pre-defined, the gas price
- *   is fetched from the provider and adjusted according to 'gasPriceAdjustment'.
- */
```
kulikthebird, line 890: This docstrict contains a lot of context that is not obvious when looking at the function body

evm/utils.js
```diff
@@ -829,7 +821,6 @@ function getContractPath(contractName) {
         path.join(findProjectRoot(__dirname), 'node_modules', '@axelar-network', 'axelar-cgp-solidity', 'artifacts', 'contracts'),
         path.join(findProjectRoot(__dirname), 'node_modules', '@axelar-network', 'interchain-token-service', 'artifacts', 'contracts'),
         path.join(findProjectRoot(__dirname), 'evm', 'legacy'),
-        path.join(findProjectRoot(__dirname), 'node_modules', '@openzeppelin', 'contracts'),
```
kulikthebird, line 832: While having the **OZ** repository specified here _and_ precompiled `AxelarTransceiver` contract, the deployment procedure was pretty well automated - a user could simply run the command and specify the `TransceiverStruct` library address on the destination chain. With the current approach a user needs to download 3rd party repo, compile the contracts and provide the path to artifacts. Let's discuss on the benefits of the current approach

evm/deploy-contract.js
```diff
@@ -501,22 +406,41 @@ async function processCommand(config, chain, options) {
...
-        options,
-        wallet,
-        deployMethod,
-        codehash,
-        predeployCodehash,
-        salt,
-        constructorArgs,
-    );
+    // Update configuration
+    if (contractName === 'ERC1967Proxy') {
```
kulikthebird, line 410: Why not use the previously implemented `updateConfig` method?

evm/axelar-transceiver.ts
```diff
@@ -1,197 +1,361 @@
...
-
-/**
- * @fileoverview AxelarTransceiver Post-Deployment Operations Script
- *
- * This script handles post-deployment operations for AxelarTransceiver contracts:
- * - Contract initialization
- * - Pauser capability transfer
- *
- * Usage: ts-node evm/axelar-transceiver.ts -e testnet -n ethereum --initialize --pauserAddress 0x...
- */
```
kulikthebird, line 11: The file overview description is really helpful for people who are not familiar with the repository (and also for our future selves after a long break).

evm/test/test-proxy-args.js
```diff
@@ -1,131 +0,0 @@
-'use strict';
```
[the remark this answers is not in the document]
kulikthebird: That's coorect when it comes to testing the contracts in this script - we don't want to do this. This test suite is focues more on our code - checks that if a user call the cmd, the procedure will run the deployment correctly and update the config

PR #864, opened 2025-07-15
feat: add solana template release docs, cli tools and template configs

eloylp:
This PR was already reviewed at https://github.com/eigerco/axelar-contract-deployments/pull/1


eloylp:
@isi8787 porting the previous approved PR to upstream

eloylp:
> Bug: Solana Address Decoding Function Fails

Fixed at https://github.com/axelarnetwork/axelar-contract-deployments/pull/864/commits/4eb3bd8209de3e4a2bec89984fcd1e8ea67abd9c

isi8787 (approved):

solana/src/types.rs
```diff
@@ -74,12 +77,11 @@ pub(crate) struct ChainAxelarId(pub(crate) String);
...
-        // Currently all network types use the same Axelar ID prefix
         match value {
-            NetworkType::Local
-            | NetworkType::Devnet
-            | NetworkType::Testnet
-            | NetworkType::Mainnet => Self("solana".to_owned()),
+            NetworkType::Local => Self("solana-local".to_owned()),
+            NetworkType::Devnet => Self("solana-devnet".to_owned()),
+            NetworkType::Testnet => Self("solana-testnet".to_owned()),
+            NetworkType::Mainnet => Self("solana".to_owned()),
```
isi8787, line 84: why cant we use a cli or ENV value for the correct mapping, this wont address the issue that stagenet is goign to have a unique name. To be clear because of the previous bad deployment the Envrionements are going to end up being:
devnet-amplifier-> solana
stagenet -> solana-2
testnet -> solana
mainnet -> solana

eloylp: If i understood correctly, the problem we are facing here is that chainId cannot be re-used across the different json config files.

This NetworkType is [calculated](https://github.com/eigerco/axelar-contract-deployments/blob/d4485034882d7213672b4fb0f67a0e700174ea53/solana/src/config.rs#L38-L39) and [used internally](https://github.com/eigerco/axelar-contract-deployments/blob/d4485034882d7213672b4fb0f67a0e700174ea53/solana/src/gateway.rs#L504-L512) for knowing which chainid to use when querying the config files.

>devnet-amplifier-> solana
stagenet -> solana-2
testnet -> solana
mainnet -> solana

So now the above is confusing for me if we cannot reuse chainid's . Shouldn't we have one per each environment ? Let me know

eloylp:
Adding @frenzox to the loop.

solana/src/types.rs
eloylp: Ok clarified , heres the fix that makes chain_id an argument 68c4e07975cb2c4d2fb6ffe9e465facd9ba43a18

PR #869, opened 2025-07-16
ci: only run evm tests on relevant changes

milapsheth:


.github/workflows/test-evm.yaml
```diff
@@ -3,7 +3,33 @@ name: Test Evm
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            evm:
+              - 'evm/**'
+            common:
+              - 'common/**'
```
isi8787, line 20: i know this is focused on EVM but there are some common files that affect stellar

isi8787 (approved):

PR #893, opened 2025-07-25
refactor: forward chains config instead of the whole config

kulikthebird:


common/utils.js
```diff
@@ -519,12 +519,12 @@ async function getDomainSeparator(config, chain, options) {
     return expectedDomainSeparator;
```
milapsheth, line 519: maybe use a more descriptive title with the right tag, e.g. `refactor: forward chains config instead of the whole config`

milapsheth: https://www.conventionalcommits.org/en/v1.0.0/#summary

milapsheth (approved):

PR #894, opened 2025-07-25
chore: remove is valid dest chain

kulikthebird:


evm/utils.js
```diff
@@ -938,8 +938,10 @@ function validateGasOptions(gasOptions) {
     }
 }
 
-function isValidChain(config, chainName) {
-    const chains = config.chains;
+function isValidChain(chains, chainName) {
+    if (chainName === '') {
+        return;
+    }
```
milapsheth, line 944: doesn't look very valid

evm/utils.js
```diff
@@ -938,8 +938,10 @@ function validateGasOptions(gasOptions) {
     }
 }
 
-function isValidChain(config, chainName) {
-    const chains = config.chains;
+function isValidChain(chains, chainName) {
```
milapsheth, line 941: function name is a misnomer since it doesn't return a bool
```suggestion
function validateChain(chains, chainName) {
```

evm/its.js
```diff
@@ -587,7 +579,7 @@ async function processCommand(config, chain, action, options) {
                 isValidAddress: { destinationTokenAddress, operator },
                 isValidNumber: { gasValue, tokenManagerType },
             });
-            isValidDestinationChain(config, destinationChain);
+            isValidChain(config.chains, destinationChain);
```
milapsheth, line 582: doesn't seem like an empty string should be treated as valid anymore, so let's fix the root cause

kulikthebird: Yeah I wanted to preserve the implementation without going deep into it's sense. You're right that it looks weird, let's remove it and we'll see if that was needed

PR #895, opened 2025-07-25
chore: refactor process command signature

kulikthebird:


evm/utils.js
```diff
@@ -484,18 +484,18 @@ const getDeployedAddress = async (deployer, deployMethod, options = {}) => {
     }
 };
 
-const getProxy = async (config, chain) => {
-    const address = (await httpGet(`${config.axelar.lcd}/axelar/evm/v1beta1/gateway_address/${chain}`)).address;
+const getProxy = async (axelarConfig, chain) => {
```
milapsheth, line 487:
```suggestion
const getProxy = async (axelar, chain) => {
```
could just name it `axelar` for brevity similar to how we name `chain` instead of `chainConfig`

evm/utils.js
```diff
@@ -751,6 +751,8 @@ const mainProcessor = async (options, processCommand, save = true, catchErr = fa
         return;
     }
 
+    const chainsSnapshot = JSON.parse(JSON.stringify(config.chains));
```
milapsheth, line 754:
```suggestion
    const chainsSnapshot = deeepcopy(config.chains);
```
create a common util with a more meaningful name?

milapsheth, line 754:
```suggestion
    const chains = JSON.parse(JSON.stringify(config.chains));
```
maybe call it `chains` or `chainsConfig` for consistency?

milapsheth (approved):

PR #920, opened 2025-08-06
fix: add backward compatibility to ITS version 2.1.1

isi8787:


evm/interchainTokenFactory.js
```diff
@@ -169,7 +169,7 @@ async function processCommand(_axelar, chain, chains, options) {
                 isValidNumber: { gasValue },
             });
 
-            if (!(await interchainTokenService.isTrustedChain(destinationChain))) {
+            if ((await interchainTokenService.trustedAddress(destinationChain)) === '') {
```
milapsheth, line 172: instead of this, change the check to use `isValidChain` helper that does a lookup in our config? this way it works for both version. Hyperliquid is on v2.2.0, so this script will break if this check persists

evm/interchainTokenFactory.js
```diff
@@ -265,7 +265,7 @@ async function processCommand(_axelar, chain, chains, options) {
 
             const deploymentSalt = getDeploymentSalt(options);
 
-            if (!(await interchainTokenService.isTrustedChain(destinationChain))) {
+            if ((await interchainTokenService.trustedAddress(destinationChain)) === '') {
```
milapsheth, line 268: same

evm/interchainTokenFactory.js
```diff
@@ -265,7 +265,7 @@ async function processCommand(_axelar, chain, chains, options) {
```
milapsheth, line 265: trusted address set/remove is still using the new methods, switch those to use `setTrustedAddress(chain, 'hub')`?

milapsheth:
bugbot run

evm/interchainTokenFactory.js
isi8787: im not sure about the approach since the actual package imported is going to have a different ABI depending on version so trying to use both methods under a helper is not going to work

evm/its.js
```diff
@@ -44,6 +46,8 @@ const tokenManagerImplementations = {
     MINT_BURN: 4,
 };
 
+const IInterchainTokenServiceV211 = require('@axelar-network/interchain-token-service-v2.1.1/artifacts/contracts/interfaces/IInterchainTokenService.sol/IInterchainTokenService.json');
```
milapsheth, line 49: use `getContractJSON` as per our convention, it takes a second argument

evm/its.js
```diff
@@ -134,7 +140,12 @@ async function processCommand(_axelar, chain, chains, action, options) {
     printInfo('Contract name', contractName);
     printInfo('Contract address', interchainTokenServiceAddress);
 
-    const interchainTokenService = new Contract(interchainTokenServiceAddress, IInterchainTokenService.abi, wallet);
+    let interchainTokenService;
+    if (itsVersion === '2.1.1') {
```
milapsheth, line 144: create a small helper for this, so that the local variable can still be a const

evm/its.js
```diff
@@ -419,15 +430,32 @@ async function processCommand(_axelar, chain, chains, action, options) {
                 return;
             }
 
-            const data = [];
+            if (itsVersion === '2.1.1') {
```
milapsheth, line 433: function is becoming complex and unreadable with all this nesting, either simplify or need a helper

evm/its.js
```diff
@@ -419,15 +430,32 @@ async function processCommand(_axelar, chain, chains, action, options) {
...
+                    const chainConfig = chains[trustedChain];
+                    if (chainConfig?.contracts?.InterchainTokenService?.address) {
+                        trustedAddresses.push(chainConfig.contracts.InterchainTokenService.address);
+                    } else {
+                        throw new Error(`No InterchainTokenService address found for chain ${trustedChain}`);
+                    }
+                }
+                
+                for (const [trustedChain, trustedAddress] of trustedChains.map((chain, index) => [chain, trustedAddresses[index]])) {
+                    const tx = await interchainTokenService.setTrustedAddress(trustedChain, trustedAddress, gasOptions);
```
milapsheth, line 445:
```suggestion
                    const tx = await interchainTokenService.setTrustedAddress(trustedChain, 'hub', gasOptions);
```
trusted address should be set to `hub`, we're only adding amplifier chains now, so we can assume that to remove all this unnecessary complexity

evm/its.js
```diff
@@ -419,15 +430,32 @@ async function processCommand(_axelar, chain, chains, action, options) {
...
+                }
+            } else {
+                const data = [];
 
-            for (const trustedChain of trustedChains) {
-                const tx = await interchainTokenService.populateTransaction.setTrustedChain(trustedChain, gasOptions);
-                data.push(tx.data);
-            }
+                for (const trustedChain of trustedChains) {
+                    const tx = await interchainTokenService.populateTransaction.setTrustedChain(trustedChain, gasOptions);
```
milapsheth, line 452: only this line needs to be different depending on the version, not the full loop

evm/its.js
```diff
@@ -444,15 +472,22 @@ async function processCommand(_axelar, chain, chains, action, options) {
                 return;
             }
 
-            const data = [];
+            if (itsVersion === '2.1.1') {
```
milapsheth, line 475: same comments

PR #921, opened 2025-08-07
fix: config value destructuring for deploying contracts

drewstaylor:
Fixes deploying Sui move contracts


drewstaylor:
Closing PR as reverting it this way will break other chains. Handing this off to @kulikthebird who will fix the Sui call

PR #924, opened 2025-08-07
feat(evm): add its version to deploy-its script

isi8787:


evm/utils.js
```diff
@@ -1083,12 +1084,24 @@ async function printTokenInfo(tokenAddress, provider) {
...
     } else {
-        return ((await interchainTokenService.isTrustedChain(destinationChain)));
+        return await interchainTokenService.isTrustedChain(destinationChain);
     }
 }
 
+function detectITSVersion(options) {
+    let itsVersion;
+
+    if (options.itsVersion) {
```
milapsheth, line 1096: it's always going to be the dependency version that is being deployed. so, allowing an override seems confusing and potentially error-prone. Let's avoid adding code we don't need yet

evm/deploy-its.js
```diff
@@ -84,6 +87,7 @@ async function deployAll(axelar, wallet, chain, chains, options) {
     contractConfig.salt = salt;
     contractConfig.proxySalt = proxySalt;
     contractConfig.deployer = wallet.address;
+    contractConfig.version = itsVersion;
 
     itsFactoryContractConfig.deployer = wallet.address;
     itsFactoryContractConfig.salt = factorySalt;
```
milapsheth, line 93: set it in the ITS factory contract as well

evm/deploy-its.js
```diff
@@ -420,7 +424,11 @@ async function upgrade(_axelar, chain, _chains, options) {
         return;
     }
 
-    printInfo(`Upgrading Interchain Token Service on ${chain.name}.`);
+    const itsVersion = detectITSVersion(options);
+
+    printInfo(`Upgrading Interchain Token Service on ${chain.name} to version ${itsVersion}.`);
+
+    contractConfig.version = itsVersion;
```
milapsheth, line 431: when upgrading, the default npm package will be on the new version too

releases/solana/Programs-upgrade-template.md
milapsheth, line 1: unrelated change?

evm/deploy-its.js
isi8787: updated the detectITSVersion to just use npm package and not take any cli input

PR #925, opened 2025-08-08
feat: add its-hub-update-chains command

AttissNgo:


cosmwasm/submit-proposal.js
```diff
@@ -358,6 +366,28 @@ const programHandler = () => {
         });
     addAmplifierOptions(registerItsChainCmd, { proposalOptions: true, runAs: true });
 
+    const updateItsChainCmd = program
```
isi8787, line 369: Seems like a lot of duplication of code, can we add an optional parameter to the its-hub-register-chains command for upgrade, that can reuse most of the code to just change the `msg: `{ "${operation}_chains": { "chains": ${JSON.stringify(chains)} } }`,` object

PR #942, opened 2025-08-19
chore: add version options for InterchainToken and TokenManager when …

AttissNgo:
…deploying Stellar ITS

### DIFF
--- a/stellar/deploy-contract/commands.js
+++ b/stellar/deploy-contract/commands.js
@@ -17,12 +17,32 @@ const CONTRACT_DEPLOY_OPTIONS = {
     AxelarExample: () => [
         new Option('--use-dummy-its-address', 'use dummy its address for AxelarExample contract to test a GMP call').default(false),
     ],
+    InterchainTokenService: () => [
+        new Option(
+            '--interchain-token-version <interchainTokenVersion>',
+            'version for InterchainToken contract (defaults to InterchainTokenService version)',
+        ),
+        new Option(
+            '--token-manager-version <tokenManagerVersion>',
+            'version for TokenManager contract (defaults to InterchainTokenService version)',
+        ),
+    ],
 };
 
 const CONTRACT_UPGRADE_OPTIONS = {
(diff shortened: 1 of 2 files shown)


stellar/deploy-contract/commands.js
```diff
@@ -17,12 +17,32 @@ const CONTRACT_DEPLOY_OPTIONS = {
...
+            'version for TokenManager contract (defaults to InterchainTokenService version)',
+        ),
+    ],
 };
 
 const CONTRACT_UPGRADE_OPTIONS = {
     AxelarGateway: () => [new Option('--migration-data <migrationData>', 'migration data').default(null, '()')],
     AxelarOperators: () => [new Option('--migration-data <migrationData>', 'migration data').default(null, '()')],
-    InterchainTokenService: () => [new Option('--migration-data <migrationData>', 'migration data').default(null, '()')],
+    InterchainTokenService: () => [
```
ahramy, line 35: We don’t want to maintain specific versions of the Interchain Token and Token Manager during upgrades. As Milap mentioned earlier, upgrading these contracts would require upgrading all instances of them. (The ITS can be upgraded independently without upgrading the Interchain Token and Token Manager)

stellar/deploy-contract/commands.js
```diff
@@ -17,6 +17,16 @@ const CONTRACT_DEPLOY_OPTIONS = {
     AxelarExample: () => [
         new Option('--use-dummy-its-address', 'use dummy its address for AxelarExample contract to test a GMP call').default(false),
     ],
+    InterchainTokenService: () => [
+        new Option(
+            '--interchain-token-version <interchainTokenVersion>',
+            'version for InterchainToken contract (defaults to InterchainTokenService version)',
```
milapsheth, line 23: I recommend requiring these to avoid incorrect usage by default. It's not even guaranteed that the ITS version is a valid release version for those crates

milapsheth (approved):

PR #991, opened 2025-09-04
feat: allow auto gas estimation for ITS msgs

benjamin852:
## Greptile Summary

Updated On: 2025-09-04 19:36:27 UTC

This PR introduces automatic gas fee estimation functionality for Interchain Token Service (ITS) operations. The changes add a new utility function `calculateItsCrossChainGas` in `common/utils.js` that integrates with Axelar's fee estimation API to automatically calculate appropriate gas fees for cross-chain operations.

The implementation works by making HTTP POST requests to Axelar's API endpoints (`api.axelarscan.io` for mainnet, `testnet.api.gmp.axelarscan.io` for testnet) to get fee estimates based on source chain, destination chain, and event type. The function includes intelligent chain name normalization logic that handles testnet L2 chains which use different naming conventions (e.g., 'optimism' becomes 'optimism-sepolia' on testnet).

In `evm/its.js`, this new functionality is integrated into three key ITS operations:
1. **Interchain transfers** - estimates gas for `InterchainTransfer` events
2. **Token metadata registration** - estimates gas for `TokenMetadataRegistered` events (always to Axelar chain)
3. **Token linking** - estimates gas for `LinkToken` events

The integration is designed to be backward compatible - when `gasValue` is 0 (the default), the system automatically estimates the required gas fee. When users explicitly provide a gas value, that value is used instead. This approach maintains existing functionality while adding the convenience of automatic estimation.

The change fits well within the existing codebase architecture by leveraging the existing `httpPost` utility function and placing the new function in `common/utils.js` where it can be reused across different modules. The CLI help text has also been updated to clarify that gas estimation will be used when no explicit gas value is provided.

## Important Files Changed

<details><summary>Changed Files</summary>

| Filename | Score | Overview |
|----------|-------|----------|
| common/utils.js | 4/5 | Added `calculateItsCrossChainGas` function for automatic gas fee estimation with API integration and chain name normalization |
| evm/its.js | 4/5 | Integrated automatic gas estimation into three ITS operations when gasValue is 0, maintaining backward compatibility |

</details>

## Confidence score: 4/5

- This PR is safe to merge with low risk of production issues
- Score reflects well-structured implementation with proper error handling and backward compatibility
- Pay attention to the API integration dependencies and error handling in the gas estimation function

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -790,7 +790,7 @@ async function estimateITSFee(chain, destinationChain, env, eventType, gasValue,
         return 0;
     }
 
-    if(gasValue != 'auto' && !isValidNumber(gasValue)) {
+    if (gasValue != 'auto' && !isValidNumber(gasValue)) {
         throw new Error(`Invalid gas value: ${gasValue}`);
     }
 


common/utils.js
```diff
@@ -795,6 +795,44 @@ function validateDestinationChain(chains, destinationChain) {
...
+        sourceChain: normalizeChainName(sourceChain),
+        destinationChain: normalizeChainName(destinationChain),
+        event: eventType,
+    };
+
+    const gasFee = await httpPost(url, payload);
+    if (!gasFee) {
+        throw new Error(`Failed to estimate ITS fee: ${JSON.stringify(gasFee)}`);
+    }
+    return gasFee;
```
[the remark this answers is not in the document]
isi8787: see above

evm/interchainTokenFactory.js
```diff
@@ -366,7 +375,7 @@ if (require.main === module) {
     program.addOption(new Option('--initialSupply <initialSupply>', 'initial supply').default(1e9));
     program.addOption(new Option('--destinationChain <destinationChain>', 'destination chain'));
     program.addOption(new Option('--destinationAddress <destinationAddress>', 'destination address'));
-    program.addOption(new Option('--gasValue <gasValue>', 'gas value').default(0));
+    program.addOption(new Option('--gasValue <gasValue>', 'gas value, will default to estimate gas fee if none provided').default(0));
```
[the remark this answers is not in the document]
isi8787: idea is that API estimate is the default, and that we dont set to 0. If user wants to pass then they can pass a value

evm/its.js
```diff
@@ -365,20 +376,30 @@ async function processCommand(_axelar, chain, chains, action, options) {
...
+            const { gasValue, env } = options;
+
+            const submittedGasValue = !gasValue
+                ? await calculateItsCrossChainGas({
+                      sourceChain: chain.axelarId,
+                      destinationChain,
+                      env,
+                      eventType: 'TokenMetadataRegistered',
+                  })
+                : gasValue;
```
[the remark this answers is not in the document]
benjamin852: not sure why it says this we pass in `'axelar'` to define destination chain confirmed in logs when i tested the `destinationChain` has a value

common/utils.js
```diff
@@ -795,6 +795,24 @@ function validateDestinationChain(chains, destinationChain) {
...
+
+    const payload = {
+        sourceChain: sourceChain,
+        destinationChain: destinationChain,
+        event: eventType,
+    };
+
+    const gasFee = await httpPost(url, payload);
+    if (gasFee === undefined || gasFee === null) {
+        throw new Error(`Failed to estimate ITS fee: ${gasFee}`);
```
milapsheth, line 811: can this even return `undefined`? Wont' the httpPost just fail if we provide bad params?

common/utils.js
```diff
@@ -795,6 +795,24 @@ function validateDestinationChain(chains, destinationChain) {
     validateChain(chains, destinationChain);
 }
 
+async function calculateItsCrossChainGas(options) {
```
milapsheth, line 798:
```suggestion
async function estimateITSFee(options) {
```
reuse the same name as API for clarity

evm/interchainTokenFactory.js
```diff
@@ -228,21 +238,30 @@ async function processCommand(_axelar, chain, chains, options) {
...
         case 'deployRemoteCanonicalInterchainToken': {
-            const { tokenAddress, destinationChain, gasValue } = options;
+            const { tokenAddress, destinationChain, gasValue, env } = options;
+
+            const submittedGasValue = !gasValue
+                ? await calculateItsCrossChainGas({
+                      sourceChain: chain.axelarId,
+                      destinationChain,
+                      env,
+                      eventType: 'InterchainTokenDeployment',
```
milapsheth, line 248: `options` as an arg is meant for pass through from CLI args. If you're just creating your own object, it's worse since you lose the type structure. So I'd pass in as separate args

common/utils.js
```diff
@@ -795,6 +795,24 @@ function validateDestinationChain(chains, destinationChain) {
...
+    const { sourceChain, destinationChain, env, eventType } = options;
```
milapsheth, line 799: confusing, better to use separate args than create your own object structure

evm/its.js
```diff
@@ -666,14 +686,23 @@ async function processCommand(_axelar, chain, chains, action, options) {
...
             const deploymentSalt = getDeploymentSalt(options);
 
+            const submittedGasValue = !gasValue
+                ? await calculateItsCrossChainGas({
+                      sourceChain: chain.axelarId,
+                      destinationChain,
+                      env,
+                      eventType: 'LinkToken',
+                  })
+                : gasValue;
```
milapsheth, line 699:
```suggestion
            const { env } = options;
            const deploymentSalt = getDeploymentSalt(options);

            const gasValue = options.gasValue
                || await calculateItsCrossChainGas({
                      sourceChain: chain.axelarId,
                      destinationChain,
                      env,
                      eventType: 'LinkToken',
                  });
```
more concise using `||`, less intermediate variables

evm/interchainTokenFactory.js
```diff
@@ -228,21 +238,30 @@ async function processCommand(_axelar, chain, chains, options) {
...
 
             validateChain(chains, destinationChain);
 
             const tx = await interchainTokenFactory['deployRemoteCanonicalInterchainToken(address,string,uint256)'](
                 tokenAddress,
                 destinationChain,
-                gasValue,
-                { value: scaleGasValue(chain, gasValue), ...gasOptions },
+                submittedGasValue,
+                { value: scaleGasValue(chain, submittedGasValue), ...gasOptions },
```
milapsheth, line 264: did you look at what `scaleGasValue` does though? the estimated value should be scaled already

evm/its.js
```diff
@@ -365,25 +376,34 @@ async function processCommand(_axelar, chain, chains, action, options) {
...
+                      eventType: 'TokenMetadataRegistered',
+                  })
+                : gasValue;
+
+            validateParameters({ isValidAddress: { tokenAddress }, isValidNumber: { submittedGasValue } });
 
-            const tx = await interchainTokenService.registerTokenMetadata(tokenAddress, gasValue, {
-                value: scaleGasValue(chain, gasValue),
+            const tx = await interchainTokenService.registerTokenMetadata(tokenAddress, submittedGasValue, {
+                value: scaleGasValue(chain, submittedGasValue),
```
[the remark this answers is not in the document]
milapsheth: this is a valid bug

evm/its.js
```diff
@@ -938,7 +967,7 @@ if (require.main === module) {
         .addArgument(new Argument('<type>', 'Token manager type').choices(Object.keys(tokenManagerTypes)))
         .argument('<operator>', 'Operator address')
         .addOption(new Option('--rawSalt <rawSalt>', 'raw deployment salt').env('RAW_SALT'))
-        .addOption(new Option('--gasValue <gasValue>', 'gas value').default(0))
+        .addOption(new Option('--gasValue <gasValue>', 'gas value'))
```
milapsheth, line 970:
```suggestion
        .addOption(new Option('--gasValue <gasValue>', 'gas value').default('auto'))
```
use an explicit default to distinguish from `0` and other values that are provided

evm/interchainTokenFactory.js
milapsheth: I'd suggest making `estimateITSFee` accept `options.gasValue` as an arg, and have it check if it's `auto` to estimate it, or just scale it if it's a number, and fail otherwise

common/utils.js
benjamin852: getting diverging recommendations here, see prev [comment](https://github.com/axelarnetwork/axelar-contract-deployments/pull/991#discussion_r2325724766). I'm open to either but just want to make sure we're consistent across the repo

axelar-chains-config/info/testnet.json
```diff
@@ -3768,6 +3768,7 @@
     "grpc": "grpc.axelar-testnet.lava.build:443",
     "tokenSymbol": "AXL",
     "axelarscanApi": "https://testnet.api.axelarscan.io",
+    "gmpAxelarscanApi": "https://testnet.api.gmp.axelarscan.io",
```
isi8787, line 3771: you are going to need to add to stagenet.json and devnet-amplifier.json . For stagenet i know there i working api but not sure about devnet-amplifier

common/utils.js
```diff
@@ -795,6 +795,28 @@ function validateDestinationChain(chains, destinationChain) {
     validateChain(chains, destinationChain);
 }
 
+async function estimateITSFee(options) {
+    const { chain, destinationChain, env, eventType, gasValue } = options;
+
+    if (gasValue === 'auto') {
+        const baseUrl = loadConfig(env).axelar.gmpAxelarscanApi;
```
isi8787, line 802: since not all environments (especially custom devnets) might have need to have a fallback mechanism -> the 0 gasValue that is default behavior

evm/its.js
```diff
@@ -366,24 +375,32 @@ async function processCommand(_axelar, chain, chains, action, options) {
...
+                destinationChain: 'axelar',
+                env,
+                eventType: 'TokenMetadataRegistered',
+                gasValue: options.gasValue,
+            });
+
             validateParameters({ isValidAddress: { tokenAddress }, isValidNumber: { gasValue } });
 
             const tx = await interchainTokenService.registerTokenMetadata(tokenAddress, gasValue, {
-                value: scaleGasValue(chain, gasValue),
```
isi8787, line 381: the scaleGasValue is still needed

evm/its.js
```diff
@@ -366,24 +375,32 @@ async function processCommand(_axelar, chain, chains, action, options) {
                 amountInUnits,
                 metadata,
                 gasValue,
-                { value: scaleGasValue(chain, gasValue), ...gasOptions },
```
isi8787, line 369: scaleGasValue should still be used, otherwise hedera might be affected

evm/interchainTokenFactory.js
```diff
@@ -19,7 +19,7 @@ const {
     printTokenInfo,
     isTrustedChain,
     encodeITSDestination,
-    scaleGasValue,
```
isi8787, line 22: why remove this?

common/utils.js
```diff
@@ -795,6 +795,28 @@ function validateDestinationChain(chains, destinationChain) {
...
+        const payload = {
+            sourceChain: chain.axelarId,
+            destinationChain: destinationChain,
+            event: eventType,
+        };
+
+        return await httpPost(url, payload);
+    } else if (isValidNumber(gasValue)) {
+        const { scaleGasValue } = require('../evm/utils');
+        return scaleGasValue(chain, gasValue);
```
isi8787, line 814: i beleive scaleGasValue is still going to be needed even if getting from the API, please double check with @blockchainguyy

evm/interchainTokenFactory.js
benjamin852: Kept the logic, just moved it to the `estimateItsFee()` function call.

common/utils.js
benjamin852: I recall @milapsheth mentioning the estimated value should be scaled already but will confirm with @blockchainguyy too

common/utils.js
```diff
@@ -795,6 +795,39 @@ function validateDestinationChain(chains, destinationChain) {
     validateChain(chains, destinationChain);
 }
 
+async function estimateITSFee(options) {
+    const { chain, destinationChain, env, eventType, gasValue } = options;
```
isi8787, line 799: Original suggestion was to reuse some of the existing variables that are in the global options but that was shortsighted since you still had some inputs that were not part of that. So  instead of creating a dedicated new local options that you just unwrap  lets them be the actual input parameters.

common/utils.js
```diff
@@ -795,6 +795,44 @@ function validateDestinationChain(chains, destinationChain) {
     validateChain(chains, destinationChain);
 }
 
+async function estimateITSFee(chain, destinationChain, env, eventType, gasValue) {
+    if (gasValue === 'auto') {
+        let baseUrl;
+        try {
+            baseUrl = loadConfig(env).axelar?.gmpAxelarscanApi;
```
kulikthebird, line 802: loadConfig is called upstreamby the mainProcessor. Read the `axelar.gmpAxelarscanApi` from the `axelar` config in the `processCommand` and pass it here as an argument

benjamin852: good catch added extra param to pass in the config.

kulikthebird (changes requested):
One more update from me.  Other than that the feature looks nice, thanks!

common/utils.js
```diff
@@ -795,6 +795,44 @@ function validateDestinationChain(chains, destinationChain) {
...
+
+        const url = `${baseUrl}/estimateITSFee`;
+
+        const payload = {
+            sourceChain: chain.axelarId,
+            destinationChain: destinationChain,
+            event: eventType,
+        };
+
+        return await httpPost(url, payload);
```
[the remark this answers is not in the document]
kulikthebird: I checked how the httpPost is used in other places and it looks good as it is now 👍

common/utils.js
```diff
@@ -795,6 +795,50 @@ function validateDestinationChain(chains, destinationChain) {
...
+    }
+}
+
+function scaleGasValue(chain, gasValue) {
+    if (typeof chain.gasScalingFactor === 'number') {
+        return BigNumber.from(gasValue).mul(BigNumber.from(10).pow(chain.gasScalingFactor));
+    }
+
+    return gasValue;
+}
```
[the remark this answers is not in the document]
benjamin852: the api returns a scaled number

axelar-chains-config/info/mainnet.json
```diff
@@ -3468,6 +3468,7 @@
     "grpc": "grpc.axelar.lava.build:443",
     "tokenSymbol": "AXL",
     "axelarscanApi": "https://api.axelarscan.io",
+    "gmpAxelarscanApi": "https://api.gmp.axelarscan.io",
```
milapsheth, line 3471:
```suggestion
    "axelarscanGmpApi": "https://api.gmp.axelarscan.io",
```
Reads better maybe

milapsheth, line 3471: Is this actually needed? We already have the API above. As per docs, you can use the path `{API}/gmp` for the request

milapsheth, line 3471: See try it
https://docs.axelarscan.io/gmp#estimateITSFee

common/utils.js
```diff
@@ -785,6 +785,39 @@ function validateDestinationChain(chains, destinationChain) {
     validateChain(chains, destinationChain);
 }
 
+async function estimateITSFee(chain, destinationChain, env, eventType, gasValue, gmpAxelarscanApi) {
+    if (!gmpAxelarscanApi || env === 'devnet-amplifier') {
```
milapsheth, line 789: Why would API not be defined. Treat it as a required arg

common/utils.js
```diff
@@ -785,6 +785,39 @@ function validateDestinationChain(chains, destinationChain) {
...
+
+    const res = await httpPost(url, payload);
+    if (res.error) {
+        throw new Error(`Error querying gas amount: ${res.error}`);
+    }
+    return res;
+}
+
+function scaleGasValue(chain, gasValue) {
+    if (typeof chain.gasScalingFactor === 'number') {
```
milapsheth, line 814: Where is gas scaling factor defined? Don't chains already have a token decimals in their config that can be reused

milapsheth, line 814: And let's always require it. Otherwise the behaviour is inconsistent and confusing

evm/interchainTokenFactory.js
```diff
@@ -181,10 +180,19 @@ async function processCommand(_axelar, chain, chains, options) {
...
 
             const deploymentSalt = getDeploymentSalt(options);
 
+            const gasValue = await estimateITSFee(
+                chain,
+                destinationChain,
+                env,
+                'InterchainTokenDeployment',
+                options.gasValue,
+                _axelar?.gmpAxelarscanApi,
```
milapsheth, line 193: Pass in _axelar itself so it can read the appropriate field internally

common/utils.js
```diff
@@ -785,6 +785,39 @@ function validateDestinationChain(chains, destinationChain) {
     validateChain(chains, destinationChain);
 }
 
+async function estimateITSFee(chain, destinationChain, env, eventType, gasValue, gmpAxelarscanApi) {
+    if (!gmpAxelarscanApi || env === 'devnet-amplifier') {
+        printInfo('Using default ITS fee: 0');
+        return 0;
+    }
+
+    if (isValidNumber(gasValue)) {
```
milapsheth, line 794: Also check if it's set to auto and fail otherwise

axelar-chains-config/info/mainnet.json
benjamin852: good catch, i updated endpoint to be `/gmp/estimateITSFee` and that fixed it so will remove this.

common/utils.js
benjamin852: it's defined in the chains config file. Only hedera related assets makes use of this field. Noting that this fn was previously defined I had just moved it here to make sure there is no circular dependency between this function and the newly added `estimateItsFee`

common/utils.js
milapsheth: I just mean handling the scenario where it's not a number or set to 'auto'. That's a bad input so we should fail validation (you can validate it in commander CLI definition probably too)

common/utils.js
```diff
@@ -785,6 +785,40 @@ function validateDestinationChain(chains, destinationChain) {
     validateChain(chains, destinationChain);
 }
 
+async function estimateITSFee(chain, destinationChain, env, eventType, gasValue, _axelar) {
+    if (env === 'devnet-amplifier') {
+        printInfo('Using default ITS fee: 0');
```
milapsheth, line 790: Verbose logging in helper method isn't needed (user doesn't really care about knowing this as long as it works)

common/utils.js
```diff
@@ -785,6 +785,40 @@ function validateDestinationChain(chains, destinationChain) {
...
+
+    const res = await httpPost(url, payload);
+
+    if (res.error) {
+        throw new Error(`Error querying gas amount: ${res.error}`);
+    }
+    return res;
+}
+
+function scaleGasValue(chain, gasValue) {
```
milapsheth, line 814: Add a docstring to explain why this exists (Hedera uses lower decimal number but EVM interaction needs 18 decimals so need to scale it up)

common/utils.js
```diff
@@ -785,6 +785,40 @@ function validateDestinationChain(chains, destinationChain) {
...
+    if (typeof chain.gasScalingFactor === 'number') {
```
milapsheth, line 815: I'd create a separate task to change the behaviour of gas value override. Instead of passing in raw wei amount, let user pass in a decimal value, 1.5 eth etc. and then scale it per chain based on it's native token decimals to be more user friendly

kulikthebird (approved):
Thanks! 👍

PR #1071, opened 2025-10-04
Fix/sui link coin encoding

drewstaylor:
## why?

Currently, the code for the `link-coin` command fails with error saying the destination token is not registered. This happens due to incorrect encoding of the destination token cli argument, which is fixed in this PR.

This same fix will likely need to be applied to other areas of the source code. The other areas where this should be tested (and fixed if required) are:
- the `linkParams` parameter, which is the argument of the same function
- line 129 of gateway.js where destination token address is being encoded with BCS

### testing

- Deploy coins on Sui and another chain
- Register the token metadata for both 
- Register the Sui coin on ITS (e.g. with the `register-custom-coin` command
- Link the registered Sui coin to the destination token:

```bash
// Example:
ts-node sui/its link-coin $SYMBOL $EVM_CHAIN $EVM_TOKEN_ADDRESS --tokenManagerMode $TOKEN_MANAGER_MODE --salt $SALT --registered --channel $CHANNEL_ID
```

### DIFF
--- a/sui/its.js
+++ b/sui/its.js
@@ -579,8 +579,6 @@ async function linkCoin(keypair, client, config, contracts, args, options) {
     const { Gateway } = AxelarGateway.objects;
     const [symbol, destinationChain, destinationAddress] = args;
 
-    const encoder = new TextEncoder();
-
     const unvalidatedParams = {
         isNonEmptyString: { symbol, destinationChain, destinationAddress },
         isNonArrayObject: { tokenEntry: contracts[symbol.toUpperCase()] },
@@ -657,9 +655,10 @@ async function linkCoin(keypair, client, config, contracts, args, options) {
             channel,
             salt,
             destinationChain, // This assumes the chain is already added as a trusted chain
-            encoder.encode(destinationAddress),
+            destinationAddress,
             tokenManagerType,
-            encoder.encode(linkParams),
+            // TODO: evaluate / test encoding of linkParams
+            bcs.string().serialize(linkParams).toBytes()
         ],
     });
 
--- a/sui/its.js
+++ b/sui/its.js
@@ -658,7 +658,7 @@ async function linkCoin(keypair, client, config, contracts, args, options) {
             destinationAddress,
             tokenManagerType,
             // TODO: evaluate / test encoding of linkParams
-            bcs.string().serialize(linkParams).toBytes()
+            bcs.string().serialize(linkParams).toBytes(),
         ],
     });
 


drewstaylor:
Since this PR removes encoding into bytes of the destination address, we need to consider how to handle addresses which contain non hex chars. For example, Cosmos addresses with their bech32 prefixes.

sui/its.js
```diff
@@ -604,9 +625,11 @@ async function linkCoin(keypair, client, config, contracts, args, options) {
     const treasuryCap = coin.objects.TreasuryCap;
     const tokenManager = options.tokenManagerMode;
 
+    const destinationTokenManager = options.destinationTokenManagerMode ? options.destinationTokenManagerMode : tokenManager;
+
     // User calls registerCustomToken on ITS Chain A to register the token on the source chain.
     // A token manager is deployed on the source chain corresponding to the tokenId.
-    let txSalt = options.salt ? options.salt : null;
+    let txSalt = options.salt ? options.salt : coin.saltAddress;
```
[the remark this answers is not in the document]
drewstaylor: False positive

sui/its.js
```diff
@@ -408,22 +419,29 @@ async function giveUnlinkedCoin(keypair, client, config, contracts, args, option
     const treasuryCapOption = await txBuilder.moveCall({ target, arguments: callArguments, typeArguments });
 
     // give_unlinked_coin<T>
-    const treasuryCapReclaimerOption = await txBuilder.moveCall({
+    const [treasuryCapReclaimerOption, channelOption] = await txBuilder.moveCall({
```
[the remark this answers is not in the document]
drewstaylor: It returns 2 types now

drewstaylor:
e2e Testing Status:
* All 4 test cases in release doc are tested on Stagenet
* Testing of the 4 test cases pending TODO on Testnet

sui/gateway.js
```diff
@@ -105,9 +105,13 @@ function getProof(keypair, commandType, data, contractConfig, options) {
 async function callContract(keypair, client, config, chain, contractConfig, args, options) {
     const packageId = contractConfig.address;
 
+    const gatewayObjectId = chain.contracts.AxelarGateway.objects.Gateway;
+
     const [destinationChain, destinationAddress, payload] = args;
 
-    const gatewayObjectId = chain.contracts.AxelarGateway.objects.Gateway;
+    validateParameters({
+        isNonEmptyString: { destinationChain, destinationAddress, payload },
```
milapsheth, line 113: same

sui/gmp.js
```diff
@@ -21,6 +21,11 @@ const {
 async function sendCommand(keypair, client, chains, chain, args, options) {
     const [destinationChain, destinationAddress, feeAmount, payload] = args;
 
+    validateParameters({
+        isNonEmptyString: { destinationChain, destinationAddress },
```
milapsheth, line 25: same

sui/its-example.js
```diff
@@ -130,7 +138,7 @@ async function sendToken(keypair, client, contracts, args, options) {
             TokenId,
             Coin,
             destinationChain,
-            destinationAddress,
+            destinationTokenAddress,
```
milapsheth, line 141: this is not a token address, it's the recipient, so `encodeITSDestination` should be used

sui/its.js
```diff
@@ -228,15 +229,35 @@ async function registerCustomCoin(keypair, client, config, contracts, args, opti
     const deployConfig = { client, keypair, options, walletAddress };
     const [symbol, name, decimals] = args;
 
+    const unvalidatedParams = { isNonEmptyString: { symbol, name } };
```
milapsheth, line 232:
```suggestion
    validateParameters({ isNonEmptyString: { symbol, name } });
```
a bit confusing, just call `validateParameters` twice for clarity

sui/its.js
```diff
@@ -228,15 +229,35 @@ async function registerCustomCoin(keypair, client, config, contracts, args, opti
...
+        unvalidatedParams.isHexString = { salt: options.salt };
     }
 
+    validateParameters(unvalidatedParams);
+
     // Deploy token on Sui
     const [metadata, packageId, tokenType, treasuryCap] = await deployTokenFromInfo(deployConfig, symbol, name, decimals);
 
+    // Mint pre-registration coins
+    const amount = !isNaN(options.mintAmount) ? parseInt(options.mintAmount) : 0;
```
milapsheth, line 244: set default value of `mintAmount` to `0` instead so it's populated correctly?

sui/its.js
```diff
@@ -228,15 +229,35 @@ async function registerCustomCoin(keypair, client, config, contracts, args, opti
...
+    if (amount) {
+        const unitAmount = getUnitAmount(options.mintAmount, decimals);
+
+        const mintTxBuilder = new TxBuilder(client);
+
+        const coin = await mintTxBuilder.moveCall({
+            target: `${SUI_PACKAGE_ID}::coin::mint`,
```
milapsheth, line 251: when you split cmds, mint coin should be a separate generic util defined in `sui/tokens.js` instead

sui/its.js
```diff
@@ -365,35 +386,27 @@ async function migrateCoinMetadata(keypair, client, config, contracts, args, opt
...
-        tokenType,
-        null,
-        options.salt ? options.salt : null,
-    );
-    if (!tokenId) throw new Error(`error resolving token id from registration tx, got ${tokenId}`);
+    const decimals = coin.decimals;
+    const metadata = coin.objects.Metadata;
+    const packageId = coin.address;
+    const tokenType = coin.typeArgument;
+    const treasuryCap = coin.objects.TreasuryCap;
```
milapsheth, line 409: as part of removing the need to store token data in the config, this should be refactored to take in the coin type, and dynamically query any metadata needed

sui/its.js
```diff
@@ -478,7 +504,7 @@ async function removeUnlinkedCoin(keypair, client, config, contracts, args, opti
     contracts[symbol.toUpperCase()].objects.TreasuryCapReclaimer = null;
 }
 
-// regiser_coin_metadata
+// register_coin_metadata
```
milapsheth, line 507: unnecessary comment? function is already named after the method

sui/its.js
drewstaylor: we have 2 separate generic commands for minting coins, however the version created by @benjamin852 should be moved from `its.js` to `tokens.js`

sui/its.js
drewstaylor: this work is partially completed in another PR, and removing tokens from config is a task in JIRA

PR #1403, opened 2026-07-17
feat(stellar): add --source-account and --offline for native-multisig…

MakisChristou:
### why
- These changes were needed for the [amplifier handbook](https://app.notion.com/p/commonprefix/stellar-39fbf672de8281549b97c68955a687e1) to work with the new multisig owners 

### how

---

> [!NOTE]
> **Medium Risk**
> Changes touch key derivation, transaction building, and signing paths used for contract owner/operator actions; mistakes could produce wrong signers or partial txs, though broadcast behavior for normal single-key use is largely preserved.
> 
> **Overview**
> Stellar deployment CLIs now support **native multisig** workflows for Soroban ops: **`--source-account`** builds txs against the multisig account while one signer signs, **`--offline <file>`** writes a partially signed XDR (with a **2-day default `--timeout`**) instead of broadcasting, and **`--hd-account`** derives keys from a BIP-39 mnemonic via SEP-5 (`@scure/bip39` added). **`PRIVATE_KEY`** accepts either an `S...` seed or a mnemonic; **`getWallet`** skips fund/balance checks when offline or co-signing and surfaces clearer funding errors for single-key use.
> 
> **`broadcast`** uses the chosen source account and documents that classic Horizon **`broadcastHorizon`** still ignores multisig flags (change-trust only). **`contract.js`** stringifies falsy return values so `paused=false` / `0` still print. **`send-tokens.js`** and its README section are removed; **Node** engine is bumped to **>=24**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a70af3a571831c61a395d63ed1ad3245c15a190. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


stellar/utils.ts
rista404, line 298: should we also handle native payments being offline? like sending from a multisig?

package.json
```diff
@@ -42,6 +42,7 @@
     "@ledgerhq/hw-app-eth": "6.32.2",
     "@mysten/ledgerjs-hw-app-sui": "^0.4.1",
     "@mysten/sui": "^1.3.0",
+    "@scure/bip39": "1.6.0",
```
rista404, line 45: we should also update `package-lock.json`? I think `npm install --package-lock-only --ignore-scripts` should do it

rista404 (approved):

h4writer (approved):

PR #1408, opened 2026-08-12
feat(evm): consensus gateway pause actions and deploy script rename

jakovmitrovski:
### why

Two prerequisites for rolling the pausable consensus gateway (cgp-solidity #318, shipping in v6.5.0) out to the legacy EVM fleet.

1. `gateway.js --action setPauseStatus` was amplifier-only and authorized via `operator()`/`owner()`. Consensus gateways gain a pauser role in v6.5.0 and had no tooling at all.
2. `evm/deploy-gateway-v6.2.x.js` has been misnamed since v6.2, and its salt hardcoded `AxelarGateway v6.2` as a *prefix* that `--salt` could only append to. CI was consequently deploying with the salt `AxelarGateway v6.2AxelarGateway v6.2`.

### how

Pause actions:

- `setPauseStatus` now branches on `connectionType`. Consensus authorizes via `pauser()`/`governance()`, matching the contract's `onlyPauserOrGovernance`. The amplifier path is unchanged.
- New `paused` and `pauser` read actions, plus `transferPauser`.
- Consensus combined with `--governance` throws and points at `governance.js schedule raw`.

### DIFF
--- a/.github/workflows/test-evm.yaml
+++ b/.github/workflows/test-evm.yaml
@@ -156,9 +156,9 @@ jobs:
       - name: Deploy AxelarGasService using create2
         run: ts-node evm/deploy-upgradable.js -c AxelarGasService -m create2 -y
 
-      # same implementation address as the one deployed, since it's derivation must always be the same
+      # read the implementation deployed by the previous step
       - name: Generate AxelarGasService upgrade proposal via AxelarServiceGovernance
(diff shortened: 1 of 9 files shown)


h4writer (approved):
package.json should also upgrade axelar-cgp-solidity to 6.5.0

PR #1409, opened 2026-08-12
feat(evm): add --authModule override to consensus gateway deploy

jakovmitrovski:
### why

cgp-solidity #310 needs a new `AxelarAuthWeighted` on the deployed consensus fleet. `authModule` is a constructor `immutable`, so that means a new implementation bound to an auth that is not the live one.

Timing is the problem. The implementation must exist ~10 days before the upgrade executes on mainnet, a fresh auth starts at epoch 0, `validateProof` only accepts sets within `OLD_KEY_RETENTION = 16`, and operators rotate daily. Seed early and the auth will not know the set signing at execution, so every batch reverts, and `transferOperatorship` cannot fix it because it arrives inside a batch that must validate first.

So: deploy an unseeded auth at a CREATE2 address, bind the implementation to it, seed just before execution, hand ownership over after. This flag is the binding step. An earlier revision bound the address before deploying there; the constructor rejects that (`AxelarGateway.sol:94`).

### how

`--authModule <addr>` binds the implementation to an existing auth instead of the live one. Requires `--reuseProxy`, rejects `--reuseAuth` and `--skipExisting`, requires a matching codehash at the address, and skips the ownership transfer (handing it over early makes the auth unseedable).

`--upgrade` now reads `governance()` and routes, since it is `AxelarServiceGovernance` on every gateway in scope and a direct call would just revert. It fails with the `governance.js` commands to run instead. `--authModule` there is a cross-check against the implementation binding.

Before any auth-swapping upgrade, this script and `governance.js execute` both assert the replacement has code, a matching codehash, is seeded, holds the live set within retention, and is owned by the proxy or the seeder (`--authSeeder`). Config records it as `pendingAuthModule` until the upgrade lands.

### step order

```
consensus-auth predict / deploy
deploy-consensus-gateway --authModule --reuseProxy --reuseHelpers
governance.js schedule + proposal
consensus-auth seed && verify
governance.js execute
consensus-auth handoff && verify
```
 Pairs with #1414.

---

> [!NOTE]
> **High Risk**
> Changes consensus gateway deploy/upgrade and governance execution paths around immutable auth swaps; a mistake could brick inbound GMP batches, though the PR adds explicit pre-flight checks.
> 
> **Overview**
> Adds **`--authModule`** to `deploy-consensus-gateway` so a new implementation can be bound to a **pre-deployed** `AxelarAuthWeighted` (immutable in the constructor) instead of the live auth—needed when auth must be seeded shortly before a timelocked upgrade. The flag requires **`--reuseProxy`**, forbids **`--reuseAuth`** / **`--skipExisting`**, validates bytecode at the address, **skips** immediate ownership transfer to the proxy, and records **`pendingAuthModule`** until the upgrade runs.
> 
> Introduces shared **auth readiness** checks (`assertAuthReady`, `authReadinessProblems`, `OLD_KEY_RETENTION`): correct codehash, seeded state with the **live** operator set within retention, and owner is the gateway proxy or **`--authSeeder`**. These run before **`--upgrade`** (with **`assertCanUpgradeDirectly`** pointing at governance when needed) and before **`governance.js`** `execute` / `execute-operator-proposal` when calldata is a consensus gateway **`upgrade`** that swaps auth. Successful upgrade execution updates **`authModule`** and clears **`pendingAuthModule`**.
> 
> **`consensus-auth handoff`** now **fails** if post-transfer verification does not show proxy ownership. Unit tests cover CLI flag validation and readiness problem detection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0f2d4c6000c699cd85e1cdf2ba2259172a29d27. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/evm/deploy-consensus-gateway.js
+++ b/evm/deploy-consensus-gateway.js
@@ -83,6 +83,16 @@ async function deploy(axelar, chain, chains, options) {
 
     const contractName = 'AxelarGateway';
 
+    if (options.authModule) {
+        if (!reuseProxy) {
+            throw new Error('--authModule requires --reuseProxy');
+        }
+
+        if (reuseAuth) {
+            throw new Error('--authModule cannot be combined with --reuseAuth');
+        }
+    }
+
     const rpc = options.rpc || chain.rpc;
     const provider = getDefaultProvider(rpc);
 
@@ -179,7 +189,14 @@ async function deploy(axelar, chain, chains, options) {
 
     contractConfig.deployer = wallet.address;
 
-    if (options.skipExisting && contractConfig.authModule) {
+    if (options.authModule) {
+        // the gateway constructor reverts InvalidAuthModule on a codeless address
+        if (!(await isContract(options.authModule, wallet.provider))) {
+            throw new Error(`Auth module ${options.authModule} has no code; deploy it before the implementation`);
+        }
+
+        auth = authFactory.attach(options.authModule);
+    } else if (options.skipExisting && contractConfig.authModule) {
         auth = authFactory.attach(contractConfig.authModule);
     } else if (reuseProxy && (reuseHelpers || reuseAuth)) {
         auth = authFactory.attach(await gateway.authModule());
@@ -514,6 +531,12 @@ async function programHandler() {
         new Option('--reuseHelpers', 'reuse helper auth and token deployer contract modules for new implementation deployment'),
     );
     program.addOption(new Option('--reuseAuth', 'reuse auth module contract for new implementation deployment'));
+    program.addOption(
+        new Option(
+            '--authModule <authModule>',
+            'bind the implementation to this already deployed auth module instead of the live one (requires --reuseProxy)',
+        ),
+    );
     program.addOption(new Option('--governance <governance>', 'governance address').env('GOVERNANCE'));
     program.addOption(new Option('--mintLimiter <mintLimiter>', 'mint limiter address').env('MINT_LIMITER'));
     program.addOption(new Option('--keyID <keyID>', 'key ID').env('KEY_ID'));
--- a/evm/deploy-consensus-gateway.js
+++ b/evm/deploy-consensus-gateway.js
@@ -360,8 +360,13 @@ async function deploy(axelar, chain, chains, options) {
     const authOwner = await auth.owner();
 
     if (authOwner !== gateway.address) {
-        printError(`ERROR: Auth module owner is set to ${authOwner} instead of proxy address ${gateway.address}`);
-        error = true;
+        // an --authModule retrofit hands ownership over later, when the auth is seeded
+        if (options.authModule) {
+            printWarn(`Auth module owner is ${authOwner}, not the proxy yet. Run \`consensus-auth seed\` to hand ownership over.`);
+        } else {
+            printError(`ERROR: Auth module owner is set to ${authOwner} instead of proxy address ${gateway.address}`);
+            error = true;
+        }
     }
 
     const gatewayImplementation = await gateway.implementation();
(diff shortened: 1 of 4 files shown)


h4writer (changes requested):
Not sure if this will work? Doesn't the AxelarGateway's constructor rejects a codeless auth module?

https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/AxelarGateway.sol#L94

evm/deploy-consensus-gateway.js
h4writer: Should this also be made aware of "authmodule"?

evm/deploy-consensus-gateway.js
```diff
@@ -83,6 +83,16 @@ async function deploy(axelar, chain, chains, options) {
...
     const contractName = 'AxelarGateway';
 
+    if (options.authModule) {
+        if (!reuseProxy) {
+            throw new Error('--authModule requires --reuseProxy');
+        }
+
+        if (reuseAuth) {
+            throw new Error('--authModule cannot be combined with --reuseAuth');
+        }
```
h4writer: Also reject --skipExisting ?

evm/deploy-consensus-gateway.js
```diff
@@ -83,6 +83,16 @@ async function deploy(axelar, chain, chains, options) {
 
     const contractName = 'AxelarGateway';
 
+    if (options.authModule) {
```
h4writer: This doesn't cover "upgrade()". 
Did you look at the paths that upgrade() do and if it still supported/works?
And do we want similar or the same tests there?

evm/deploy-consensus-gateway.js
```diff
@@ -343,8 +360,13 @@ async function deploy(axelar, chain, chains, options) {
     const authOwner = await auth.owner();
 
     if (authOwner !== gateway.address) {
-        printError(`ERROR: Auth module owner is set to ${authOwner} instead of proxy address ${gateway.address}`);
-        error = true;
+        // an --authModule retrofit hands ownership over later, when the auth is seeded
+        if (options.authModule) {
+            printWarn(`Auth module owner is ${authOwner}, not the proxy yet. Run \`consensus-auth seed\` to hand ownership over.`);
```
h4writer: Nit: consensus-auth seed => consensus-auth handoff

evm/deploy-consensus-gateway.js
```diff
@@ -179,7 +189,14 @@ async function deploy(axelar, chain, chains, options) {
 
     contractConfig.deployer = wallet.address;
 
-    if (options.skipExisting && contractConfig.authModule) {
+    if (options.authModule) {
+        // the gateway constructor reverts InvalidAuthModule on a codeless address
+        if (!(await isContract(options.authModule, wallet.provider))) {
```
h4writer, line 307: Can we also check that authModule actually contains the correct code? Codehash comparison?

evm/deploy-consensus-gateway.js
jakovmitrovski: `governance()` is `AxelarServiceGovernance` everywhere (ethereum, polygon, blast, celo all 0x7Acbae6C...), so a wallet call just reverts. 

Still fine where the wallet is governance, so it reads `governance() `and routes now instead of throwing, and points you at the `governance.js`.

`upgrade()` didn't check the auth. I added a gate now (code, codehash, seeded, holds the live set, owned by proxy or seeder) and put it in `governance.js` execute too. Also added tests.

h4writer (approved):
Looks good. Thanks for the changes.

Small nit 1:
`governance.js execute` has `assertGatewayUpgradeAuthReady`.
We should add the same to `governance.js execute-operator-proposal`. If I understand correctly that allows to execute the proposal sooner, but it doesn't check if the auth module is ready.

Small nit 2:
There is no end state verification. We now have `seed/verify`, `governance execute` and lastly `handoff`. Given that handoff can fail. Should we add a warning if handoff fails and / or put another `consensus-auth verify` after `handoff` in the runbook?

PR #1411, opened 2026-08-12
docs(evm): document gateway pause commands and the versioned salt

jakovmitrovski:
### why

Two gaps left by the v6.5.0 pausable gateway work.

1. Consensus gateways became pausable in cgp-solidity 6.5.0 and `evm/gateway.js` gained the actions in #1408, but nothing documents them. The emergency-response playbook has no pause entry for `AxelarGateway` at all, while ITS documents its pause under both operator and owner actions.
2. #1408 changed `--salt` from a prefix-with-append into a full override and made the default unversioned. Following the README's gateway upgrade steps verbatim now derives a different implementation address than a release intends, and nothing says so.

### how

`evm/README.md`

- New "Gateway pause commands" section covering `paused`, `pauser`, `setPauseStatus` and `transferPauser`, with the authorization rules for both gateway types: `pauser` or `governance` on consensus, `operator` or `owner` on amplifier.
- Spells out that a freshly upgraded gateway has `pauser() == address(0)`, so the first assignment must come from governance, and that since `governance` is a contract on every environment it has to be routed via `governance.js schedule raw` rather than this script.
- States the blast radius plainly: pausing blocks `callContract` / `callContractWithToken` and gates `validateContractCall*` with a governance bypass, but leaves `execute` working, so it halts all GMP and ITS traffic on the chain rather than just token flows.
- A note on the Gateway Upgrade steps that `--salt` should be passed for a versioned deployment.

`evm/docs/emergency-response.md`

- New "Pauser Actions" section, matching the existing Operator and Mint Limiter sections.
- A "Transfer Pauser" entry under Governance Actions for the initial assignment.
- Notes that governance can pause without a pauser being set, and that this needs `schedule-operator` to avoid the timelock, which is the relevant path on any chain where the pauser is left unassigned.

### DIFF
--- a/evm/README.md
+++ b/evm/README.md
@@ -59,6 +59,8 @@ Deploy the original Axelar gateway contract for legacy consensus-based connectio
 6. Upgrade to the new implementation contract
    `ts-node evm/deploy-consensus-gateway.js --upgrade`
 
+*Note: pass `--salt` for a versioned deployment, e.g. `-s "AxelarGateway v6.5.0"`. The default salt is the unversioned `AxelarGateway`, so omitting it derives a different implementation address than a release intends.*
+
 ## AxelarGasService
 
 1. Run the following to deploy the gas service,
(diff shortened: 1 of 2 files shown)


evm/README.md
```diff
@@ -59,6 +59,8 @@ Deploy the original Axelar gateway contract for legacy consensus-based connectio
 6. Upgrade to the new implementation contract
    `ts-node evm/deploy-consensus-gateway.js --upgrade`
 
+*Note: pass `--salt` for a versioned deployment, e.g. `-s "AxelarGateway v6.5.0"`. The default salt is the unversioned `AxelarGateway`, so omitting it derives a different implementation address than a release intends.*
```
h4writer: Could it be that this only works for `-m create3` ? I think the default `-m create` doesn't use the salt. Should that be documented? Or maybe even better should we fail `-s salt` without `-m create3`?

h4writer (approved):

PR #1414, opened 2026-08-14
feat(evm): add consensus auth module deploy and verify script

jakovmitrovski:
### why

Retrofitting `AxelarAuthWeighted` on the legacy consensus gateways needs an auth contract deployed at an address known in advance, seeded with the operator sets it must accept, and finally owned by the gateway proxy. Nothing existing does this:

- `deploy-consensus-gateway.js` deploys an auth, but with plain `create`, seeded through the constructor, and only when it is **not** reusing the proxy, so it is unreachable during an upgrade.
- The SDK's `deployAndInit` helper hardcodes calling `init(...)`, which `AxelarAuthWeighted` does not have.
- `deploy-contract.js` uses plain `deploy`, which would leave the auth owned by the deployer contract with no way to move it, permanently unable to rotate operators.

### how

`evm/consensus-auth.js` with five subcommands.

`predict` prints the CREATE2 address. Constant empty-seed constructor args keep the init code, and therefore the address, independent of the operator sets. Depends on the deployer EOA, so the same key must predict and deploy.

`deploy` deploys an **unseeded** auth via `ConstAddressDeployer.deployAndInit(initCode, salt, transferOwnership(EOA))`. CREATE2 keeps the deployer contract as the constructor's `msg.sender`, so the init call can hand ownership to the deployer wallet. It must be CREATE3-free: CREATE3 deploys through an ephemeral intermediate, so `Ownable(msg.sender)` would strand ownership in a throwaway contract.

`seed` registers the current operator sets. **Repeatable and keeps ownership**, skipping anything already registered so it cannot trip `DuplicateOperators`.

`handoff` transfers ownership to the gateway proxy. Refuses unless the newest epoch matches the live operator set, because this step is one way: afterwards only the gateway can add sets, so a stale seed would be unrecoverable.

`verify` is the gate before executing an upgrade: owner is the proxy, and the newest epoch hash matches the live set.

Also exports `getAuthParams` from `deploy-consensus-gateway.js` rather than duplicating the key-ID logic, and fixes its key-rotation error, which printed `for undefined: [object Object]` because `getAuthParams` passes a string where the message read `chain.name`.

### DIFF
--- a/evm/consensus-auth.js
+++ b/evm/consensus-auth.js
@@ -5,7 +5,7 @@ const { ethers } = require('hardhat');
 const {
     ContractFactory,
     Contract,
-    utils: { keccak256 },
+    utils: { defaultAbiCoder, keccak256 },
 } = ethers;
 
 const {
@@ -25,11 +25,15 @@ const { addEvmOptions } = require('./cli-utils');
 const { getWallet } = require('./sign-utils');
 const { getAuthParams } = require('./deploy-consensus-gateway');
 
+const AxelarGateway = require('@axelar-network/axelar-cgp-solidity/artifacts/contracts/AxelarGateway.sol/AxelarGateway.json');
 const AxelarAuthWeighted = require('@axelar-network/axelar-cgp-solidity/artifacts/contracts/auth/AxelarAuthWeighted.sol/AxelarAuthWeighted.json');
 const IDeployer = require('@axelar-network/axelar-gmp-sdk-solidity/interfaces/IDeployer.json');
 
 // deploying with no seed keeps the init code, and therefore the CREATE2 address, independent of the operator sets
 const EMPTY_SEED = [[]];
+
+// AxelarAuthWeighted rejects a proof whose operator set is this many epochs behind the current one
+const OLD_KEY_RETENTION = 16;
 
 function authFactory(wallet) {
     return new ContractFactory(AxelarAuthWeighted.abi, AxelarAuthWeighted.bytecode, wallet);
@@ -47,6 +51,36 @@ async function predictAddress(wallet, chain, salt) {
     });
 
     return { address, deployerContract };
+}
+
+async function resolveAuthAddress(wallet, chain, options) {
+    if (Boolean(options.salt) === Boolean(options.authModule)) {
+        throw new Error('Provide exactly one of --salt or --authModule');
+    }
+
+    if (options.authModule) {
+        return options.authModule;
+    }
+
+    return (await predictAddress(wallet, chain, options.salt)).address;
+}
+
+// candidates are oldest first, so epochs land in the same order the live auth saw them
+async function selectSetsToSeed(auth, candidates) {
+    const seeds = [];
+
+    for (const set of candidates) {
+        // skip anything already registered: transferOperatorship reverts DuplicateOperators
+        if (!seeds.includes(set) && (await auth.epochForHash(operatorsHash(set))).eq(0)) {
+            seeds.push(set);
+        }
+    }
+
+    return seeds;
+}
+
+function encodeOperatorSet(operators, weights, threshold) {
+    return defaultAbiCoder.encode(['address[]', 'uint256[]', 'uint256'], [operators, weights, threshold]);
 }
 
 function gatewayAddress(chain) {
@@ -70,6 +104,7 @@ function operatorsHash(params) {
 }
 
 async function reportState(auth, expectedOwner) {
+    const allowedOwners = [].concat(expectedOwner || []);
     const owner = await auth.owner();
     const currentEpoch = await auth.currentEpoch();
 
@@ -80,8 +115,8 @@ async function reportState(auth, expectedOwner) {
         printInfo('Auth newest epoch hash', await auth.hashForEpoch(currentEpoch));
     }
 
-    if (expectedOwner && owner.toLowerCase() !== expectedOwner.toLowerCase()) {
-        printError(`Owner is ${owner}, expected ${expectedOwner}`);
+    if (allowedOwners.length > 0 && !allowedOwners.some((allowed) => allowed.toLowerCase() === owner.toLowerCase())) {
+        printError(`Owner is ${owner}, expected ${allowedOwners.join(' or ')}`);
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


evm/consensus-auth.js
```diff
@@ -0,0 +1,334 @@
...
+    program.name('consensus-auth').description('Deploy and verify a replacement AxelarAuthWeighted for a consensus gateway');
+
+    const addOptions = (cmd, { salt = true } = {}) => {
+        addEvmOptions(cmd);
+
+        if (salt) {
+            cmd.addOption(new Option('-s, --salt <salt>', 'CREATE2 salt for the auth module').makeOptionMandatory(true));
+        }
+
+        cmd.addOption(new Option('--prevKeyIDs <prevKeyIDs>', 'comma separated older key IDs to seed alongside the current one'));
```
h4writer: I think this shouldn't be here but only for the "seed" command?

evm/consensus-auth.js
```diff
@@ -0,0 +1,334 @@
...
+    const program = new Command();
+
+    program.name('consensus-auth').description('Deploy and verify a replacement AxelarAuthWeighted for a consensus gateway');
+
+    const addOptions = (cmd, { salt = true } = {}) => {
+        addEvmOptions(cmd);
+
+        if (salt) {
+            cmd.addOption(new Option('-s, --salt <salt>', 'CREATE2 salt for the auth module').makeOptionMandatory(true));
+        }
```
h4writer: What is the idea of "salt". In this example "salt" is never false. Should we remove this variable?

evm/consensus-auth.js
```diff
@@ -0,0 +1,334 @@
...
+    printInfo('Auth deployed unseeded; run `seed` shortly before executing the gateway upgrade', address);
+}
+
+async function seed(axelar, chain, chains, options) {
+    const provider = ethers.getDefaultProvider(chain.rpc);
+    const wallet = await getWallet(options.privateKey, provider, options);
+    const gasOptions = await getGasOptions(chain, options, 'AxelarAuthWeighted');
+
+    const proxy = gatewayAddress(chain);
+    const address = options.authModule || (await predictAddress(wallet, chain, options.salt)).address;
```
h4writer: `options.salt` is currently mandatory. So using `options.authModule` doesn't make much sense, given you also have to provide salt?

Should we:
1) keep `salt` mandatory and test here that `options.authModule `address is the same as `predictAddress`
2) Or don't make `salt` mandatory and make the runner choose between `options.authModule` or `options.salt`. And never both?

evm/consensus-auth.js
```diff
@@ -0,0 +1,334 @@
...
+    const auth = authFactory(wallet).attach(address);
+
+    if (!(await reportState(auth, wallet.address))) {
+        throw new Error('Auth ownership was not handed to the deployer wallet; it cannot be seeded');
+    }
+
+    printInfo('Auth deployed unseeded; run `seed` shortly before executing the gateway upgrade', address);
+}
+
+async function seed(axelar, chain, chains, options) {
```
h4writer, line 252: Wouldn't it be better if seed just copies all current seeds of the current implementation? Making sure we have the data of all 16 epochs?

h4writer (commented):
So if I understand correctly the idea is to create an empty authModule using EOA, which we still control. That way the Gateway can be pointing to that authModule during the governance proposal. Just before the Gateway is upgraded, we seed the authModule and give the Gateway the ownership.

1) consensus-auth deploy
2) deploy-consensus-gateway --reuseProxy --reuseHelpers --authModule 0x9c3f…
3) consensus-auth seed
4) consensus-auth handoff
5) consensus-auth verify

A few questions I had while reading the code and weren't immediately clear to me.
- consensus-auth seed. Will we run this multiple times (on every rotation) during governance proposal. Or just before we upgrade the gateway?
- do we already want a runbook in releases/ or will you do that later?

evm/consensus-auth.js
jakovmitrovski: Chose option 2, seems like the cleaner option.

evm/consensus-auth.js
jakovmitrovski: I implemented a configurable param, to be able to have the data for `x` epochs. I am wondering if synching all 16 epochs may cause historical data access problems.

jakovmitrovski:
> So if I understand correctly the idea is to create an empty authModule using EOA, which we still control. That way the Gateway can be pointing to that authModule during the governance proposal. Just before the Gateway is upgraded, we seed the authModule and give the Gateway the ownership.

Pretty much. Though, the gateway isn't pointing at the new auth during the proposal. The implementation is bound to it, but the proxy keeps using the old one until the upgrade executes. So, an unseeded auth sitting there for 10 days doesn't hurt.
 
> 1. consensus-auth deploy
> 2. deploy-consensus-gateway --reuseProxy --reuseHelpers --authModule 0x9c3f…
> 3. consensus-auth seed
> 4. consensus-auth handoff
> 5. consensus-auth verify

Only change to your order is handoff moving after the upgrade:
1. consensus-auth predict / deploy
2.  deploy-consensus-gateway --reuseProxy --reuseHelpers --authModule 0x9c3f…
3.  governance.js schedule + proposal
4.  consensus-auth seed
5.  consensus-auth verify
6. governance.js execute
7.  consensus-auth handoff

If a rotation lands between seeding and execution we still own the auth and fix it with one tx.

> A few questions I had while reading the code and weren't immediately clear to me.
>
> * consensus-auth seed. Will we run this multiple times (on every rotation) during governance proposal. Or just before we upgrade the gateway?

Just once right before the execute.

> * do we already want a runbook in releases/ or will you do that later?

I think we should do the runbook later. Want to actually run this on devnet first and write it from that. Wdyt?

evm/consensus-auth.js
jakovmitrovski: `salt` is the `CREATE2` salt, it's what gives us a deterministic address so predict and deploy agree where the auth lands and we can bind the implementation to it before it exists. seed/handoff/verify only need it to re-derive that address. They now take the `salt` or the `address` instead.

h4writer (approved):
Looks good. I only got some nits about `seed`

1) Do you still need `--prevKeyIDs` or can we just remove it to decrease complexity of the script and since we have `--copyEpochs`. If you prefer to keep, maybe make it that you can only use one or the other?

2) I think you can shoot yourself in the foot with `seed --copyEpochs`
```
consensus-auth seed --copyEpochs 3   // copies the 3 most recent ones -> epochs 1,2,3
consensus-auth seed --copyEpochs 5   // the 3 newest are already registered, so it only
                                     // copies the 2 older ones, on top, at epochs 4,5.
                                     // The newest epoch is now an older operator set.
consensus-auth handoff               // fails: newest epoch on the new auth != current
                                     // operator set on the live auth
```

